PR #282, opened 2014-12-15
Prevent duplicate session in auth code grant

### LINKED ISSUE #266: Two sessions created for each auth code
Hey Alex, whenever I have a client request a new access token via the auth code flow, two sessions are created. One of them is tied to the auth code when it gets issued, the other is tied to the access token when it gets issued. After I have received the access and refresh token, the two sessions stay in my database even though one is effectively useless (at least as far as I can tell, I suppose). I'm not calling any code that would create duplicate sessions, so it has to be something happening in the package I'd imagine.

So, I was wondering, is this indeed how it is supposed to be? And if it is how it is supposed to be, could you please explain to me why OAuth handles sessions that way? I'd really appreciate it. Thanks!

FYI, I'm using @lucadegasperi's Laravel bridge package to implement my server.

maknz:
The session already exists in the database, so we don't need to save it again. Doing so results in the session used for the auth code hanging around in the database with nothing associated to it, while the access token is associated to a new session caused by the `save()` method creating a duplicate. Fixes #266.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -260,7 +260,6 @@ class AuthCodeGrant extends AbstractGrant
         $code->expire();
 
         // Save all the things
-        $session->save();
         $accessToken->setSession($session);
         $accessToken->save();
 


vvllaadd:
The fix is good but incomplete, the save() call was used to associating the auth code scopes to the session. See: https://github.com/thephpleague/oauth2-server/issues/299#issuecomment-70008297 for the required change for fixing this.

PR #294, opened 2015-01-06
Save the new access token into the old refresh token

bp1222:
With my change for preventing RefreshTokens from being rotated, a potential problem exists with storage.

In the RefreshToken complete flow, we go ahead and expire the old Access Token.  Since the library doesn't define what expire entails one might delete the old access token from the storage.  If this is the case, then once we issue a new Access Token, we never update the Access Token the Refresh Token relates to.  So the second time requesting a new Access Token via Refresh Token, we attempt to pull the Access Token from storage, but can't (since it -may- have been deleted).

This change sets the new Access Token to the old Refresh Token, and calls save.  However, and maybe this is kinda goofy, save(), is more of a wrapper for storage::create().  So it would be up to the user to have create do updates as well. (Read: rather than an `insert into` storage containers could do a `replace into`, since refresh_token_id, is the primary key it's ok)

### DIFF
--- a/src/Grant/RefreshTokenGrant.php
+++ b/src/Grant/RefreshTokenGrant.php
@@ -186,6 +186,8 @@ class RefreshTokenGrant extends AbstractGrant
             $this->server->getTokenType()->setParam('refresh_token', $newRefreshToken->getId());
         } else {
             $this->server->getTokenType()->setParam('refresh_token', $oldRefreshToken->getId());
+            $oldRefreshToken->getAccessToken($newAccessToken);
+            $oldRefreshToken->save();
         }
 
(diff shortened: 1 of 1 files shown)


bp1222:
@alexbilbie  hate to be "that guy" but, ping

atrauzzi:
@bp1222 FWIW, I've introduced a major refactor to this package: https://github.com/thephpleague/oauth2-server/issues/314

It may help when addressing these kinds of issues by slightly tweaking the design.  As part of it, I allow multiple access tokens per refresh token and have removed the concept of "sessions".  Although @alexbilbie still has to review it, I'm hoping to get as much input as possible.

PR #303, opened 2015-01-23
Boyscouting the php docs to always use FQCNs

hannesvdvreken:


alexbilbie:
Thanks @hannesvdvreken

PR #308, opened 2015-01-29
Associate Scopes to Session in AuthCode Grant.

bp1222:
Scopes assigned to the $session for AuthCode grant completeFlow() are never actually triggering a write in the StorageContainer.

associateScope() only adds the scope into the scopes array.  Through the AuthCode complete, we never call save() on the $session.  Save would be wrong as well, since it's a wrapper for create, not update.  So new method is needed to correctly issue a write to the Storage to associate the scope to the session.

### DIFF
--- a/src/Entity/SessionEntity.php
+++ b/src/Entity/SessionEntity.php
@@ -299,8 +299,15 @@ class SessionEntity
         );
 
         $this->setId($id);
+        $this->applyScopes();
+    }
 
-        // Associate the scope with the session
+    /**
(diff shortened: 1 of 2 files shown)


bp1222:
I should note, that this change does expect the implementation of the Storage container to anticipate you might get an associate request for a scope that already exists.  For an RDBMS you could use replace rather than an insert.  For other types of storage, you would need to handle accordingly.

bp1222:
@alexbilbie hate to be "that guy" but, ping

PR #319, opened 2015-03-02
Fixed missing session scope

### LINKED ISSUE #297: Refreshed access token (via refresh_token grant) does not get the same scope as original access token
In the following method, notice that SessionEntity is never associated with the requested ScopeEntities.

> League\OAuth2\Server\Grant\AuthCodeGrant::newAuthorizeRequest(...)

Also notice that in the following method, ScopeEntities are associated to SessionEntity but the save() method of SessionEntity is never called (therefore not persisted into database).

> League\OAuth2\Server\Grant\AuthCodeGrant::completeFlow()

However in the following method (when using refresh_token grant type), it tries to retrieve ScopeEntities from the SessionEntity which gives an empty array. As a result the new access token ("REFRESHED" token) is never re-populated with the scopes from the original access token.

> League\OAuth2\Server\Grant\RefreshTokenGrant::completeFlow()

Fuxy22:
Close #297

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -148,7 +148,6 @@ class AuthCodeGrant extends AbstractGrant
         $session = new SessionEntity($this->server);
         $session->setOwner($type, $typeId);
         $session->associateClient($authParams['client']);
-        $session->save();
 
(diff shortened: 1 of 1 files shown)


hoyo:
+1

PR #323, opened 2015-03-20
Updated Interface Docs

rdohms:
Made phpdocs match expectations like null when not found and using array notation for indicating array of <object>


alexbilbie:
Thanks :+1:

Issue #321, opened 2015-03-04
Exceptions are not helpful, must have error description and send it with response


mcd-php:
See: [AuthCodeGrant.php#L219](https://github.com/thephpleague/oauth2-server/blob/ea6edf572a79e133a141e1c65165a5dece2ecdf6/src/Grant/AuthCodeGrant.php#L219)
```php
        // Ensure the auth code hasn't expired
        if ($code->isExpired() === true) {
            throw new Exception\InvalidRequestException('code');
        }
```
**Observed**: No error reason in response, no reason in log: 
> exception 'League\OAuth2\Server\Exception\InvalidRequestException' with message 'The request is missing a required parameter, includes an invalid parameter value,includes a parameter more than once, or is otherwise malformed. Check the "code" parameter.'

Absolutely clueless and misleading, have to look into the source.

**Expected**: Helpful and immediately available error messages in logs and in response, not only in code comments:

> exception '...' with message 'The code has expired'

alexbilbie:
All of the error messages are taken from the OAuth 2.0 specification.




If you can find a more appropriate error from the spec please let me know and I'll change if I agree, otherwise for now it's how it is.



—

mcd-php:
I have read http://tools.ietf.org/html/rfc6749 , section 5.2 allows optional `error_description`. 

Yes, it will take rework of exception hierarchy, but not giving human-worded description of error will impede cooperation between services using the library, and their clients.

I have both client and server in neighbour IDE windows (splitting my project in SOA way), but what if client is foreign developer ? How can he know what failed in server, or what is his own fault ?

alexbilbie:
I agree. I'll have a look into what can be done add support for the `error_description` field

vvllaadd:
+1 on this, an error description will help a lot

tpavlek:
Thank goodness there's progress on this. Just spent an hour tearing out my hair because I thought I wasn't including an auth code, when it was just expired in fact.

Thumbs up alex :+1:

dav-m85:
Issue to be closed ?

mcd-php:
Here's the closing criteria:

Let's https://alice.net uses this library, and https://bob.org has junior web-master Pogo who tries to establish trust with https://alice.net, but does not know what language and library does the https://alice.net use, let alone any opportunity to view it's source and debug side by side.

[...]

Issue #318, opened 2015-03-02
Make all Entity use EntityTrait or otherwise be uniformly bulk-settable


mcd-php:
Now `EntityTrait` is only used by `ClientEntity` and `ScopeEntity`.
Why not `EntityInterface` with `hydrate()` ? 
Why not all entities are made equal ?

Now to hydrate all entities uniformly I had to re-implement [YiiBase::configure()](https://github.com/yiisoft/yii2/blob/82c1136613f531b6830e6dcdf5d055d8193c6b99/framework/BaseYii.php#L515) with reflection ~~and hookers~~.

rdohms:
:+1: 

Actually I believe the hydrate method should actually be static allowing you to build a new Entity without an instance of it.

`$client = ClientEntity::build($array)`

Makes for clean and expressive code.
@alexbilbie I would be glad to PR this is you define the best direction to go with it.

mcd-php:
Static is plague, if not otherwise proven with math rigor on case-by-case basis. For example, here it will kill partial hydration, constructor options and class inheritance (may resurrect w/ ugly functions like `get_called_class()`, equals to reflection and occult magick).

Just hoist `hydrate()` into base class like `\League\OAuth2\Server\Entity\AbstractTokenEntity`, and turn into `HydratableInterface` and `HydrateTrait`.

Here's how I use it, so ugly:
```php
        $entity = new $entityClass($this->_server);
        // https://github.com/thephpleague/oauth2-server/issues/318
        if ( method_exists($entity,'hydrate')) {
            $entity->hydrate($fields);
        } else {
            ObjectHelper::configureForeign($entity,$fields);
        }
```

rdohms:
@mcd-php static is plague is used incorrectly.

This case an entity is a Value Object and using a static to instantiate a VO is perfectly usable.
Its my understanding the constructor in this case should not be changing anyway.

Class inheritance issues can be dealt with a simple `new static`.

Either way, the static is not the issue and would just make more readable code.
This can also be achieved in 2 steps anyway, one to make hydrate available in every object and second to have a trait for static building.

alexbilbie:
Agreed about using EntityTraits.

The `hydrate` method needs to die.

rdohms:
@alexbilbie you lost me. :P which path do you want to take?

alexbilbie:
So v5 will have entity traits which implement all of the required methods for this library to work. Then you can use your own Doctine/Eloquent/whatever entities.

[...]

Issue #307, opened 2015-01-29
Replace all occurrences of the term "Storage" with "Repository"


atrauzzi:
It's a minor distinction, but your storage classes and interfaces abide more or less to the repository pattern.  By naming it such, you'll get a lot of free _understanding_ from people consuming the library.

alexbilbie:
I agree and in my local out-of-sync copy I've actually already done this - I will merge with dev soon :+1:

atrauzzi:
Haha, awesome.  Sorry for being such a pain! :)

Feel free to use this issue to track/close with the commit.

alexbilbie:
No trouble at all, I'm just glad I'm helping people out with my code :grinning:

m4tthumphrey:
Could the repositories possibly introduce factories so we can extend the Entity classes? Or provide factories for each of the Entity types as standalone classes?

alexbilbie:
@m4tthumphrey yep it's coming in v5. I might be able to backport to v4 too

m4tthumphrey:
@alexbilbie amazing! Any timeline for either of those scenarios? :smile:

alexbilbie:
This is done in V5

Issue #306, opened 2015-01-29
Remove `session` concept as it's confusing


alexbilbie:

ashleydw:
I'm wondering if you have WIP for this? I'd like to remove the session concept too.

atrauzzi:
I think @alexbilbie is fairly busy right now so I'm taking it upon myself to refactor this package.  Part of this refactor is going to be to remove the session concept.  It's going to take a bit as I'm cleaning up the injection and coupling in the library.

To that end, I think the end result is going to be that authorization tokens, refresh tokens and authorization codes are all going to have a relation to auth codes.  I'm just mentioning this here in case someone notices that I've missed something! ;)

alexbilbie:
I am busy but I'm actually sat doing some updates as I type this.




Only one grant uses auth codes so it'd be a mistake to base relations on those. I've got some ideas on a more elegant way of removing sessions.



—

atrauzzi:
I suspect both of our changes are quite an overhaul, hopefully they're compatible in some way.  The basic gist of it is that I'm making the library able to support any ORM and as I mentioned already, reworking the dependencies.

I've got base entity interfaces, I've converted your actual Entity *classes* into traits.  The idea being that any integration (Laravel, Doctrine) can be done as a simple repository + entity library, optionally leveraging the traits for default implementation.

I mentioned elsewhere that I'll be working on this 9-5, so if you want to hold off in any way and preview what I've done, I should be done by the end of the week.  It won't be PSR-2 initially, but that's only on the off chance that you totally hate it and I have to keep my fork.  If you do like it, I'll do the little bit of work to make it proper PSR-2.

atrauzzi:
@alexbilbie Here's a preview of things: https://github.com/atrauzzi/oauth2-server/tree/cleanup/src/Entity

I'm still tearing things apart in the rest of the repository.  But feel free to poke around to see what I've done with repositories, and services.  I'm just working on the grants right now.

atrauzzi:
@alexbilbie An example of how I'm thinking grants can look: https://github.com/atrauzzi/oauth2-server/blob/e390a7793acd84e92319457f6a7c7a9272df14d8/src/GrantType/AuthorizationCode.php

Basically flow dispatching is always handled at the service level and grants don't have to be directly called by implementers. Constants are used when grants support/require more than one flow.

[...]

PR #326, opened 2015-04-05
Version 5

alexbilbie:
### Goal

The goal of version 5 is to make this library significantly easier to implement. The big pain points are the "storage" repositories are too verbose, and using this library's entities with other implementations (e.g. Doctrine entites/Eloquent models). Also remove the concept of a session as it isn't an OAuth concept.

An example implementation of each of the grant types will be added here as I finish implementing them - https://github.com/thephpleague/oauth2-server/tree/v5/examples/public. These examples will form the basis of functional tests too.

### TODO

- [x] Replace all occurrences of the term "Storage" with "Repository" - #307
- [x] Simplify repositories
- [x] Introduce EntityTrait and entity interfaces - #318
- [ ] Make exceptions more helpful - #321
- [ ] Integration with StackPHP - #200
- [x] Add support for JWT - #209
- [ ] Remove `session` concept as it's confusing - #306
- [ ] Update documentation - #193
- [ ] Find a solution to running Codeception tests on HHVM in TravisCI
- [ ] Relational database map (maybe as part of documentation?)

Grant refactors:
- [x] Client credentials
- [x] Password
- [ ] Auth code
- [ ] Refresh token
- [ ] MAC


rdohms:
Looking good, will give it a closer look alter when time becomes available.
One thing i think would be great is to add a Relational DB map, or at least some description of the relations between the entities, right now you have to reverse engineer it from the examples and then turn that into implementation. Having a suggested diagram would make that step a lot faster.

alexbilbie:
@rdohms I will do that

examples/public/client_credentials.php
```diff
@@ -0,0 +1,32 @@
...
+include(__DIR__ . '/../vendor/autoload.php');
+
+// Setup the authorization server
+$server = new Server();
+$server->addRepository(new ClientRepository());
+$server->addRepository(new ScopeRepository());
+$server->addRepository(new AccessTokenRepository());
+
+// Enable the client credentials grant which will return access tokens that last for 24 hours
+$server->enableGrantType('ClientCredentialsGrant', null, new \DateInterval('PT24H'));
```
shadowhand: Why is the grant type added using a string, instead of a specific class?

src/Entities/Traits/ClientEntityTrait.php
```diff
@@ -0,0 +1,28 @@
...
+        return $this->name;
+    }
+
+    /**
+     * Set the client's name
+     * @param string $name
+     */
+    public function setName($name)
+    {
+        $this->name = $name;
```
shadowhand: Personally, I really like when setters `return $this;` so that setters can be chained:

```
$something->setFoo(...)->setBar(...);
```

Would this be an acceptable modification?

src/Entities/Traits/TokenEntityTrait.php
```diff
@@ -0,0 +1,135 @@
...
+        $this->client = $client;
+    }
+
+    /**
+     * Has the token expired?
+     * @return bool
+     */
+    public function isExpired()
+    {
+        return (new \DateTime()) > $this->getExpiryDateTime();
```
shadowhand: Any reason not to have `use DateTime;` at the top instead of rooting the namespace here?

src/Grant/AbstractGrant.php
```diff
@@ -29,112 +32,87 @@
...
-        if ($this->accessTokenTTL) {
-            return $this->accessTokenTTL;
-        }
-
-        return $this->server->getAccessTokenTTL();
+    public function __construct(
+        Emitter $emitter,
+        ClientRepositoryInterface $clientRepository,
+        ScopeRepositoryInterface $scopeRepository,
+        AccessTokenRepositoryInterface $accessTokenRepository
```
shadowhand: Is there a strong reason to define the dependencies via the constructor? If not, I would suggest removing the `__construct` method entirely and using setters for each dependency. It's more flexible and make dependency injection more flexible.

src/Grant/AbstractGrant.php
```diff
@@ -29,112 +32,87 @@
...
-        $scopesList = explode($this->server->getScopeDelimiter(), $scopeParam);
+    public function validateScopes(
+        $scopeParamValue,
+        $scopeDelimiter,
+        ClientEntityInterface $client,
+        $redirectUri = null
+    ) {
+        $scopesList = explode($scopeDelimiter, trim($scopeParamValue));
 
         for ($i = 0; $i < count($scopesList); $i++) {
```
shadowhand: Couldn't this `for` loop be replaced by an `array_filter`?

src/Grant/AuthCodeGrant.php
```diff
@@ -11,14 +11,22 @@
...
+use League\OAuth2\Server\Exception\InvalidRequestException;
+use League\OAuth2\Server\Repositories\AccessTokenRepositoryInterface;
+use League\OAuth2\Server\Repositories\AuthCodeRepositoryInterface;
+use League\OAuth2\Server\Repositories\ClientRepositoryInterface;
+use League\OAuth2\Server\Repositories\RefreshTokenRepositoryInterface;
+use League\OAuth2\Server\Repositories\ScopeRepositoryInterface;
+use League\OAuth2\Server\TokenTypes\TokenTypeInterface;
+use League\OAuth2\Server\Utils\SecureKey;
+use Symfony\Component\HttpFoundation\Request;
+use DateInterval;
```
shadowhand: :+1:

src/Grant/AuthCodeGrant.php
```diff
@@ -26,6 +34,32 @@
...
+        Emitter $emitter,
+        ClientRepositoryInterface $clientRepository,
+        ScopeRepositoryInterface $scopeRepository,
+        AccessTokenRepositoryInterface $accessTokenRepository,
+        AuthCodeRepositoryInterface $authCodeRepository,
+        RefreshTokenRepositoryInterface $refreshTokenRepository = null
+    ) {
+        $this->authCodeRepository = $authCodeRepository;
+        $this->refreshTokenRepository = $refreshTokenRepository;
+        parent::__construct($emitter, $clientRepository, $scopeRepository, $accessTokenRepository);
```
shadowhand: This is a perfect example of why I think setters are better than constructors... knowing what order these are in the parent class requires additional cognitive load that can be avoided.

src/Grant/ClientCredentialsGrant.php
```diff
@@ -31,92 +33,70 @@ class ClientCredentialsGrant extends AbstractGrant
...
-     *
-     * @throws
-     */
-    public function completeFlow()
-    {
+    public function getAccessTokenAsType(
+        Request $request,
+        TokenTypeInterface $tokenType,
+        DateInterval $accessTokenTTL,
+        $scopeDelimiter = ' '
```
shadowhand: Wouldn't it be preferable to have the scope delimiter configured at the class level, instead of a method parameter?

src/Grant/ClientCredentialsGrant.php
```diff
@@ -31,92 +33,70 @@ class ClientCredentialsGrant extends AbstractGrant
...
+        $scopeParam = $request->request->get('scope', '');
+        $scopes = $this->validateScopes($scopeParam, $scopeDelimiter, $client);
 
         // Generate an access token
-        $accessToken = new AccessTokenEntity($this->server);
-        $accessToken->setId(SecureKey::generate());
-        $accessToken->setExpireTime($this->getAccessTokenTTL() + time());
+        $accessToken = new AccessTokenEntity();
+        $accessToken->setIdentifier(SecureKey::generate());
+        $accessToken->setExpiryDateTime((new \DateTime())->add($accessTokenTTL));
```
shadowhand: This was confusing to read the first time around. I would prefer to see the complete `DateTime` constructed outside:

```php
$expiration = (new DateTime)->add($accessTokenTTL);
```

src/Grant/ClientCredentialsGrant.php
```diff
@@ -31,92 +33,70 @@ class ClientCredentialsGrant extends AbstractGrant
...
-        $session->associateClient($client);
+        $scopeParam = $request->request->get('scope', '');
+        $scopes = $this->validateScopes($scopeParam, $scopeDelimiter, $client);
 
         // Generate an access token
-        $accessToken = new AccessTokenEntity($this->server);
-        $accessToken->setId(SecureKey::generate());
-        $accessToken->setExpireTime($this->getAccessTokenTTL() + time());
+        $accessToken = new AccessTokenEntity();
+        $accessToken->setIdentifier(SecureKey::generate());
```
shadowhand: I don't feel __very__ strongly about this, but I think it would be preferable to avoid static classes whenever possible. This kind of usage prevents the `SecureKey` class from being overloaded, because it is a non-injected dependency.

src/Grant/PasswordGrant.php
```diff
@@ -46,137 +47,133 @@ class PasswordGrant extends AbstractGrant
...
+    public function getAccessTokenAsType(
+        Request $request,
+        TokenTypeInterface $tokenType,
+        DateInterval $accessTokenTTL,
+        $scopeDelimiter = ' '
+    ) {
         // Get the required params
-        $clientId = $this->server->getRequest()->request->get('client_id', $this->server->getRequest()->getUser());
+        $clientId = $request->request->get('client_id', $request->getUser());
         if (is_null($clientId)) {
```
shadowhand: What if `$clientId = ''`? Is it valid?

src/Grant/PasswordGrant.php
```diff
@@ -46,137 +47,133 @@ class PasswordGrant extends AbstractGrant
...
-        if (($client instanceof ClientEntity) === false) {
-            $this->server->getEventEmitter()->emit(new Event\ClientAuthenticationFailedEvent($this->server->getRequest()));
+        if (($client instanceof ClientEntityInterface) === false) {
+            $this->emitter->emit(new Event('client.authentication.failed', $request));
             throw new Exception\InvalidClientException();
         }
 
-        $username = $this->server->getRequest()->request->get('username', null);
+        $username = $request->request->get('username', null);
         if (is_null($username)) {
```
shadowhand: What if `$username = ''`?

src/Grant/PasswordGrant.php
```diff
@@ -46,137 +47,133 @@ class PasswordGrant extends AbstractGrant
...
 
         // Check if user's username and password are correct
-        $userId = call_user_func($this->getVerifyCredentialsCallback(), $username, $password);
+        $user = $this->userRepository->getByCredentials($username, $password);
 
-        if ($userId === false) {
-            $this->server->getEventEmitter()->emit(new Event\UserAuthenticationFailedEvent($this->server->getRequest()));
+        if (($user instanceof UserEntityInterface) === false) {
+            $this->emitter->emit(new Event('user.authentication.failed', $request));
             throw new Exception\InvalidCredentialsException();
```
shadowhand: Is it okay to emit an event and then immediately throw an exception? I haven't used this pattern before and I'm curious how it would be handled...

src/Repositories/ClientRepositoryInterface.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Client storage interface
+ */
+interface ClientRepositoryInterface extends RepositoryInterface
+{
+    /**
+     * Get a client
+     *
+     * @param string $clientIdentifier The client's identifier
+     * @param string $clientSecret     The client's secret (default = "null")
```
shadowhand: I don't think the `(default ...)` is necessary here at all, and the current documentation is actually a bit misleading, because it makes it look like the default value is a string `"null"`.

src/Repositories/RepositoryInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ * @license     http://mit-license.org/
+ * @link        https://github.com/thephpleague/oauth2-server
+ */
+
+namespace League\OAuth2\Server\Repositories;
+
+/**
+ * Repository interface
+ */
+interface RepositoryInterface
```
shadowhand: If the default interface is completely empty, why are other interfaces extending it? I don't think that's a good idea.

src/Server.php
```diff
@@ -0,0 +1,137 @@
...
+     * @param TokenTypeInterface $defaultResponseType
+     * @param DateInterval                                                $defaultAccessTokenTTL
+     */
+    public function __construct(
+        TokenTypeInterface $defaultResponseType = null,
+        DateInterval $defaultAccessTokenTTL = null
+    ) {
+        $this->defaultResponseType = ($defaultResponseType instanceof TokenTypeInterface)
+            ? $defaultResponseType
+            : new BearerTokenType();
```
shadowhand: This is another good example of why setters are preferable to constructor arguments... the default values can be *always* be created in the constructor without having to check if a value is given:

```php
if (empty($this->defaultResponseType)) {
    $this->defaultResponseType = ...;
}
if (empty(...)) { ... }
````

File: src/AbstractServer.php [hunk 17]
```diff
@@ -35,107 +39,14 @@
...
-
-    /**
-     * Returns the event emitter
-     *
-     * @return \League\Event\Emitter
-     */
-    public function getEventEmitter()
-    {
-        return $this->eventEmitter;
+        $this->setContainer(new Container());
```
shadowhand: I **very** strongly feel that this is not a good idea. Putting a dependency container inside of an interactor completely violates [IoC](https://en.wikipedia.org/wiki/Inversion_of_control).

tests/api/ApiTester.php
```diff
@@ -0,0 +1,2389 @@
+<?php //[STAMP] b08a68f76e9949287107c90f548717f0
+
+// This class was automatically generated by build task
```
shadowhand: Shouldn't auto-generated files be excluded from the respository?

src/Entities/Traits/ClientEntityTrait.php
alexbilbie: Previous versions provided a fluent interface so yes I guess so

src/Entities/Traits/TokenEntityTrait.php
alexbilbie: Nope, will fix

src/Grant/AbstractGrant.php
alexbilbie: Yes

src/Grant/ClientCredentialsGrant.php
alexbilbie: For the sake of readability sure

src/Grant/ClientCredentialsGrant.php
alexbilbie: The SecureKey class has always been overridable using:

```php
SecureKey::setAlgorithm(KeyAlgorithmInterface $algorithm);
```

src/Grant/PasswordGrant.php
alexbilbie: It isn't, you either have whitelisted clients or you don't. And this library only supports whitelisted clients for the sake of best practise out of the box.

src/Grant/PasswordGrant.php
alexbilbie: Yep, handles as expected - anything listening to the event is executed, then the exception is thrown.

src/Repositories/ClientRepositoryInterface.php
alexbilbie: Fixed

src/Server.php
alexbilbie: I can't think of a reason why to not do this, good point

tests/api/ApiTester.php
alexbilbie: Yep

alexbilbie:
@shadowhand I've pushed a whole load of fixes that address your comments.

Regarding the container; my intention was to use the container to bootstrap all of the built in grants to reduce the amount of bootstrapping user land code required.

I like constructor injection because then there are guarantees that the required dependencies have been set. With setter injection you need to verify that you have all of the required dependencies set before you can proceed.

I'm not necessarily convinced by the container myself but if I didn't try it out I wouldn't know :smile:

src/Grant/ClientCredentialsGrant.php
shadowhand: That let's us change the algorithm, but what about the implementation?

src/Grant/PasswordGrant.php
shadowhand: My point is that using `is_null` is not an appropriate test here. The correct check is to use `!empty($clientId)` or simple `!$clientId`.

src/Grant/PasswordGrant.php
shadowhand: Same thing here... checking for `null` is only one of several possible empty values, none of which are valid.

examples/src/Entities/UserEntity.php
```diff
@@ -0,0 +1,17 @@
...
+
+class UserEntity implements UserEntityInterface
+{
+    /**
+     * Return the user's identifier
+     * @return mixed
+     */
+    public function getIdentifier()
+    {
+        return 1;
```
shadowhand: Shouldn't `UserEntity` be an abstract class?

src/Entities/Traits/TokenEntityTrait.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * Get an associated scope by the scope's identifier
+     * @param string $identifier
+     * @return ScopeEntityInterface|null  The scope or null if not found
+     */
+    public function getScopeWithIdentifier($identifier)
+    {
+        return (isset($this->scopes[$identifier])) ? $this->scopes[$identifier] : null;
```
shadowhand: What do you think about creating a `league/utils` package that has common traits for this kind of thing, for instance:

```php
trait ArrayKeyOrDefault
{
    protected function getKeyOrDefault(array $array, $key, $default = null)
    {
        return isset($array[$key]) ? $array[$key] : $default;
    }
}
```

I often find myself writing these same operations over and over.

composer.json
```diff
@@ -6,11 +6,15 @@
 	"require": {
 		"php": ">=5.4.0",
 		"symfony/http-foundation": "~2.4",
-		"league/event": "1.0.*"
+		"league/event": "~2.1",
+		"league/container": "~1.0",
+		"firebase/php-jwt": "~2.0"
```
shadowhand: Considering that this package hasn't been updated in a very long time, and the [recent report of vulnerabilities in JWT packages](https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/), maybe we don't want to depend on this package?

Sadly, there aren't many alternatives that do `HS*` validations, it might mean forking the package over to League.

src/Server.php
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * @var string
+     */
+    protected $scopeDelimiter = ' ';
+
+    /**
+     * New server instance
+     */
+    public function __construct() {
+        $this->defaultTokenType = new BearerTokenType();
```
shadowhand: This isn't exactly what I was suggesting, and I'm not sure it is a good idea... if a class extends `Server` and defines a default, this will overwrite it. I think you need the check, or you need to define a getter for it, as per my examples.

src/Entities/Traits/TokenEntityTrait.php
rdohms: or maybe use something that already exists? https://github.com/mathiasgrimm/arraypath

shadowhand: That's not the same thing at all. I don't want to use a static class that does way more than an `isset` check.

rdohms: You suggested creating yet another package that does a sub-set of this, i see no real need for all of that if this class does exactly what is needed, but can do more. What's the cost involved in using it?

shadowhand: I'm suggesting creating a package of **traits** that do this and other common operations. Fetching a deep array value by string is not one of the things I would want to implement.

The given package doesn't link to any build status or code quality or even say what PSR standards it uses. I'd like to make sure the code is high quality.

rdohms: > Fetching a deep array value by string is not one of the things I would want to implement.

`return (isset($this->scopes[$identifier])) ? $this->scopes[$identifier] : null;`
becomes
`return a::get($identifier, $this->scopes, null)`

Solves the problem and you don't have to do anything you don't want to

> The given package doesn't link to any build status or code quality or what PSR standards it uses or even contain a license declaration. I'd like to make sure the code is high quality.

That is bad attitude, all of this is solvable by PRs and contributions and keeping the spirit of OSS. 

Anyway, what i'm saying is you are trying to solve a problem that at least 5 packages already do, let's fight NIH and share more code around. That's supposed to be the spirit of the League.

src/Grant/PasswordGrant.php
alexbilbie: See for original intention for events here http://oauth2.thephpleague.com/authorization-server/events/

I've not looked at these again yet. I'm still in the refactoring to the new setup phase.

examples/src/Repositories/ScopeRepository.php
```diff
@@ -0,0 +1,38 @@
...
+        $scopes = [
+            'basic' => [
+                'description' => 'Basic details about you'
+            ],
+            'email' => [
+                'description' => 'Your email address'
+            ]
+        ];
+
+        if (array_key_exists($scopeIdentifier, $scopes) === false) {
```
alexbilbie: It doesn't matter for now, it's just a quick and dirty implementation to help me out for now. Future examples will use a DB as the storage medium.

composer.json
alexbilbie: See https://github.com/firebase/php-jwt#200--2015-04-01

src/Entities/Traits/TokenEntityTrait.php
shadowhand: Show me 5 packages that provide **traits** to implement common array operations. I'd love to know what they are, but every time I go looking I find stuff like ArrayPath that I don't want to use.

src/Grant/AbstractGrant.php
rdohms: The usual rule of thumb i have seen is: dependencies added via constructor are "required" and the ones added via setter injection are optional, makes it easier for whomever is providing objects to inject to know what he needs to provide. This case i think they are all required.

src/Entities/Traits/TokenEntityTrait.php
rdohms: What is the advantage of using a trait here? Its adding functionality that the object does not need, making every class a little bigger at runtime with no real benefit. This is not a trait of an object, even the name does not describe a behaviour of the object, its describing a tool used to simplify an use case, hence its better as an external or even better a language construct.

src/Grant/AbstractGrant.php
shadowhand: That's one pattern. I've also seen (and prefer) this pattern:

```php
function setCollaborator($collab) { ... }
function getCollaborator()
{
    if (empty($this->collab)) {
        throw new RuntimeException('Collaborator must be defined');
    }
    return $this->collab;
}
```

This can even be completed with a trait:

```php
trait EnsureCollaborator
{
    protected function getCollaboratorOrFail($collaborator, $errorMessage)
    {
        if (empty($collaborator)) {
            throw new RuntimeException($errorMessage);
        }
        return $collaborator;
    }
}
```

Implements as:

```php
return $this->getCollaboratorOrFail($this->collab, 'Collaborator is required');
```

Following this pattern, you end up with much smaller units for testing, and everything becomes very consistent in the internal API.

src/Entities/Traits/TokenEntityTrait.php
shadowhand: > no real benefit.

We obviously have very different definitions of "benefit". Used properly, traits are reusable fragments that provide utility to a class without having to repeat the same code over and over, in a **much more flexible way than statics**.

composer.json
shadowhand: :grinning: awesome, I was not aware they had actually updated the package.

src/Entities/Traits/TokenEntityTrait.php
rdohms: why would you say a trait is more flexible? I would disagree, both are as flexible as the other. Also solving the same code over and over is the same purpose of this object.

I would argue traits have the extra onus of risk of nameclashes and harder testing the testing a implementation like ArrayPath. Statics are a valid solution for this problem as it requires no context to be shared, is a clear operation with in/out and no external requirements.

shadowhand: Trait can be overloaded, static class cannot.

rdohms: This hardly seems like a use case for overloading.

src/Grant/AbstractGrant.php
rdohms: This would not be caught at instantiation, only during runtime when you actually use the object. So essentially an error like this would most likely be caught in production, unless you have proper tests that stress every angle of the code and catch this execution path. 
Constructor injection will break as soon as the DIC get's put together, meaning you are more likely to catch the error before it ships out.

shadowhand: > an error like this would most likely be caught in production

Good thing this package ships with a complete test suite. ;) Again, there are two parts to this:

1. Having setters for every collaborator
2. Choosing to use a constructor

My primary goal is that the first be adopted, regardless of the second.

rdohms: That "test suite" makes actually no difference, since this would be an implementation error, so its the users who will suffer from poor architecture, not the library itself.

shadowhand: > suffer from poor architecture, not the library itself

That sounds like an argument in favor of removing the constructor. ;) Kidding aside, I do see your point, and would iterate that there are two separate issues here and one does not preclude the second.

composer.json
lcobucci: You can look http://github.com/lcobucci/jwt too.

shadowhand: :+1: this looks great!

alexbilbie: Only thing against https://github.com/lcobucci/jwt that I can see without playing with it is that it requires PHP 5.5

Then again if you're thinking about security (hence installing this library) then you should be running the latest version of PHP

shadowhand: @alexbilbie I don't see a strong reason to depend on it if oauth-server is going to continue supporting 5.4, but I'll certainly be using it in my own projects.

examples/public/client_credentials.php
spronkey: Seconded. Magic strings are nasty, mmkay.

src/AbstractServer.php
```diff
@@ -167,191 +80,28 @@ public function getRequest()
...
-        return $this->macStorage;
-    }
-
-    /**
-     * @param MacTokenInterface $macStorage
-     */
-    public function setMacStorage(MacTokenInterface $macStorage)
-    {
-        $this->macStorage = $macStorage;
+        switch ($repository) {
```
spronkey: Nasty. Why not use interface name for the container, and get rid of the switch statement somehow?

src/Grant/AuthCodeGrant.php
spronkey: In addition to @shadowhand 's commend above, I think this DI for the grants isn't the best design. It means we're forced to use string identification when adding grants to an authorization server.

How about instead of DIing in the grants, adding a grant similar to the old method, $server->addGrant(new SomeGrant()), and have the server inject the storages needed into the grant itself via DI methods?

src/Grant/AbstractGrant.php
eidng8: I would keep the constructor, because what it does seems fundamental to me. However, I would also add parameters (default to null) to the constructor to ease the pain while inject/test-ing the class. So if the constructor got null parameters, it would do as usual, and while there are something passed in, it'll use the injected stuff.

shadowhand: I am going to contradict my previous opinion and say that setters is the wrong way to go. What we did in oauth2-client was use this signature:

```php
public function __construct(array $options = [], array $collaborators = []) { ... }
```

This is simple and allows for easy customization of the collaborators with dependency injection without having to worry about argument order or setting defaults.

eidng8: Just a few of my own feelings on this kind of signature.

1. While the parameter list is short, I tend to put stuff directly to parameter list. It's easier to document, allow compile time (or sort of) checking, and let modern IDE to do their hinting works.
2. While the parameter list is long, I tend to put stuff in arrays. However, this is harder to document, and cannot be linting, and IDE can't get a hint on it (at least not until phpdoc supports detailed array element documenting).

Sometimes (if not many), I sway. I has particular concern on coding conventions, so the above 2 may not co-exist in the same code base. But each of them has their particular temptations. Such as, for instance, the signature we are talking about.

For this instance, I backed @shadowhand 's opinion above. As it enables enforcement of all fundamental requirements shall be met, right at the moment of creating instance, by checking passed-in values at run time.

shadowhand: I think there is value in understanding the context in which these libraries are used too. While many applications do use auto-resolving injectors, some do not. Having `array $options = [], array $collaborators = []` is very slightly less convenient for auto-resolving injectors but much better for using less sophisticated injector or no injector at all.

PR #335, opened 2015-04-22
New Invalid Token exception.

Irto:
Adding **invalid_token** error type. I think that it missing in v4 too, but I don't know if it's dangerous for existings apps over v4.

https://tools.ietf.org/html/rfc6750#section-3.1
https://tools.ietf.org/html/rfc6750#section-6.2.2

### DIFF
--- a/src/Exception/InvalidTokenException.php
+++ b/src/Exception/InvalidTokenException.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * OAuth 2.0 Invalid Token Exception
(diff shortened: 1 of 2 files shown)


starJammer:
Hey Irto, I created this pull request and realized it's very similar to yours. Can you look at mine as well. I added tests as well as some modifications for properly including the WWW-Authenticate headers.

https://github.com/thephpleague/oauth2-server/pull/399

PR #363, opened 2015-08-20
Patch for auth code client id check.

cesutherland:
* add check for client id against the client id of the auth code's owner
* remove client session association (should be associated).

Access token requests's client ID wasn't being checked against the
original client ID of the auth code.  I've added this check.  I don't
think the old session association did anything useful.

http://tools.ietf.org/html/rfc6749#section-4.1.3
> ensure that the authorization code was issued to the authenticated
> confidential client, or if the client is public, ensure that the
> code was issued to "client_id" in the request,

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -227,8 +227,11 @@ class AuthCodeGrant extends AbstractGrant
             throw new Exception\InvalidRequestException('redirect_uri');
         }
 
+        // Check client ID presented matches client ID originally used in authorize request
         $session = $code->getSession();
-        $session->associateClient($client);
+        if ($session->getClient()->getId() !== $client->getId()) {
+            throw new Exception\InvalidRequestException('code');
+        }
 
         $authCodeScopes = $code->getScopes();
 
--- a/tests/unit/Grant/AuthCodeGrantTest.php
+++ b/tests/unit/Grant/AuthCodeGrantTest.php
@@ -546,6 +546,63 @@ class AuthCodeGrantTest extends \PHPUnit_Framework_TestCase
         $server->issueAccessToken();
     }
 
+    public function testCompleteFlowRedirectClientIdMismatch()
+    {
+        $this->setExpectedException('League\OAuth2\Server\Exception\InvalidRequestException');
+
+        $_POST = [
+            'grant_type'    =>  'authorization_code',
+            'client_id'     =>  'failface',
+            'client_secret' =>  'foobar',
+            'redirect_uri'  =>  'http://foo/bar',
+            'code'          =>  'foobar',
+        ];
+
+        $server = new AuthorizationServer();
+        $grant = new AuthCodeGrant();
+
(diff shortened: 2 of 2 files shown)


cesutherland:
This prevents one client from being granted an access token for an auth code issued to another client.

The tests on `develop` weren't passing, so I created this PR from `master` (and assigned the PR to `master`).  If I should assign it to `develop` let me know!

apollopy:
I am late, I have same idea about this. but if checked, can associate $client to $session.

```php
        // Check client ID presented matches client ID originally used in authorize request
        $session = $code->getSession();
        if ($session->getClient()->getId() != $client->getId()) {
            throw new Exception\InvalidRequestException('code');
        }
        $session->associateClient($client); // It is not necessary to delete the line
```

alexbilbie:
@cesutherland thanks for this.

Why did you remove the `$session->associateClient($client);` line?

dekker-m:
@alexbilbie I think he removed this line so the the client from the session is validated against the client making the request. This way, only the client who requested the AuthCode can request the Access Token.

PS: I also mailed you about the same issue a few days ago ;-)

apollopy:
@dekker-m 
I think check is need, but if checked, can associate $client to $session.

dekker-m:
@apollopy If you check the client against the client who requested the authorization code, associating the $client to $session is not necessary. They will be the same if the check passes and the client should already associated to the session in the storage.

cesutherland:
@alexbilbie @dekker-m @apollopy it should already be associated.  When the auth code is originally created, the client is associated with it.  That's how we're able to recall the client to make the check before issuing the access token :)

https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php#L148-L165

apollopy:
@cesutherland 
sorry, my english is bad.

You said associated to database; I said associated to current object. only in this way, the code is consistent.

can see:
https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php#L160
https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php#L241

cesutherland:
@apollopy Thanks, I see!  I think it's OK, in any case.  The client was already created, associated and saved as part of the session in the authorization request.  `$session->getClient()` simply recalls the stored client and sets it on the session.  I think that's what we want here.

https://github.com/thephpleague/oauth2-server/blob/c174b6fc65f034dd6547da6ab7bb429a20988eb0/src/Entity/SessionEntity.php#L225-L246

It looks like [this lookup](https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php#L196-L207) might be removable now.  We can compare the `$clientId` and `$clientSecret` to that already stored on the session.  The `$redirectUri` is already compared to the auth code and was validated when the auth code was issued.  And we looked at the grant id in... the server I think?

I'm starting to agree with #306... the session abstraction — which is very neat — really muddles some things.

apollopy:
@cesutherland 
I simply think remove the `$session->associateClient($client);` line don't help fix this bug. Just keep the same code style

PR #364, opened 2015-08-21
Too idealistic. Should allow the client and server have some time difference.

apollopy:


michaelhogg:
+1

It's definitely unreasonable to expect the timestamp to be identical to `time()`. This is a good improvement. It would be even better if the allowable time difference could be configured, instead of being hard-coded.

I'm not sure how closely this library is intending to follow the [OAuth 2.0 Message Authentication Code (MAC) Tokens](https://tools.ietf.org/html/draft-ietf-oauth-v2-http-mac-05) draft spec, but the [Timestamp Verification](https://tools.ietf.org/html/draft-ietf-oauth-v2-http-mac-05#section-6.1) section suggests calculating and storing the time delta of each client, so that clients aren't required to have accurate clocks.

That section of the draft spec also suggests a value of 5 minutes for the "allowable clock skew".

src/TokenType/MAC.php
```diff
@@ -84,7 +84,7 @@ public function determineAccessTokenInHeader(Request $request)
             return;
         }
 
-        if ((int) $params->get('ts') !== time()) {
+        if (abs($params->get('ts') - time()) > 3600) {
```
alexbilbie: 3600 seconds is excessive. I'd change it to 300 to allow for 5 minutes instead

dekker-m:
I'm sure which version of the draft for MAC is implemented, but v2 of the draft has a pretty way of checking the timestamp. On the first request the time difference is measured between server and client, this difference should be saved to the database. Subsequent requests should/can be checked against a much smaller time difference, possibly even a few seconds.

Checking the timestamp should prevent replay attacks. Even a window of 5 minutes, in my opinion, is way to much to prevent replay attacks. The proposed way of checking the ts parameter in the v2 of the draft is a pretty good way to keep the window as small as possible.

apollopy:
if want to prevent replay attacks, can save `ts`+`nonce` to check uniqueness. but need more code. 

Anyway, `((int) $params->get('ts') !== time())`  too idealistic

dekker-m:
Saving the ts + nonce is indeed another good way to prevent replay attacks. Nonetheless, I think performaning the check as proposed by @michaelhogg (as defined in the RFC) would be a good improvement over the current check. There is always the possibilty two clients do not have the correct, current timestamp. By checking the timestamp as proposed in the RFC, the difference does not mather, is should even be possible to have a time difference of 3 hours. The check will still make sure every request after the first one fall within a window of few seconds.

starJammer:
+1

alexbilbie:
Merging this for now so that at least the more reasonable time comparison is in place.

Regarding nonce replay validation and checking time differences in subsequent requests that can implemented later on.

PR #365, opened 2015-08-21
[REVIEW] Add interfaces for authorization server

arnovr:
We needed to have interfaces for the authorization server, because we wanted some specific implementation.
@alexbilbie


alexbilbie:
I'm not going to merge because I've got an alternative interface implementation which I want to use instead. But thank you for the contribution

arnovr:
I am not seeing any implementation of a Server Interface anywhere @ master.
Could you enlighten me for your solution. We are having some issues on circular dependencies. This gives us more flexibility.

PR #377, opened 2015-09-04
AuthCodeGrant and RefreshTokenGrant don't require client_secret

starJammer:
Hey,

I started to implement OAuth2 in my workplace using this library. We have a requirement to NOT store client secrets in native applications. The RFC suggests as much in section https://tools.ietf.org/html/rfc6749#section-2.1 where it says public clients are incapable of keeping their credentials secure. 

From my understanding, the spec allows client_ids to be created with no client_secret, which is our use case at the moment. 

I added a flag to the authcode grant and the refreshtoken grants to reflect this. It was a quick hack so why not?

This leads to maybe some confusion with `ClientInterface->get` because even if the passed in client_secret is null, the implementation should return null if the client_id does have a client_secret associated with it.

Well, let me know what you think. I plan on using this since this is my use case right now. I like that your library is flexible in some areas and actually has a MAC implementation.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -61,6 +61,14 @@ class AuthCodeGrant extends AbstractGrant
     protected $authTokenTTL = 600;
 
     /**
+     * Whether to require the client secret when
+     * completing the flow.
+     *
+     * @var boolean
+     */
+    protected $requireClientSecret = true;
+
+    /**
      * Override the default access token expire time
      *
      * @param int $authTokenTTL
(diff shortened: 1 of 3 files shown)


alexbilbie:
Seems perfectly reasonable to me. Thanks.

eidng8:
Not sure if I get it right. Should we skip an extraneous by making this change?

```php
$clientSecret = null;
if ($this->shouldRequireClientSecret()) {
    $clientSecret = $this->server->getRequest()->request->get('client_secret',
        $this->server->getRequest()->getPassword())
    if(is_null($clientSecret)) {
        throw new Exception\InvalidRequestException('client_secret');
    }
}
```

alexbilbie:
Hi,

What you're trying to achieve is already included in the library - https://oauth2.thephpleague.com/client-repository-interface/

> The client secret may or may not be provided depending on the request sent by the client. The boolean $mustValidateSecret parameter will indicate whether or not the client secret must be validated. If the client is confidential (i.e. is capable of securely storing a secret) and $mustValidateSecret === true then the secret must be validated.

So in your implementation of `getClientEntity` you determine the client secret doesn't need validating then just validate the client identifier.

PR #381, opened 2015-09-23
Fixed problem with empty Authorization header

fabik:
On my machine the variable `$_SERVER['HTTP_AUTHORIZATION']` is set to an empty string when no Authorization header is present. As a result the detection of access token does not work properly.

I don't really know why the `$_SERVER['HTTP_AUTHORIZATION']` is set to an empty string. I'm using Mac OS X with Apache 2.2 and PHP 5.6 installed using homebrew.


paragonie-scott:
Scrutinizer was cancelled? o_o

alexbilbie:
Have you tried adding `SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1` into your `.htaccess` file?

PR #388, opened 2015-10-13
Added priority argument

m4tthumphrey:
Added the option to pass the event listener priority when adding an event listener to a server.


alexbilbie:
Thanks @m4tthumphrey

PR #391, opened 2015-10-23
Remove client secret check for PasswordGrant

saga64:
In terms of https://tools.ietf.org/html/rfc6749, since v2-16,
PasswordGrant and RefreshGrant are no longer require client secret validation
by default. Pleaseread section 4.3.1 and section 6.

This commit will remove client secret validation and remove the
correspond test.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-	"name": "saga64/oauth2-server",
+	"name": "league/oauth2-server",
 	"description": "A lightweight and powerful OAuth 2.0 authorization and resource server library with support for all the core specification grants. This library will allow you to secure your API with OAuth and allow your applications users to approve apps that want to access their data from your API.",
 	"homepage": "http://oauth2.thephpleague.com/",
 	"license": "MIT",
@@ -12,6 +12,12 @@
 		"phpunit/phpunit": "4.3.*",
 		"mockery/mockery": "0.9.*"
 	},
+	"repositories": [
+		{
+			"type": "git",
+			"url": "https://github.com/thephpleague/oauth2-server.git"
+		}
+	],
 	"keywords": [
 		"oauth",
 		"oauth2",
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-	"name": "league/oauth2-server",
+	"name": "saga64/oauth2-server",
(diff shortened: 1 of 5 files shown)


mschinis:
Can we merge this?

alexbilbie:
I don't understand why you wouldn't want confidential clients?

mschinis:
Clients should never expose private keys. It's not a matter of not wanting confidential clients, but more of a "keeping up to date with the latest standards". Client implementations like [ember-simple-auth](https://github.com/simplabs/ember-simple-auth), have stopped supporting earlier versions of the oauth2 standard, so I believe this PR would help a lot of other users.

whitefoxy:
+1 Waiting for this merge

alexbilbie:
But OAuth 2 is the latest standard...

This library has always tried to promote a secure by default implementation
of OAuth 2 and I think it's irresponsible to support non-confidential
clients.

If you want to support them then you can create your own variation of the
password grant very easily that supports your specific use case.

whitefoxy:
I don't understand why in PasswordGrant `client_secret` adds more security? It just adds more problems to store them. How the PasswordGrant can be used in desktop or mobile apps? The request must be proxied by the server and it is absolutely the same if the PasswordGrant will not require the `client_secret`. In many docs, for example on IBM website, they don't use client_secret in PasswordGrant.

mschinis:
@alexbilbie Most implementations of oauth2, including client oauth2 libraries follow the [RFC6749](https://tools.ietf.org/html/rfc6749).

Sharing the `client_secret` doesn't add any extra security, as mentioned by @whitefoxy, since the client has to expose its secret publicly at some point. Wether that is through the js console in web apps, or taking a mobile application apart and looking through the configuration files.

heisian:
I agree, but would actually prefer the `client_secret` to pass a hash check such that an HMAC-signed version of the `client_secret` is checked on the server.  That way `client_secret` needs not be passed.  I would say this is important for the `client_credentials` grant as well.

alexbilbie:
@heisian what is the updated standard? RFC6749 was published in October 2012

alexbilbie:
I've read through https://tools.ietf.org/html/rfc6819 and I will merge this PR but after I've updated the documentation as part of V5 launch which will better describe appropriate implementations depending on requirements of the authorization server

alexbilbie:
V5 now supports both public and confidential clients as of 3b36ae9000328ccba408b20752021cfbc51e7a6d

mschinis:
thanks!

PR #399, opened 2015-11-13
Better www-authenticate headers inclusion

starJammer:
Usage of the Bearer token specifies that a `www-authenticate` header should be returned when there is a token problem when accessing resources.

The changes I made here include:

1. Adding two new exceptions
    1. `NoTokenException`
    2. `InvalidTokenException`
2. Altering the `getHttpHeaders` function in the OAuthException class to not include anything related to the www-authenticate header
3. Using `Request::createFromGlobals()` when creating the request object in the exception to properly fetch the `Authorization` header from the request. Before the request was created using `$request = new Request();` and it resulted in the `$request` object not being populated with anything.
4. Changing the resource server to use these two exceptions when validating the token.
5. Edited EXISTING tests that expected `InvalidRequestException`s to expect one of the two new exceptions.
6. Added new `getTypeName` method to `TokenTypeInterface` so we have a way of getting the type of token being used. This is used with the `NoTokenException`.

What I DIDN'T do

1. Add new unit tests for the new exception classes.

These changes aren't "complete" by any means but they help conform a bit more to the recommended usage of the Bearer token.

Any suggestions are welcome.

### DIFF
--- a/src/Exception/InvalidTokenException.php
+++ b/src/Exception/InvalidTokenException.php
@@ -60,7 +60,7 @@ class InvalidTokenException extends OAuthException
 
         switch ($authScheme) {
             case 'Bearer':
-                $headers[] = 'WWW-Authenticate: ' . $authScheme . 
+                $headers[] = 'WWW-Authenticate: ' . $authScheme . ' ' .
                     'error = "invalid_token", ' . 
                     'error_description="There is no access token present."';
             break;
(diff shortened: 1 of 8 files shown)


starJammer:
awww, code coverage failed. It says the external service took too long. What do I do?

starJammer:
Updated branch to latest master.

PR #404, opened 2015-11-23
pass ResponseInterface to be handled

juliangut:
Apart from allowing the library to create Zend\Diactoros\Response when needed, passing an already PSR7 response object allows oauth2-server to be leveraged at the end of the middleware chain, it might even use `__invoke` instead of `handleTokenResponse`

With this `generateHttpResponse` doesn't really has a meaning so I've changed it to `getResponse($response)`

There is a question I have, why using `League\OAuth2\Server\Exception\OAuthServerException` as a factory instead of a collection of Exceptions?

### DIFF
--- a/src/Repositories/ClientRepositoryInterface.php
+++ b/src/Repositories/ClientRepositoryInterface.php
@@ -26,5 +26,5 @@ interface ClientRepositoryInterface extends RepositoryInterface
      *
      * @return \League\OAuth2\Server\Entities\Interfaces\ClientEntityInterface
      */
-    public function getClientEntity($clientIdentifier, $grantType, $clientSecret = null, $redirectUri = null);
+    public function getClientEntity($clientIdentifier, $clientSecret = null, $redirectUri = null, $grantType = null);
 }
--- a/src/Repositories/ClientRepositoryInterface.php
+++ b/src/Repositories/ClientRepositoryInterface.php
@@ -20,11 +20,11 @@ interface ClientRepositoryInterface extends RepositoryInterface
      * Get a client
      *
      * @param string      $clientIdentifier The client's identifier
+     * @param string      $grantType        The grant type used
      * @param string|null $clientSecret     The client's secret
      * @param string|null $redirectUri      The client's redirect URI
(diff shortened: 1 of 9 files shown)


src/Repositories/ClientRepositoryInterface.php
```diff
@@ -26,5 +26,5 @@
      *
      * @return \League\OAuth2\Server\Entities\Interfaces\ClientEntityInterface
      */
-    public function getClientEntity($clientIdentifier, $grantType, $clientSecret = null, $redirectUri = null);
+    public function getClientEntity($clientIdentifier, $clientSecret = null, $redirectUri = null, $grantType = null);
```
alexbilbie: Why was this change made?

The grant should always be injected in so the implementation can decide if the client is permitted to use the grant type

alexbilbie:
Regarding your comment about exceptions - see http://rosstuck.com/formatting-exception-messages/

I've not finished exceptions yet; I still need to finish how translations are going to work

src/Repositories/ClientRepositoryInterface.php
juliangut: I has a look at `OAuth2ServerExamples\Repositories\ClientRepository` and the phpdoc on ClientRepositoryInterface and noticed the order of the paremeters was shifted, you can see grantType is at the end of the list

alexbilbie: `$grantType` will always be set though and as per [PSR2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#44-method-arguments) "Method arguments with default values MUST go at the end of the argument list."

juliangut: OK then, I'm reverting that piece of code, updating the docblock and changing `OAuth2ServerExamples\Repositories\ClientRepository`

juliangut:
I really need this to be PSR7 compliant so I'll be making some PR in the next days, do you have a roadmap to V5? anyway it can be tracked so I can help?

alexbilbie:
Closing as I believe I've addressed this

PR #409, opened 2015-12-03
Rename phpunit.xml file

vinkla:


alexbilbie:
Why?

vinkla:
This will allow developers to keep a custom phpunit.xml for development.

alexbilbie:
As it's such a barebones `phpunit.xml` file in the first place I'm not against others adding to it in a PR if they have useful improvements

vinkla:
It is just a way to let developers customize the settings in their local environment. It is common to use .dist(ribution) when pushing packages to GitHub. Nothing important really :) Just a best practice.

GawainLynch:
Just a note in support of this PR. PHPUnit will check for the `.xml.dist` file if the `.xml` isn't present, so there is no negative impact on the project… but it can make short term testing changes a lot easier for a developer.

vinkla:
:+1:

PR #412, opened 2015-12-11
Allow Symfony 3.0

derrabus:
Symfony 3.0 has been released and I'd like to migrate my Symfony projects to that new release. Your library is currently blocking the migration because its composer.json file defines a dependency against Symfony HTTP Foundation 2.

As far as I can tell, the code should work just fine with Symfony HTTP Foundation 3, so I've marked the 3.x series as compatible in this PR.


vinkla:
:+1:

PR #422, opened 2016-01-15
V5 - allow middleware use

juliangut:
Pass Response object to `respondToRequest` to be composed with result of auth check instead of creating a new Response
`__invoke` method to allow to be used as middleware

### DIFF
--- a/src/Exception/OAuthServerException.php
+++ b/src/Exception/OAuthServerException.php
@@ -204,8 +204,12 @@ class OAuthServerException extends \Exception
      *
      * @return ResponseInterface
      */
-    public function generateHttpResponse()
+    public function generateHttpResponse(ResponseInterface $response = null)
     {
+        if (!$response instanceof ResponseInterface) {
+            $response = new Response();
+        }
+
         $headers = $this->getHttpHeaders();
 
         $payload = [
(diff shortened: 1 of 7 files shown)


src/Exception/OAuthServerException.php
```diff
@@ -221,12 +225,13 @@ public function generateHttpResponse()
...
         }
 
-        $response = new Response(
-            'php://memory',
-            $this->getHttpStatusCode(),
-            $headers
-        );
-        $response->getBody()->write(json_encode($payload));
+        foreach ($headers as $header => $content) {
+            $response->withHeader($header, $content);
```
andersonamuller: You forgot the immutability: 

``` php
$response = $response->withHeader($header, $content);
```

src/Exception/OAuthServerException.php
```diff
@@ -221,12 +225,13 @@ public function generateHttpResponse()
...
+        }
+
+        $response
```
andersonamuller: You forgot the immutability: 

``` php
$response = $response
              ->withStatus($this->getHttpStatusCode())
              ->getBody()->write(json_encode($payload));
```

or 

```
return $response
          ->withStatus($this->getHttpStatusCode())
          ->getBody()->write(json_encode($payload));
```

alexbilbie: Good spot!

alexbilbie:
Thanks @juliangut, will fix the immutability issues highlighted by @andersonamuller

PR #425, opened 2016-01-17
V5 - normalize validatescopes

juliangut:
Normalize `validateScopes` to the way it's done in `validateClient`, `validateUser` and `validateOldRefreshToken`: extraction and validation all together


alexbilbie:
Bear with me a moment, my next push is going to break some of this

alexbilbie:
@juliangut I've done a big push which has caused conflicts on this PR

juliangut:
No prob @alexbilbie, conflicts solved

juliangut:
By the way, does oauth2 standard tells scope delimiter is limited to ' ' and thus it's now a constant or is it just because of convenience?. I find it useless to be able to change it, I've never done it

alexbilbie:
It's expected to be space delimited.

[Section 3.3](https://tools.ietf.org/html/rfc6749#section-3.3):

>  The value of the scope parameter is expressed as a list of space-
   delimited, case-sensitive strings.  The strings are defined by the
   authorization server.  If the value contains multiple space-delimited
   strings, their order does not matter, and each string adds an
   additional access range to the requested scope.

PR #431, opened 2016-01-17
V5 - use Psr\Http\Message\UriInterface

juliangut:
Compound redirect URI using already available Diactoros `Psr\Http\Message\UriInterface` implementation


alexbilbie:
RedirectUri is also used by the authorization code grant which I've not yet pushed

juliangut:
Does RedirectUri have any other method but make? if it is not the case you can use the same strategy using Diactoros Uri implementation in AuthCodeGrant

alexbilbie:
Will bear this in mind as I work on auth code

PR #433, opened 2016-01-17
V5 - use phpseclib for encryption/decryption

juliangut:
Make use of [phpseclib](https://github.com/phpseclib/phpseclib/tree/2.0) for encryption/decryption of refresh tokens payload, it allows to change algorithm (symmetric/asymmetric) if needed, using RSA by default

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
         "zendframework/zend-diactoros": "~1.1",
         "namshi/jose": "^6.0",
         "lcobucci/jwt": "^3.1",
-        "paragonie/random_compat": "^1.1"
+        "paragonie/random_compat": "^1.1",
+        "phpseclib/phpseclib": "^2.0"
     },
     "require-dev": {
         "phpunit/phpunit": "4.8.*",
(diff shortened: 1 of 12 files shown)


alexbilbie:
How does this work if the key has a password?

juliangut:
You've come to a good point! This line is needed in decrypt method
```php
self::getAlgorithm()->setPassword($privateKeyPassword);
```
It adds an extra parameter to be handled in case RSA key has a password

It's added logic but I "guess" the benefits of being able to switch algos outruns the drawbacks of having to store and pass an extra parameter, maybe??

juliangut:
Added `privateKeyPassword` parameter.

Parameters on `Server::__construct` and `AbstractResponseType::__construct` needed reordering to allow default empty string for privateKeyPassword

juliangut:
I'm not very fond of this, should probably rewrite this completely. Anyway my goal is to avoid "Helper" classes such as KeyCrypt, IMO such functionality belong in a Trait

juliangut:
Dropping in favor of #468

PR #435, opened 2016-01-19
V5 - Exception based access token check

juliangut:
Access token check by exception handling

* Move `Authorization` header check into `AbstractResponseType`
* Pass exceptions when erroring on determining access token instead of passing request attributes
* Makes `ResourceServerMiddleware` more concise

### DIFF
--- a/src/Middleware/ResourceServerMiddleware.php
+++ b/src/Middleware/ResourceServerMiddleware.php
@@ -38,7 +38,9 @@ class ResourceServerMiddleware
         } catch (OAuthServerException $exception) {
             return $exception->generateHttpResponse($response);
         } catch (\Exception $exception) {
-            return $response->withStatus(500)->write($exception->getMessage());
+            $response->getBody()->write($exception->getMessage());
+
+            return $response->withStatus(500);
         }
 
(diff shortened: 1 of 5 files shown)


src/Middleware/ResourceServerMiddleware.php
```diff
@@ -33,18 +33,12 @@ public function __construct(Server $server)
...
-
-        if ($request->getAttribute('oauth_access_token') === null) {
-            $exception = OAuthServerException::accessDenied($request->getAttribute('oauth_access_token_error'));
-
+        try {
+            $request = $this->server->getResponseType()->determineAccessTokenInHeader($request);
+        } catch (OAuthServerException $exception) {
             return $exception->generateHttpResponse($response);
+        } catch (\Exception $exception) {
+            return $response->withStatus(500)->write($exception->getMessage());
```
andersonamuller: The write method returns the number of written bytes, not the response. You could do:
```php
$response->getBody()->write($exception->getMessage());
return $response->withStatus(500);
```

juliangut: Somehow I keep making this mistake from time to time

PR #437, opened 2016-01-20
V5 - Allow refresh token TTL assign

juliangut:
Allow assign of refresh token TLL along with access token TTL

ResponseType is globally unique (defined in Server), no need to keep an array of references to the same object.

### DIFF
--- a/src/Grant/ClientCredentialsGrant.php
+++ b/src/Grant/ClientCredentialsGrant.php
@@ -32,14 +32,15 @@ class ClientCredentialsGrant extends AbstractGrant
     public function respondToRequest(
         ServerRequestInterface $request,
         ResponseTypeInterface $responseType,
-        \DateInterval $tokenTTL
+        \DateInterval $accessTokenTTL,
+        \DateInterval $refreshTokenTTL
     ) {
         // Validate request
         $client = $this->validateClient($request);
         $scopes = $this->validateScopes($request, $client);
 
         // Issue and persist access token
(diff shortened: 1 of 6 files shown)


src/Grant/ClientCredentialsGrant.php
```diff
@@ -32,14 +32,15 @@ class ClientCredentialsGrant extends AbstractGrant
     public function respondToRequest(
         ServerRequestInterface $request,
         ResponseTypeInterface $responseType,
-        \DateInterval $tokenTTL
+        \DateInterval $accessTokenTTL,
+        \DateInterval $refreshTokenTTL
```
alexbilbie: Client credentials grant doesn't support refresh tokens so this might be confusing

juliangut:
Yes, that's the only point in which it seams odd. The other way to do this is by using `setRefreshTokenTTL` method inside Server::enableGrantType controlled by an `instanceof` check

alexbilbie:
This is the problem I keep hitting up against; different grants have different requirements :stuck_out_tongue_closed_eyes:. The auth code grant is such an ugly beast to try and retrofit into an otherwise nice API.

juliangut:
I can imagine, I've not tried to implement authCode grant because you were already working on it :grinning:

Anyway I think that allowing assing of refresh token TTL instead of hardcoded to 1 month is a must

juliangut:
Just make refresh token TTL configurable on a grant basis

PR #444, opened 2016-02-12
V5 - Secure access to body params

juliangut:
directly accessing $request->getParsedBody() as array is insecure, it can return null|array|object


alexbilbie:
I've just pushed a big set of changes which have caused a conflict here

juliangut:
forced $request->getParsedBody() to array

PR #445, opened 2016-02-12
V5 - Abstract access token validation

juliangut:
Abstract access token validation from request behind a method making getResponseType protected

Extra: set DateInterval to be FQN

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -26,7 +26,7 @@ class Server implements EmitterAwareInterface
     protected $enabledGrantTypes = [];
 
     /**
-     * @var DateInterval[]
+     * @var \DateInterval[]
      */
     protected $grantTypeAccessTokenTTL = [];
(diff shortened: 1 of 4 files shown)


src/Server.php
```diff
@@ -144,11 +144,25 @@ public function respondToRequest(ServerRequestInterface $request = null, Respons
...
     /**
+     * Determine the access token validity
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request
+     *
+     * @return \Psr\Http\Message\ServerRequestInterface
+     *
+     * @throws \League\OAuth2\Server\Exception\OAuthServerException
+     */
+    public function validateRequest(ServerRequestInterface $request)
```
alexbilbie: Wondering if `validateAuthenticatedRequest` might be a better name?

PR #446, opened 2016-02-12
V5 - unify examples

juliangut:
middleware example has both Authentication and Resource to show off how to integrate both services together


juliangut:
protected_api example have been merged into middleware_use

PR #457, opened 2016-03-09
Allow different template engines

juliangut:
I've nothing against Plates as template engine but only allowing one template engine for authorization/login templates is kind of limiting.

Through `\League\OAuth2\Server\TemplateRenderer\RendererInterface` any template engine can be added, bridges for Twig, Smarty, Mustache and Plates (default) have been added.

`AbstractAuthorizeGrant` has been introduced to abstract commonalities between AuthCodeGrant and ImplicitGrant

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,6 +64,9 @@
         }
     },
     "suggest": {
-        "league/plates": "Required for parsing authorization code templates"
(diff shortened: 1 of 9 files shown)


alexbilbie:
This is awesome; thank you!

I will have a good look at it shortlty

PR #468, opened 2016-03-15
V5 - Move cryptography handling to trait

juliangut:
* Removed `pathTo*Key` attributes and methods, they have been normalized to `*KeyPath` as they were used in Server::__construct
* private and public key paths are kept in CryptTrait
* Instead of adding private and public keys on responseType creation Server sets them in Server::getResponseType
* key paths are tested to start with ' file://' and prepended if needed
* test suit updated accordingly :blush:


paragonie-scott:
https://github.com/juliangut/oauth2-server/blob/cryptography_trait/src/CryptTrait.php#L74

If you don't specify the padding mode for RSA, you end up in the same boat as [ZF2015-10](http://framework.zend.com/security/advisory/ZF2015-10).

juliangut:
Thanks for the advice @paragonie-scott, updated to use OPENSSL_PKCS1_OAEP_PADDING

alexbilbie:
@juliangut mind having a look to see what is causing tests to fail please?

alexbilbie:
@juliangut can you pull from upstream which should resolve merge + test errors

juliangut:
I see the problem:

[private_encrypt](http://php.net/manual/en/function.openssl-private-encrypt.php) and [public_decrypt](http://php.net/manual/en/function.openssl-public-decrypt.php) only accept `OPENSSL_PKCS1_PADDING`or `OPENSSL_NO_PADDING` as padding param, while [public_encrypt](http://php.net/manual/en/function.openssl-public-encrypt.php) and [private_decrypt](http://php.net/manual/en/function.openssl-private-decrypt.php) do accept more padding options, including `OPENSSL_PKCS1_OAEP_PADDING`

Now that I write it down it seems the encryption process is upside-down, it should be encrypted with public key so only the server will be able to decrypt it with private key.

I'll wait for your comment @alexbilbie to make the change. May we have a word from @paragonie-scott  as well?

alexbilbie:
My thinking is that the authorisation server signs things with it's private key.

The resource server and anything else that needs to validate access tokens can use the public key to verify the integrity of the token.

alexbilbie:
@juliangut I've implemented this manually along with #473 but predominately used your code. Thanks again!

juliangut:
I'm afraid test are failing bad

Issue #469, opened 2016-03-15
V5 - Default templates


juliangut:
Shouldn't default templates be moved to TemplateRenderer maybe? seems odd inside ResponseTypes directory

alexbilbie:
Fixed by #472

PR #472, opened 2016-03-16
V5 - Template renderer holds template related information

juliangut:
Template related information belong in renderer, also grants get cleaner.
Additionally addressess #469


alexbilbie:
I like all of this. Thanks!

PR #473, opened 2016-03-16
[V5] Remove Zend Diactoros dependency; depend on PSR-7

frederikbosch:
Why not rely on PSR-7 only? With Zend Diactoros as non-dev dependency you will limit many people to a specific version of that package.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,13 +6,14 @@
     "require": {
         "php": ">=5.5.9",
         "league/event": "~2.1",
-        "zendframework/zend-diactoros": "~1.1",
+        "psr/http-message": "~1.0",
         "lcobucci/jwt": "^3.1",
         "paragonie/random_compat": "^1.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8",
-        "league/plates": "^3.1"
+        "league/plates": "^3.1",
+        "zendframework/zend-diactoros": "~1.1"
     },
     "repositories": [
         {


frederikbosch:
Forgot to add how great your work is on this new version 5!

alexbilbie:
The library actually uses Diactoros internally currently so I will need to do some refactoring first before I can remove the dependency

frederikbosch:
@alexbilbie You want me to help on this? I can update the PR by replacing Diactoros completely for PSR-7. Or would you rather focus on it yourself?

alexbilbie:
@frederikbosch it won't take me too long to do it, thanks for the offer though

alexbilbie:
Hmm having had a look at this it's going to require a lot more refactoring than I expected.

The problem (as ever) is the auth code grant which generates HTML, JSON and redirect responses. It also uses Diactoros' `Psr\Http\Message\UriInterface` implementation.

alexbilbie:
Also the `OAuthServerException` class uses `UriInterface`

alexbilbie:
The `composer.json` is specifying `"zendframework/zend-diactoros": "^1.1"` which will anything >= 1.1 && <2.0 so this won't lock anyone to an absolute version of the library.

I'm thinking for the sake of getting v5 out of the door to resolve this in v5.1 if there is demand for it

frederikbosch:
@alexbilbie Alright, thanks for your explanation. To be honest: I think v5 *must* lean on PSR-7 instead of Diactoros. It would defeat the purpose of a PSR if packages would still lean on concretions rather than interfaces. Therefore I think I am going to have a look at it. Maybe turn the Diactoros stuff into an adapter. Will update the PR when I am done.

alexbilbie:
I can solve the PSR-7 Request and Responses more easily than the UriInterface

alexbilbie:
This is resolved

frederikbosch:
Oh wow, fantastic! I am very grateful for resolving it so fast. Wonderful. Many thanks!

PR #489, opened 2016-03-24
remove unnecessary methods from interfaces

frederikbosch:
The interfaces should only contain methods that are required because the server consumes those methods. A few of them were not used, and could therefore be removed,


frederikbosch:
General comment: I find that the interfaces are containers, which is understandable in scope of this library package. But maybe there are situations where the name of the method could change from a getter/setter to a more operational name (tell instead of ask).

alexbilbie:
I agree with your proposed changes.

Could you please pull from HEAD so I can merge.

PR #491, opened 2016-03-24
less usage of null, throw exception when login fails in user repo

frederikbosch:
Less `instanceof`, less `null`, throw exception when cannot login

### DIFF
--- a/src/Exception/UnknownUserException.php
+++ b/src/Exception/UnknownUserException.php
@@ -1,0 +1,6 @@
+<?php
+namespace League\OAuth2\Server\Exception;
+
+class UnknownUserException extends \Exception
+{
+}
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -4,8 +4,8 @@ namespace League\OAuth2\Server\Grant;
 
 use DateInterval;
 use League\OAuth2\Server\Entities\Interfaces\ClientEntityInterface;
-use League\OAuth2\Server\Entities\Interfaces\UserEntityInterface;
 use League\OAuth2\Server\Exception\OAuthServerException;
+use League\OAuth2\Server\Exception\UnknownUserException;
 use League\OAuth2\Server\Repositories\AuthCodeRepositoryInterface;
 use League\OAuth2\Server\Repositories\RefreshTokenRepositoryInterface;
 use League\OAuth2\Server\Repositories\UserRepositoryInterface;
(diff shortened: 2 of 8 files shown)


alexbilbie:
I'm not sure about this. The advantage of checking that the response is an implementation of an interface is that anything other than an implementation will result in a failure.

Asking client implementations to throw library specific exceptions seems like more of a barrier for implementers in my opinion.

Am open to other peoples' opinions though

frederikbosch:
@alexbilbie An exception creates a better contract. When I was implementing this, I was wondering: what should I do when the login is incorrect. The doc block did not contain any hint what to do. And, alright, it's a WIP, but let's consider there would have been a docblock. See below.

```php
/**
 * @return UserEntityInterface|null Return the user entity when correct login, otherwise null
 */
```

I believe this designs a better contract.
```php
/**
 * @return UserEntityInterface Return the user entity after logged in
 *
 * @throws UnknownUserException in case of an unknown user
 */
```

You even have the possibility to create more exceptions in case of other failures. More important, it eliminates a bunch of null checks. Look how many null checks [this block](https://github.com/thephpleague/oauth2-server/blob/V5-WIP/src/Grant/AuthCodeGrant.php#L57) contains.

frederikbosch:
I'd like to add something more here. What would you do if this library would only support PHP7? Then I am sure you would like to use return types, right? But then you cannot even use return null. See the [paragraph "Disallowing NULL on Return Types" in the RFC for more details](https://wiki.php.net/rfc/return_types) on this. So one has to use an exception when it is not possible to return the required type. The signature of the method for PHP would then had to look similar to this.

```php
    /**
     * Get a user entity.
     *
     * @param string                                                          $username
     * @param string                                                          $password
     * @param string                                                          $grantType    The grant type used
     * @param \League\OAuth2\Server\Entities\Interfaces\ClientEntityInterface $clientEntity
     *
     * @throws \League\OAuth2\Server\Exception\UnknownUserException unknown user
     * @throws \League\OAuth2\Server\Exception\InvalidUserCredentials cannot authenticate
     *
     * @return \League\OAuth2\Server\Entities\Interfaces\UserEntityInterface
     */
    public function getUserEntityByUserCredentials(
        $username,
        $password,
        $grantType,
        ClientEntityInterface $clientEntity
    ) : UserEntityInterface;
```

Therefore, I would like to propose not only to change the `UserRepositoryInterface` but also other interfaces that can return null now.

PR #493, opened 2016-03-24
Negate the statement

frederikbosch:
if `$hint` is not `null`, hint should be changed with the `sprintf` call. If `hint` is null it should stay `null`. Now it is the other way around.

### DIFF
--- a/src/Exception/OAuthServerException.php
+++ b/src/Exception/OAuthServerException.php
@@ -70,7 +70,7 @@ class OAuthServerException extends \Exception
     {
         $errorMessage = 'The request is missing a required parameter, includes an invalid parameter value, ' .
             'includes a parameter more than once, or is otherwise malformed.';
-        $hint = ($hint === null) ? sprintf('Check the `%s` parameter', $parameter) : $hint;
+        $hint = ($hint !== null) ? sprintf('Check the `%s` parameter', $parameter) : $hint;
 
         return new static($errorMessage, 3, 'invalid_request', 400, $hint);
     }


alexbilbie:
It is correct; if you don't provide a hint it will set one for you

frederikbosch:
It is not. This is my response body.

```json
{
  "error": "invalid_request",
  "message": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed.",
  "hint": "`%s` parameter is missing"
}
```

frederikbosch:
With the change it becomes.

```json
{
  "error": "invalid_request",
  "message": "The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed.",
  "hint": "Check the `password` parameter"
}
```

alexbilbie:
I fixed an instance I found a short while ago https://github.com/thephpleague/oauth2-server/commit/b1cf6a8436ac29e490d560987827cf63e7304556

frederikbosch:
@alexbilbie Please have another look at it.

1. No username [here](https://github.com/thephpleague/oauth2-server/blob/V5-WIP/src/Grant/PasswordGrant.php#L80)
2. Passes a second argument (`$hint`) as string [here](https://github.com/thephpleague/oauth2-server/blob/V5-WIP/src/Exception/OAuthServerException.php#L69).
3. So `$hint` is not null
4. However, `$hint === null` skips the `sprintf` call [here](https://github.com/thephpleague/oauth2-server/blob/V5-WIP/src/Exception/OAuthServerException.php#L73).
5. That leads to response body from my second message in this thread, while it should be the third.

frederikbosch:
Maybe my proposed solution is wrong. But the current code leads to a wrong response body.

alexbilbie:
The current code is incorrect :smile: 

Will fix. Thanks for finding the issue

frederikbosch:
Alright, thanks :+1:

PR #494, opened 2016-03-24
prevent double persist of token when doing refresh grant

frederikbosch:
Tokens were saved twice when doing a refresh grant. Has been fixed and tested.


alexbilbie:
Where are they persisted the first time?

src/Grant/RefreshTokenGrant.php
```diff
@@ -74,8 +74,6 @@ public function respondToRequest(
         // Issue and persist new tokens
         $accessToken = $this->issueAccessToken($accessTokenTTL, $client, $oldRefreshToken['user_id'], $scopes);
```
frederikbosch: First save access token

src/Grant/RefreshTokenGrant.php
```diff
@@ -74,8 +74,6 @@ public function respondToRequest(
...
         $refreshToken = $this->issueRefreshToken($accessToken);
```
frederikbosch: First save refresh token

alexbilbie:
Great spot. Thanks!

PR #497, opened 2016-03-25
allow data to be appended to jwt

frederikbosch:
At the moment the content of the JWT is defined from inside the library. It is not possible to append data. But an access token entity already has an interface. So it was not hard to implement the feature.

There must be a location that creates the access token, which can be implemented by the developer. So I added another method to the `AccessTokenRepositoryInterface` called `createNewToken`.

```php
public function createNewToken(ClientEntityInterface $clientEntity, array $scopes, $userIdentifier = null)
{
    return new MyAccessTokenEntity();
}
```

Now the developer can create his own his own implementation of `AccessTokenEntityInterface` which in turn implements for `convertToJWT`.

Secondly, after the access token is validated it must be possible to read contents from the jwt. I created an additional event for this `BearerWasValidated`. From the outside, this might look like this.

```php
$validator = new BearerTokenValidator();
$validator->addListener('validated', function (BearerWasValidated $event) { $jwt = $event->getToken(); });
$server = new Server(.. $validator);
```

When the access token is validated, the content from the jwt can be read.

### DIFF
--- a/src/AuthorizationValidators/BearerTokenValidator.php
+++ b/src/AuthorizationValidators/BearerTokenValidator.php
@@ -4,6 +4,8 @@ namespace League\OAuth2\Server\AuthorizationValidators;
 
 use Lcobucci\JWT\Parser;
 use Lcobucci\JWT\Signer\Rsa\Sha256;
+use League\Event\EmitterTrait;
+use League\OAuth2\Server\BearerWasValidated;
 use League\OAuth2\Server\CryptTrait;
 use League\OAuth2\Server\Exception\OAuthServerException;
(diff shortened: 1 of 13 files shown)


frederikbosch:
After thinking things through, it would be best to move `convertToJWT` to a separate class. Then we can skip `createNewToken` too. I would be happy to implement either method.

```php
class AccessTokenToJwtConverter implements AccessTokenToJwtConverterInterface {
  private $builder;

  public function __construct(Builder $builder) {
    $this->builder = $builder;
  }

  public function convert (AccessTokenEntity $accessToken) {
    // conversion
   return $jwt;
  }
}
```

PR #498, opened 2016-03-25
Client identifier passed where user identifier is expected

frederikbosch:
With client credentials grant, the wrong identifier is passed into the `issueAccessToken` method.

### DIFF
--- a/src/Grant/ClientCredentialsGrant.php
+++ b/src/Grant/ClientCredentialsGrant.php
@@ -34,7 +34,7 @@ class ClientCredentialsGrant extends AbstractGrant
         $scopes = $this->scopeRepository->finalizeScopes($scopes, $this->getIdentifier(), $client);
 
         // Issue and persist access token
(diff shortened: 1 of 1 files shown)


juliangut:
In the case of Client Credentials grant the 'client' is the resource owner, there is no user involved as the Application itself (client) is the one being granted access

frederikbosch:
@juliangut Sure. But that does not withstand the fact that a client identifier is passed to a method that expects a user identifier. Not every method is aware of the grand type that is requested, specifically the issueAccessToken method in AbstractGrant. This method is called by all grant type implementations. In turn that method calls the AccessTokenRepository which will store a user identifier while a client identifier is passed to it. Suppose you save the access token to a database with foreign keys, things will break!

And if you would have had a look at ghe PR, you would have seen that I pass in null to user identifier parameter, because - as your saying correctly - there is no user with the client credentials grant type.

Issue #499, opened 2016-03-27
Change AuthCodeGrant to make Templating optional


Bobselp:
Hey,

I tried to switch to the new V5-RC1. The client credentials grant works fine, however I have an issue with the authentication code grant.

In my opinion, `AuthCodeGrant->respondToAuthorizationRequest()` tries to do too much in a single function, and even stuff that in my opinion doesn't belong in the oauth-server code: Logging the user in via a web-based form. Every page that has users already has a login form, maybe extra secured or whatnot. The current flow forces you to create yet another login form, or to make your current login form & template work with the way required by the oauth-server code.

I would propose that if no templateRenderer is provided to `AuthCodeGrant::_construct`, `respondToAuthorizationRequest()` returns at around line 90 of `AuthCodeGrant.php`. It is then up to the developer to make the login & app validation functions and assemble a `UserEntity`. Lines 193-226 (after the validation part in the server-code) get moved to an own function that can be called with the UserEntity to generate the URL to redirect to after the (custom) login stuff validated the user.
This kind of resembles the flow of V4, where you could call `checkAuthorizeParams()`, do the validation, and then call `newAuthorizeRequest()` with the user ID to generate the redirect URI with the code.

This is only a rough proposal, there certainly is a better way to solve this. I hope you understand the problem I have with the code that tries to re-do what every website already has! :)

Other than that, congratulations to V5, besides this it really looks awesome so far!

alexbilbie:
Hi,

Thanks for getting in touch. I'm away until the 8th April, but will get
back to you as soon as I'm home.

Please could you flesh out this issue as much as you can so I can think
about it whilst I'm travelling and can better answer you when I'm back?

Thanks,

Alex

warlock39:
I have the same issue. I want to render oauth dialog by single page application, instead of render template on server side.

frederikbosch:
This can be solved together with PR #513, if we change the signature of [newHtmlResponse inside](https://github.com/frederikbosch/oauth2-server/blob/b6e967e39ca8717e93ab41a7a5749fa1c9ba547e/src/ResponseTypes/ResponseFactoryInterface.php). I am willing to help, if my help is being appreciated.

frederikbosch:
Then we could also remove AbstractAuthorizeGrant and by doing also reduce inheritance. I feel a grant has to too many layers: it implements, extends and uses (traits). This makes things too complex.

[...]

PR #513, opened 2016-04-04
Add response type factory for specific responses. Allows to append data to jwt. Abstract renderer.

frederikbosch:
Because I was struggling how to properly inject things to a response type, I came to the conclusion that this is not possible in current setup. There is a layer missing. I created `ResponseFactoryInterface`. This solving all kinds of problems.

1. JwtToken can be extended easily
2. ResponseType parameter in the method `respondToRequest` of a grant was only used in case of a BearTokenResponse. Now it takes a factory so it can create the response type that is needed.
3. The ResponseType implementations now only implement the methods required by the interface.
4. Less getters and setters with response types: no more `setAccessToken` for a response type that does not require an access token
5. An access token is not supposed to know anything of a JWT. Only the converter now knows both (because it is a converter).

This is quite a large PR, but I hope you can take the time to see if it beneficial to the library. I think it is.


frederikbosch:
Oh, and I also merged PR #501 by @juliangut because it helps solving conflicts beforehand.

src/AccessTokenToJwtConverter.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Generate a JWT from the access token
+     *
+     * @param AccessTokenEntityInterface $accessTokenEntity
+     *
+     * @return string
+     */
+    public function convert(AccessTokenEntityInterface $accessTokenEntity)
```
warlock39: How does it allow to append custom data to JWT? As i can see, it doesn't. Or?

frederikbosch:
@warlock39 The PR is updated. The things that you addressed have been taken care of.

1. You can now append data to a JWT token by using a decorator
2. You can now inject a `NullRenderer` to the `ResponseFactory`
3. You can now replace the used JWT implementation with your own by creating your own implementation of `ResponseFactoryInterface`

Regarding making the refresh token optional: that was not part of this PR. The idea of this PR is to give more flexibility with regard to the renderer and resulting token.

frederikbosch:
@warlock39 I hope you can give the same kind of feedback again. That would be appreciated since it helps getting better results!

frederikbosch:
@warlock39 Did some more commits. Could you have a look?

frederikbosch:
Maybe @juliangut can review too, since he has done quite some work too for this library. Hope he does not mind embedding him in this message.. :)

alexbilbie:
@frederikbosch if you haven't already could you please update the relevant examples so I can get a better idea of how the final API looks? I'm hoping to be able to review all of the open PRs and issues over the weekend

src/Grant/AuthCodeGrant.php
```diff
@@ -137,59 +145,30 @@ protected function respondToAuthorizationRequest(
...
-                    $request->getQueryParams()
-                ),
-            ]);
-
-            $htmlResponse = new HtmlResponse($this->accessTokenRepository);
-            $htmlResponse->setStatusCode(403);
-            $htmlResponse->setHtml($html);
-
-            return $htmlResponse;
+            return $responseFactory->newHtmlLoginResponse(
```
warlock39: Does it allow to use JSON representation as response? Or XML? This method forces me to render html on backend side. It would be great to use strategy here or so. API could look like:
```
return $responseFactory->newLoginResponse(
    LoginData($loginError, $postbackUri, $request->getQueryParams())
);
```

frederikbosch: No, I am going to change more. I had a discussion with @Bobselp already in #499. The `ResponseFactoryInterface` should act as middleware for those people that do not want templating.

frederikbosch: @warlock39 So I am going to extract more of the `AuthCodeGrant` to the `ResponseFactory`. I am going to take your suggestion into account.

src/Grant/AuthCodeGrant.php
```diff
@@ -137,59 +145,30 @@ protected function respondToAuthorizationRequest(
...
-
-            return $htmlResponse;
+            $encryptedUserId = $this->messageEncryption->encrypt(
+                json_encode([
+                    'user_id' => $userId,
+                ])
+            );
+
+            return $responseFactory->newHtmlAuthorizeResponse(
+                new AuthorizeData($client, $scopes, $postbackUri, $request->getQueryParams(), $encryptedUserId)
```
warlock39: To reduce number of arguments (and copy-pastes) `$postbackUri` can be moved to `ResponseFactory` and instead of passing `$request->getQueryParams()` we can pass just `$request`

src/Jwt/BearerRedirectResponse.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->accessToken = $accessToken;
+        $this->accessTokenToJwtConverter = $accessTokenToJwtConverter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function generateHttpResponse(ResponseInterface $response)
+    {
+        $jwtAccessToken = $this->accessTokenToJwtConverter->convert($this->accessToken)->getToken();
```
warlock39: `getToken()` is the method of `Lcobucci\JWT\Builder`, so it should not be exposed to client code (because `Builder` is internal dependency of `AccessTokenConverter`)

src/Grant/AuthCodeGrant.php
frederikbosch: Good suggestion, will change.

src/Jwt/BearerRedirectResponse.php
frederikbosch: Requires another interface, but will change it.

src/ResponseTypes/Dto/AuthorizeData.php
```diff
@@ -0,0 +1,90 @@
+<?php
```
warlock39: I am not sure, but it seems DTO is redundant

frederikbosch: What do you mean? The namespace?

warlock39:
@frederikbosch commented PR. Not so global, but i hope it will help

frederikbosch:
@alexbilbie Have been working on this PR today. To summarize what I have been doing.

1. Only pass a `ServerRequestInterface` and a `DateInterval` to a grant type. It should a return `ResponseTypeInterface`, not also receive one as parameter.
2. I extracted some code from the grant to a `ResponseFactoryInterface`. A response factory is injected into the grant type and could be seen as an object that shares code among the grant types that is related to the response.
3. A default `ResponseFactory` is included and receives as `RendererInterface` and a `AccessTokenConverterInterface`. This encapsulates the default response behaviour as you have defined it in the library: jwt for tokens and html for auth and implicit codes. One can inject its own renderer or access token converter. JWT and the corresponding library `lcobucci/jwt` is abstracted this way.
4. Someone that does not use a renderer, e.g. because only password grant is used, can inject a `NullRenderer` to the default `ResponseFactory`.
5. Someone that does not want html but rather returns `json` to auth code or implicit grants, because they use a SPA, should create their own implementation of `GrantTypeInterface`. It took me a while to think about this. But it is logical. `AuthCodeGrant` is *implementation* of a grant type so it defines how to process a request. If you want to define other behaviour, you should change the implementation. Hence @Bobselp and @warlock39 should create their own `GrantTypeInterface` implementations for their json responses.
6. There is no more `CryptTrait`. There is now a `MessageEncryptionInterface`. This allows people to their own encryption mechanism and/or own encryption library, e.g. `libsodium`.
7. `BearerTokenResponse` and `BearerRedirectResponse` are now unaware of a JWT token. They receive a access token converter.

I also changed the examples as you have requested. However, I did not test the examples. Hopefully you like what I have been doing. Of course, I am open for changes and discussion.

alexbilbie:
@frederikbosch is there any way this can be split up into smaller PRs please? This PR is changing way too much for me to try and get my head around all of it

PR #537, opened 2016-04-10
Update refresh token expire_time

ivyhjk:
**Refresh token** expire_time is currently taking the **access token** expire_time value


alexbilbie:
Great spot!

Thanks

PR #546, opened 2016-04-13
Updated link to Composer autoloading documentation

Loilo:
The autoloading section seems to have moved so I updated the link accordingly.


alexbilbie:
Thanks

PR #553, opened 2016-04-20
Remove redundant parameters in getNewToken()

markinjapan:
It seems unnecessary to include the `$client`, `$scopes`, and `$userIdentifier` parameters in `AccessTokenRepositoryInterface::getNewToken()`, as these are set in `AbstractGrant::issueAccessToken()` anyway. My apologies if I've misunderstood something though.


alexbilbie:
Thanks, I agree with this change. I'm going to merge it in for version `5.1.0`

ismailbaskin:
Is this project follow semver? If so this change totaly break something. Interfaces are kind of promise for developers.

alexbilbie:
I've reverted this PR for now. A more backwards compatible method of needs to be found.

PR #558, opened 2016-04-25
Setters in ClientTrait

markinjapan:
I noticed the `ClientTrait` was missing setters, but I think these would be useful to provide.


alexbilbie:
These setter methods originally existed but were removed in #490 after some discussion in another issue that really these methods weren't used by the library and were really the responsibility of implementations

alexbilbie:
Thanks for your contribution though

PR #560, opened 2016-04-27
Fix hints in PasswordGrant

luke83:
When omitting `username` or `password` you get wrong hint `` `%s` is missing ``, with this PR you will get `` `username` is missing`` or `` `password` is missing``


alexbilbie:
Thanks @luke83. These parameters shouldn't have been sent as they're handled by the parent method

luke83:
I thought you wanted a specific message for those parameters, parent method handling is more consistent, much better!

PR #561, opened 2016-04-28
Add meaning of `Resource owner` to terminology.md

bobmulder:
I saw the section `Roles` of the [OAuth2 specification](https://tools.ietf.org/html/rfc6749#section-1.1) and could not find the meaning of `Resource owner` while all other roles where defined. To complete the list, I've added the missing role.


alexbilbie:
Thanks

PR #563, opened 2016-04-30
Use constant for event name instead of explicit string

ismailbaskin:


alexbilbie:
Thanks for your PR.

I'm wondering if these constants could be added to the `RequestEvent` class instead?

ismailbaskin:
It is only convention. Do you want to do I do?

alexbilbie:
I think having the constants in the `RequestEvent` class is better; no need to introduce a brand new class into the library

PR #573, opened 2016-05-04
Include redirect_uri check on authorization endpoint

ismailbaskin:
If multiple redirection URIs have been registered, if only part of
the redirection URI has been registered, or if no redirection URI has
been registered, the client MUST include a redirection URI with the
authorization request using the "redirect_uri" request parameter.

https://tools.ietf.org/html/rfc6749#section-3.1.2.3


alexbilbie:
This also needs applying to the Implicit Grant too

ismailbaskin:
applied to Implicit Grant too

alexbilbie:
Thanks; I will review this as soon as I get a chance

Sephster:
Looks good to me. Minor changes to be made to keep consistency with the branch e..g use `@expectedException` in test but this can be resolved outside of this PR.

PR #578, opened 2016-05-11
unify middleware exception responses

juliangut:
If it isn't an OAuthServerException, result still needs to be properly formatted (json) and not just a string error message


alexbilbie:
Looks good!

PR #581, opened 2016-05-19
Update auth-server-auth-code.md

bobmulder:
Guess this is not the Authorization Server ;) Didn't know how to call it so I named it to 'client callback'.

### DIFF
--- a/auth-server-auth-code.md
+++ b/auth-server-auth-code.md
@@ -24,7 +24,7 @@ All of these parameters will be validated by the authorization server.
 
 The user will then be asked to login to the authorization server and approve the client.
 
-If the user approves the client they will be redirected back to the authorization server with the following parameters in the query string:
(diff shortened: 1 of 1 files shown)


bobmulder:
Wow, made a mistake there ;P haha sorry!

alexbilbie:
You've found a mistake in the docs; could you change it to:

"If the user approves the client they will be redirected from the authorisation server to the client's redirect URI with the following parameters in the query string:"

bobmulder:
Are you sure it was a mistake @alexbilbie? I am trying to implement the Authorization Code Grant right now, but I really struggle at this point...

alexbilbie:
Yes, once the user has authorised the client they are redirect back to the client with the authorisation code which is exchanged for an access token. Your PR is correct

bobmulder:
Pfew :)

PR #601, opened 2016-06-17
Added a check for unique access token constraint violation

### LINKED ISSUE #596: Discussion: generation issue access tokens and randomness vs uniqueness
At the moment the new access token is issued with the following code

```php
    protected function issueAccessToken(
        \DateInterval $accessTokenTTL,
        ClientEntityInterface $client,
        $userIdentifier,
        array $scopes = []
    ) {
        $accessToken = $this->accessTokenRepository->getNewToken($client, $scopes, $userIdentifier);
        $accessToken->setClient($client);
        $accessToken->setIdentifier($this->generateUniqueIdentifier());
        $accessToken->setExpiryDateTime((new \DateTime())->add($accessTokenTTL));

        foreach ($scopes as $scope) {
        }

        $this->accessTokenRepository->persistNewAccessToken($accessToken);

        return $accessToken;
    }
```
and the new access token identifier is generated with

```php
    protected function generateUniqueIdentifier($length = 40)
    {
        try {
            return bin2hex(random_bytes($length));
            // @codeCoverageIgnoreStart
        } catch (\TypeError $e) {
            throw OAuthServerException::serverError('An unexpected error has occurred');
        } catch (\Error $e) {
            throw OAuthServerException::serverError('An unexpected error has occurred');
        } catch (\Exception $e) {
            // If you get this message, the CSPRNG failed hard.
            throw OAuthServerException::serverError('Could not generate a random string');
        }
        // @codeCoverageIgnoreEnd
    }
```

And that's where a problem appears: even though it is 40 bytes, one still cannot interchange the "unique" term with the "random" term.

Suggestion: the method that issues an access token might have done that in a loop with N maximum iterations (to prevent an infinite loop in case of an implementation or other issues) and catch for the specifically designed exception thrown from `$this->accessTokenRepository->persistNewAccessToken($accessToken);`.

An exception can be of a predefined type, eg: `League\OAuth2\Server\Exception\UniqueAccessTokenIdentifierConstraintViolationException`.

zerkms:
If it happens - the server will retry to generate token at most 10 times until it gives up

Fixes #596

### DIFF
--- a/src/Exception/UniqueTokenIdentifierConstraintViolationException.php
+++ b/src/Exception/UniqueTokenIdentifierConstraintViolationException.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * @author      Ivan Kurnosov <zerkms@zerkms.com>
+ * @copyright   Copyright (c) Alex Bilbie
+ * @license     http://mit-license.org/
+ *
+ * @link        https://github.com/thephpleague/oauth2-server
+ */
+
+namespace League\OAuth2\Server\Exception;
+
+
+class UniqueTokenIdentifierConstraintViolationException extends OAuthServerException
+{
+    public static function create()
+    {
+        $errorMessage = 'Could not create unique access token identifier';
+
+        return new static($errorMessage, 100, 'access_token_duplicate', 500);
+    }
+}
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -16,6 +16,7 @@ use League\OAuth2\Server\Entities\AccessTokenEntityInterface;
 use League\OAuth2\Server\Entities\ClientEntityInterface;
 use League\OAuth2\Server\Entities\ScopeEntityInterface;
 use League\OAuth2\Server\Exception\OAuthServerException;
+use League\OAuth2\Server\Exception\UniqueTokenIdentifierConstraintViolationException;
 use League\OAuth2\Server\Repositories\AccessTokenRepositoryInterface;
 use League\OAuth2\Server\Repositories\AuthCodeRepositoryInterface;
 use League\OAuth2\Server\Repositories\ClientRepositoryInterface;
@@ -34,6 +35,8 @@ abstract class AbstractGrant implements GrantTypeInterface
     use EmitterAwareTrait, CryptTrait;
 
     const SCOPE_DELIMITER_STRING = ' ';
+
+    const MAX_RANDOM_TOKEN_GENERATION_ATTEMPTS = 10;
 
     /**
      * @var ClientRepositoryInterface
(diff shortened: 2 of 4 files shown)


alexbilbie:
I think the loop just needs to be around `$accessToken->setIdentifier($this->generateUniqueIdentifier());` and the try/catch block; no need to continuously create access token entity objects

zerkms:
@alexbilbie well, that's what I thought initially, but there is no contract that guarantees consistency of the token between multiple `setIdentifier`'s

alexbilbie:
Why would this not work?

```php
protected function issueAccessToken(
    \DateInterval $accessTokenTTL,
    ClientEntityInterface $client,
    $userIdentifier,
    array $scopes = []
) {
    $maxGenerationAttempts = 10;

    $accessToken = $this->accessTokenRepository->getNewToken($client, $scopes, $userIdentifier);
    $accessToken->setClient($client);
    $accessToken->setExpiryDateTime((new \DateTime())->add($accessTokenTTL));

    foreach ($scopes as $scope) {
    }

    while ($maxGenerationAttempts-- > 0) {
        $accessToken->setIdentifier($this->generateUniqueIdentifier());
        try {
            $this->accessTokenRepository->persistNewAccessToken($accessToken);
            return $accessToken;
        } catch (UniqueAccessTokenIdentifierConstraintViolationException $e) {
            if ($maxGenerationAttempts === 0) {
                throw $e;
            }
        }
    }
}
```

zerkms:
This might work and my initial implementation was identical to that, but then I thought that explicit is better.

By that I mean: what if some crazy implementation resets all the fields set in case when you invoke `setIdentifier` for the second time. Like: it is another identifier, so the identity of this entity has changed so the other fields must be reset. (and yes, I understand it is even less likely than 40 byte random value to duplicate) :-)

alexbilbie:
I understand the concern but I think that's being overly defensive 😄

zerkms:
Sure, I'll change it to the proposed one, since mine is a bit paranoid and less readable.

zerkms:
Done.

Btw, as you noticed I've put my tests into the `ImplicitGrant` test suite, since it has the smallest bootstrap overhead compared to others.

Ideally every grant should be covered, since it's a detail of their behaviour, but those would be literally copy-paste.

alexbilbie:
Thanks.

Am just thinking that this should probably extend to the `issueAuthCode` and and `issueRefreshToken` methods too?

zerkms:
I will definitely have a look slightly later, for me it's friday evening just started 🐱

zerkms:
I've added the `issue*` methods modifications and will commit tests later today (and will squash all the commits together)

alexbilbie:
Thanks so much @zerkms!

zerkms:
So the all code squashed and pushed. I have concerns about bloated tests, but I don't have any solution for that at the moment (apart of using dirty reflection tricks, which I personally would avoid).

zerkms:
It's funny - I could get a race condition on travis :-D The changeset id on this page and the changeset id checked against do not match.

alexbilbie:
I've seen race conditions on Travis too.

Regarding tests I agree they are bloated; the vast majority of the tests need converting to functional tests (i.e. spinning up a web server and sending HTTP calls to it to verify conditions). That would make adding new tests much easier (as it'd move to a one test one file model). Longer term fuzz tests can be introduced to try and trick the authorization server with garbage data.

zerkms:
@alexbilbie if there is nothing we need/can improve at the moment, may be we could merge it then? :-)

PR #604, opened 2016-06-22
Look at Authorization header directly for HTTP Basic auth checks

iansltx:
Should allow for better compatibility with server implementations that aren't sitting on top of a standard SAPI (e.g. persistent web servers building a PSR-7 compatible request from a socket-received message).

One catch here is that I've seen Apache hijack the HTTP Authorization header in the past, though that would probably impact the other aspects of the server just as much as it would this, so I think that risk is manageable.

Added tests to cover all paths through the new code, so the AbstractGrant type still has 100% coverage :)

Noticed that, as of the latest versions of PHPUnit, the mock creation method is deprecated. Maybe that needs to be updated? Haven't checked to see whether the replacements are PHPUnit 4.8 compatible though, so maybe they need to stay in order to test on older PHP versions?

Finally, I didn't see a develop branch, so PRing this against master. I'll re-PR if that's not the right place at this point.

### DIFF
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -137,21 +137,15 @@ abstract class AbstractGrant implements GrantTypeInterface
      */
     protected function validateClient(ServerRequestInterface $request)
     {
-        $clientId = $this->getRequestParameter(
-            'client_id',
-            $request,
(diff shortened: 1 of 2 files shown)


alexbilbie:
Cheers!

PR-ing against master is correct.

I will open a new ticket regarding PHPUnit mock creation

PR #605, opened 2016-06-22
Added catch Runtime exception when parsing JWT string

jfilla:


alexbilbie:
Is this not caught by the `InvalidArgumentException` catch block? https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationValidators/BearerTokenValidator.php#L77-L80

jfilla:
No, it is a different exception.

https://github.com/lcobucci/jwt/blob/3.1/src/Parsing/Decoder.php#L36

Great work btw, thanks.

Issue #550, opened 2016-04-17
How to implement ClientRepositoryInterface::getClientEntity()


lookyman:
Talking about Authorization Code grant, but it may apply to others as well.

If the client is confidential, you have to pass the `client_secret` parameter to the access_token endpoint. If you don't do that, this https://github.com/thephpleague/oauth2-server/blob/5b192b3548c769cb74f7565418f7cbc2a7d55203/src/Grant/AbstractGrant.php#L161 gets `null` as a `$clientSecret` argument. So you have to make a check in that method, something like "if the client has secret set, compare it with the passed `$clientSecret` and scream if it doesn't match".

All of that is still OK. However, the same method gets called from here https://github.com/thephpleague/oauth2-server/blob/5b192b3548c769cb74f7565418f7cbc2a7d55203/src/Grant/AuthCodeGrant.php#L166, and because here you don't send the `client_secret` parameter, it gets passed `null` every time. But now the check in that method starts to scream.

Am I just missing something important, or is this a bug?

alexbilbie:
You're right, you've found a bug.

The easiest way I can think of resolving this is introducing a 4th parameter on the `getClientEntity` method on the `ClientRepositoryInterface`. This parameter would be a boolean `$mustValidateSecret`.

When `$mustValidateSecret === true` the client repository must attempt to validate the secret if the client is confidential.

For the authorisation and implicit grants, when validating the client as part of the authorisation request `$mustValidateSecret` will be `false`. For the other grants it will be `true`.

(cc/ @assembledadam; I believe you've also found this bug)

alexbilbie:
I've released `5.0.1` which addresses the issue here.

lookyman:
Thank you for a quick response!

PR #607, opened 2016-06-30
Make ClientRepositoryInterface more flexible

lucadegasperi:
This small change will allow the use of the `ClientRepositoryInterface` for more use cases than simply validating clients when authorizing them. There might be some places where this change will affect the behavior. I also think the `$mustValidateSecret` is redundant since in an implementation a check could be done wether `$clientSecret` is null or not.

### DIFF
--- a/src/Repositories/ClientRepositoryInterface.php
+++ b/src/Repositories/ClientRepositoryInterface.php
@@ -17,12 +17,12 @@ interface ClientRepositoryInterface extends RepositoryInterface
      * Get a client.
      *
      * @param string      $clientIdentifier   The client's identifier
-     * @param string      $grantType          The grant type used
+     * @param null|string $grantType          The grant type used (if sent)
      * @param null|string $clientSecret       The client's secret (if sent)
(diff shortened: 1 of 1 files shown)


lookyman:
https://github.com/thephpleague/oauth2-server/issues/550

alexbilbie:
I understand your requirement - the problem is that changing the interface will break everyones' code so in the constraints of semantic versioning I'd have to release a new major version.

Having said this I do think a new major version is on the horizon that addresses some small short comings of v5. As there won't be that much change I've been debating supporting both v5 and v6 side by side for 12-18 months before deprecating v5. I'll write a formal Github issue about this though.

Sephster:
This looks good to me although I will do some more concrete checks later this week. I am planning on releasing v7 some time this week so will aim to get this included. 

The 12-18 month period of support for v5 has now passed as well so I will look to retire this version at the same time. Thanks for your patience!

Sephster:
Merging this in although I'm going to change the `mustValidateSecret` back to default to true instead of false as I think it is better to be cautious about this check. Most places in the default library are specifying this anyway but I think it is good to default to the most secure option

PR #613, opened 2016-07-08
ResourceServer: set public key only for BearerTokenValidator

lookyman:
`AuthorizationValidatorInterface` does not contain `setPublicKey()` method.


lookyman:
ping @alexbilbie

Sephster:
I think you'd fixed this in your recent PHPStan PR so will close this one. Thanks for the fix regardless and apologies we took a while to pick this up

PR #614, opened 2016-07-08
Improved tests

lookyman:


alexbilbie:
Are the mock builder changes with both PHPUnit 4.8 + 5.0 ?

lookyman:
Tested it for both and it works. There were mixed mock/mockbuilder tests before anyway. I just unified it to get rid of the deprecated warnings in PHPUnit 5.0.

lookyman:
ping @alexbilbie

PR #615, opened 2016-07-08
AuthorizationServer: set private key only for BearerTokenResponse

lookyman:
`ResponseTypeInterface` does not contain `setPrivateKey()` method.


lookyman:
Test incoming..

lookyman:
ping @alexbilbie

PR #616, opened 2016-07-09
Updated PHPDoc

lookyman:


alexbilbie:
Was this done with some automated tool? Just there are some changes in here that don't match my preferred code style - e.g. changing `$foo === false` to `!$foo`

lookyman:
I made those changes purely to be consistent, there were more instances of `!$foo` than `$foo === false` in the code, so I changed the latter into the former. Should I reverse it? Also, let me know if you want anything else done differently.

alexbilbie:
I've had a look through and that's the only thing I'm not comfortable with. Thanks 😄

lookyman:
Done.

Issue #623, opened 2016-07-16
Difficult to specify additional data on Bearer Token response


iansltx:
RFC 6749 allows for adding arbitrary parameters to the response body when a token is returned. That's doable in the current version of the library:

1. Create a `BearerTokenResponse` replacement class (probably extending `AbstractResponseType`) that includes your extra params in `generateHttpResponse()`.
2. Constructor-inject your new class into `AuthorizationServer`.

The above would be fine, but you have to rewrite (aka copy-paste) the entire `generateHttpResponse()` call from `BearerTokenResponse` and add a few lines in order to add the params you want, when all you really want to do is an array_merge between what the system spits out and your extra fields.

Also, I think we might have a regression on #417, because while the access token's scopes are returned in the JWT, there's no easy way to return them at the top level of the response object JSON. That's one of my use cases for the above, though I have another field or two that I need to add as well.

I'll have a PR shortly for the parameter injection piece. Don't have time to add tests for the scope piece (can't use getScopes() in this case because it'll hand back an array rather than an space-sep string), but maybe later...

alexbilbie:
Am very happy to review a PR for this, I think it's a good idea 👍

iansltx:
Cool. Working on tests now. Just realized that the responseType is a constructor param for AuthorizationServer, and updated the issue to reflect that adding params is (and will continue to be) a two-step process rather than three.

PR #624, opened 2016-07-16
Allow easy addition of custom fields to Bearer token response

iansltx:
Solves #623.

See tests/ResponseTypes/BearerTokenResponseWithParams.php for an example of how to add fields. Note that fields already defined in the response will not be overwritten.


pounard:
Yay ! That's exactly what I do need, I had to extend the BearerTokenResponse and rewrite almost everything to achieve what I needed.

bobmulder:
Any documentation about this? :)

PR #625, opened 2016-07-19
Key file auto-generation from string

juliangut:
Allow storing keys in database instead of plain files.

RSA key format is detected and a file is automatically generated on sys_get_temp_dir() and used instead


juliangut:
Tests coverage

alexbilbie:
This is great! Thanks @juliangut

PR #633, opened 2016-08-04
better solution for reading key content (string)

### LINKED ISSUE #631: CryptKey using a different protocol than file://
My key is using different protocol then `file://`. This is not possible because [`file://` is prepended](https://github.com/thephpleague/oauth2-server/blob/master/src/CryptKey.php#L40) at this moment. I have no idea why. Maybe because of some openssl function requiring it? So this fails.

```php
$cryptKey = new CryptKey('protocol://host/file/key.key');
```

Moreover, I am sorry to say, [the constructor of CryptKey is horrible](https://github.com/thephpleague/oauth2-server/blob/master/src/CryptKey.php#L35). All the logic of that class takes place in the constructor. That is not what a constructor is for. A decent CryptKey file with same functionality should look as follows. Too bad it is backward incompatible. I suggest creating a `CryptKeyInterface` so people can come up with their own CryptKey implementations.

```php
final class CryptKey
{
    protected $keyPath;
    protected $passPhrase;

    public function __construct($keyPath, $passPhrase = null)
    {
        if (!file_exists($keyPath) || !is_readable($keyPath)) {
            throw new \LogicException(sprintf('Key path "%s" does not exist or is not readable', $keyPath));
        }

        $this->keyPath = $keyPath;
        $this->passPhrase = $passPhrase;
    }

    public function getKeyPath()
    {
        return $this->keyPath;
    }

    public function getPassPhrase()
    {
        return $this->passPhrase;
    }

    public static function fromString($keyContent) {
        $keyPath = sys_get_temp_dir() . '/' . sha1($keyContent) . '.key';

        if (!file_exists($keyPath) && !touch($keyPath)) {
            // @codeCoverageIgnoreStart
            throw new \RuntimeException('"%s" key file could not be created', $keyPath);
            // @codeCoverageIgnoreEnd
        }

        file_put_contents($keyPath, $keyContent);

        return new self ('file://' . $keyPath);
   }
}

frederikbosch:
Fixes #631. Reverts solution from #625.

### DIFF
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -13,9 +13,6 @@ namespace League\OAuth2\Server;
 
 class CryptKey
 {
-    const RSA_KEY_PATTERN =
-        '/^(-----BEGIN (RSA )?(PUBLIC|PRIVATE) KEY-----\n)(.|\n)+(-----END (RSA )?(PUBLIC|PRIVATE) KEY-----)$/';
-
     /**
      * @var string
      */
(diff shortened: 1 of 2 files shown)


frederikbosch:
@alexbilbie This is weird. I do not see the commit in the commit tree of the master, but this PR says it is merged. What happened?

PR #636, opened 2016-08-06
Added support for Symfony HttpFoundation

alexbilbie:
* `AuthorizationServerMiddleware` renamed to `Psr7AuthorizationServerMiddleware`
* `ResourceServerMiddleware` renamed to `Psr7ResourceServerMiddleware`
* Added deprecation warning on `AuthorizationServerMiddleware`
* Added deprecation warning on `ResourceServerMiddleware `
* Added new `HttpMessageConverter` class
* Added `SymfonyAuthorizationServerMiddleware`
* Added `SymfonyResourceServerMiddleware`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,9 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8 || ^5.0",
-        "zendframework/zend-diactoros": "^1.0",
-        "indigophp/hash-compat": "^1.1"
+        "zendframework/zend-diactoros": "^1.3",
+        "indigophp/hash-compat": "^1.1",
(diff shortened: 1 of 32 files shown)


alexbilbie:
cc/ @taylorotwell

pounard:
Actually this might not be necessary, I am using it with Symfony, but the sensio-extra-bundle already provide automatic request/response conversion to/from PSR7, and it's working nicely without any additions. You just have to type your controllers with Psr7 interfaces instead of Symfony's Request class.

lookyman:
I also don't think this is necessary. The library should be framework agnostic, and framework integrations can easily be in separate packages.

alexbilbie:
Agreed; I'm going to release it as a separate package

PR #645, opened 2016-08-30
Update Laravel link

vinkla:
The passport package has replaced oauth2-server-laravel for future versions of Laravel.


alexbilbie:
Please can you retain the existing link and add the Passport link as
"Laravel Passport (Official)"

vinkla:
Done.

PR #648, opened 2016-08-31
Alternative approach to RFC 7636

lookyman:
Basically just trying to see what you think. It does break BC, but could be more useful in the long run. Plus it fixes some issues, mainly that according to the RFC, new projects should not support plain method unless absolutely necessary, while the original implementation has that baked in.

### DIFF
--- a/src/CodeChallengeVerifiers/S256Verifier.php
+++ b/src/CodeChallengeVerifiers/S256Verifier.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * @author      Lukáš Unger <lookymsc@gmail.com>
+ * @copyright   Copyright (c) Lukáš Unger
+ * @license     http://mit-license.org/
+ *
+ * @link        https://github.com/thephpleague/oauth2-server
+ */
+
+namespace League\OAuth2\Server\CodeChallengeVerifiers;
+
+class S256Verifier implements CodeChallengeVerifierInterface
+{
+    /**
+     * Return code challenge method.
+     *
+     * @return string
+     */
+    public function getMethod()
+    {
+        return 'S256';
+    }
+
+    /**
+     * Verify the code challenge.
+     *
+     * @param string $codeVerifier
+     * @param string $codeChallenge
+     *
+     * @return bool
+     */
+    public function verifyCodeChallenge($codeVerifier, $codeChallenge)
+    {
+        return hash_equals(
+            strtr(rtrim(base64_encode(hash('sha256', $codeVerifier, true)), '='), '+/', '-_'),
+            $codeChallenge
+        );
+    }
+}
--- a/tests/CodeChallengeVerifiers/PlainVerifierTest.php
+++ b/tests/CodeChallengeVerifiers/PlainVerifierTest.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace LeagueTests\CodeChallengeVerifiers;
+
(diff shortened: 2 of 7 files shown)


tests/CodeChallengeVerifiers/PlainVerifierTest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace LeagueTests\CodeChallengeVerifiers;
+
+use League\OAuth2\Server\CodeChallengeVerifiers\PlainVerifier;
+
+class PlainVerifierTest extends \PHPUnit_Framework_TestCase
```
Sephster: Please could you change this to extend `TestCase` and use `PHPUnit\Framework\TestCase`. We should replace all instances of `PHPUnit_Framework_TestCase` in this manner for this PR

tests/CodeChallengeVerifiers/PlainVerifierTest.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    public function testGetMethod()
+    {
+        $verifier = new PlainVerifier();
+        self::assertEquals('plain', $verifier->getMethod());
```
Sephster: To maintain consistency with our other tests, we should call `$this->assertEquals` instead of `self::assertEquals`. If you could change all instances to keep in line with the current tests that would be appreciated. Many thanks

src/CodeChallengeVerifiers/S256Verifier.php
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @param string $codeVerifier
+     * @param string $codeChallenge
+     *
+     * @return bool
+     */
+    public function verifyCodeChallenge($codeVerifier, $codeChallenge)
+    {
+        return hash_equals(
+            urlencode(base64_encode(hash('sha256', $codeVerifier))),
```
Sephster: Please update this to match the new implementation currently in master. It removes base64 padding, the need to url encode, and ensures that hashing occurs after we've base 64 encoded as per the RFC.

lookyman:
Sure, I'll look into it asap..

lookyman:
@Sephster Rebased on master and modified according to your suggestions.

Also, if this ever gets merged, the documentation has to be updated. (just putting that here so we don't forget..)

Sephster:
Thanks for the quick turn around @lookyman. I will look to get something set up for this soon. We already have a version 7 branch but it hasn't been touched in a while. I believe it has the initial workings for an Open ID implementation. I will need to consult the other maintainers but it might be the case that we remove this branch or work on it to complete the implementation depending on its state and then merge this into a version 7 branch.

Thanks again for making the changes!

lookyman:
@Sephster Fixed according to https://github.com/thephpleague/oauth2-server/pull/842 and rebased on master.

Sephster:
@lookyman I've changed the way we are handling PKCE in this PR: https://github.com/thephpleague/oauth2-server/pull/938

I really like the interfaces you have created for the plain and s256 verifiers though and was wondering if you'd be ok with me integrating those changes in to the aforementioned pull request?

Also if you have any thoughts on the approach I've proposed that would be appreciated. Thanks!

Sephster:
Thanks @lookyman. I will aim to get the changes integrated into the PR tomorrow. Thank you

Sephster:
Thanks for your efforts here @lookyman. I've added some of the changes here to my pull request as discussed. Thanks very much for your efforts. I will close this PR now as I think it is no longer required.

PR #652, opened 2016-09-08
less restrictive on Authorization header check

rickshawhobo:


alexbilbie:
I've just pushed `4.1.6`. Sorry about the delay

PR #688, opened 2016-11-30
incompatability with PHP 5.4

georgewilde:
Fixed PHP 5.4 erroring with "Can't use method return value in write context".

empty() needs to access the value by reference (in order to check whether that reference points to something that exists), and PHP before 5.5 didn't support references to temporary values returned from functions as outlined in this post http://stackoverflow.com/a/4328049

The authorization has been moved into a variable and this is now checked as to whether it's empty.

### DIFF
--- a/src/ResourceServer.php
+++ b/src/ResourceServer.php
@@ -138,7 +138,9 @@ class ResourceServer extends AbstractServer
      */
     public function determineAccessToken($headerOnly = false)
     {
-        if (!empty($this->getRequest()->headers->get('Authorization'))) {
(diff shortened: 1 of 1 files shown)


georgewilde:
I'm unsure why StyleCI has failed as it's interface doesn't make that clear. It highlights changes I've not made to the file. Please message me with any changes I can make to my PR to help it get accepted.

Thanks

wilsonge:
The library requires PHP 5.5? https://github.com/thephpleague/oauth2-server/blob/master/composer.json#L7

PR #689, opened 2016-12-13
password_verify incorrect use

chaegumi:


examples/src/Repositories/ClientRepository.php
```diff
@@ -21,6 +21,7 @@ public function getClientEntity($clientIdentifier, $grantType, $clientSecret = n
     {
         $clients = [
             'myawesomeapp' => [
+				'passwd'          => 'abc123',
```
lookyman: You have tabs here instead of spaces.

alexbilbie:
Your changes aren't correct I'm afraid.

The `$clientSecret` is the plain text that wants verifying against the hash

chaegumi:
http://www.php.net/manual/en/function.password-verify.php

·boolean password_verify ( string $password , string $hash )·

```php

<?php
// See the password_hash() example to see where this came from.
$hash = '$2y$07$BCryptRequires22Chrcte/VlQH0piJtjXl.0t1XkA8pw9dMXTpOq';

if (password_verify('rasmuslerdorf', $hash)) {
    echo 'Password is valid!';
} else {
    echo 'Invalid password.';
}
?>

```

alexbilbie:
Where is `$clientSecret` a hashed value?

chaegumi:
Oh, It's my fault, I passed the hashed value.

```php

'form_params' => [
               'grant_type' => 'authorization_code',
               'client_id' => 'myawesomeapp',
               'client_secret' => '$2y$10$0cnZgLgUlIBkges1dDxsYe4w9fSvrj7zpeg33Lj7QdDbG6k4tYZOW', // Here must be abc123 
               'redirect_uri' => 'http://foo/callback',
               'code' => $code
           ]

```

The correct use is:

```php

'form_params' => [
               'grant_type' => 'authorization_code',
               'client_id' => 'myawesomeapp',
               'client_secret' => 'abc123', 
               'redirect_uri' => 'http://foo/callback',
               'code' => $code
           ]

```

PR #691, opened 2016-12-24
Getting parameters from request

gasmx:
I'm using PHP 5.6.16 with Laravel 5.3 and I received some exceptions after trying to generate access tokens via GET.

The issue is in getRequestParameter(), where the property 'parsedBody' comes as an empty array.

This little change worked for me, I expect to get your opinion.


alexbilbie:
Why are you passing the parameters via query string and a GET request?

gasmx:
@alexbilbie Why note use a GET request?

alexbilbie:
Because the OAuth specification dictates which HTTP methods each grant should use and this library does it's best to follow the specification as closely as possible.

gasmx:
Got it, thanks for reply.

PR #697, opened 2017-01-24
Fix PKCE code verifier encoding to match specification

fkooman:
The current implementation of PKCE does not follow the specification
correctly regarding the encoding of the code verifier. This patch
correctly encodes the hash of the code verifier according to
Appenix A of RFC 7636.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -144,7 +144,7 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
                 case 'S256':
                     if (
                         hash_equals(
-                            urlencode(base64_encode(hash('sha256', $codeVerifier))),
+                            strtr(rtrim(base64_encode(hash('sha256', $codeVerifier)), '='), '+/', '-_'),
                             $authCodePayload->code_challenge
                         ) === false
                     ) {
--- a/tests/Grant/AuthCodeGrantTest.php
+++ b/tests/Grant/AuthCodeGrantTest.php
@@ -630,7 +630,7 @@ class AuthCodeGrantTest extends \PHPUnit_Framework_TestCase
                             'user_id'               => 123,
                             'scopes'                => ['foo'],
                             'redirect_uri'          => 'http://foo/bar',
-                            'code_challenge'        => urlencode(base64_encode(hash('sha256', 'foobar'))),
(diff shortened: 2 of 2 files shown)


fkooman:
I just noticed #648...

Sephster:
This looks good to me. Have checked and all appears to be working so going to merge in. I noticed that we are hashing before base64 encoding and in RFC 7636, in appendix A it shows that we should has _after_ base64 encoding but I will make that change outside of this branch. Thanks for your contribution

fkooman:
I'd also recommend using [paragonie/constant-time-encoding](https://github.com/paragonie/constant_time_encoding) for constant time Base64 encoding.

Sephster:
Thank you for the suggestion. I think that this would likely be implemented on the client side rather than the server side as, for PKCE, it is the client device that is vulnerable for man in the middle attacks. If you have any documentation statin that this is not the case, I will be happy to revisit this. Thanks again for your PR

sgomez:
This is wrong, `base64_encode` must be done over the binary output of hash('sha256'). The process should be this:

`strtr(rtrim(base64_encode(hash('sha256', $codeVerifier, true)), '='), '+/', '-_')` 

This can be checked with the example described in [RFC7636 (Appendix B)](https://tools.ietf.org/html/rfc7636#appendix-B):

```
<?php
$codeVerifier = 'dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk';
$codeChallenge = 'E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM';

echo hash_equals(
  strtr(rtrim(base64_encode(hash('sha256', $codeVerifier, true)), '='), '+/', '-_'),
  $codeChallenge
) ? 'True' : 'False';
```

This returns `True`.

With the implementation of this PR:

```
<?php
$codeVerifier = 'dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk';
$codeChallenge = 'E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM';

echo hash_equals(
  hash('sha256', strtr(rtrim(base64_encode($codeVerifier), '='), '+/', '-_')),
  $codeChallenge
) ? 'True' : 'False';
```

This return `False`. 

hash('sha256', ..., false) returns a hexadecimal string and there are no way than match with _code challenge_ that it is a base64 string.

PR #703, opened 2017-02-04
Add hasRedirect() method for OAuthServerException

### LINKED ISSUE #694: No way to tell whether an OAuthException will trigger a redirect
Use case here is that, if an OAuthException is thrown while creating an AuthorizationRequest, you may well be interacting with that authorization request in the context of a rendered page. In that case, if you aren't redirecting the user back to the caller, you'll want to render a response on your own, but if there is a redirect involved, you want to let that pass through.

The catch here is that there's no way to simply check whether an OAuthException will trigger a redirect, without knowing which error codes have redirects and which don't. This is because redirectUri is private (not that you can easily extend the class anyway), and it's basically the thing that determines whether you get redirected or not. You can't tell via the status code etc., because they don't take into account whether there's a redirect or not either.

"Well, you can generate an HTTP response and check the status code on that" would be a valid workaround here, except the one thing that's referenced rather than cloned when creating a response is the body, so if you want to build on top of a response Slim-style, you can't render the response, then back out when you realize that you aren't getting a redirect; by then, the stream associated with your response (which is shared even if you save into a new variable) already has the error JSON in it.

The workaround here is to create a brand new response, throw the exception into it, and either return that response if it's a redirect or throw it away if it's not. Which may be fine the way middleware stack interactions are going, but it's counterintuitive when you're dealing with the older onion-style middlewares. Or maybe I missed somewhere in the docs where this idiosyncrasy (which isn't this lib's fault) was mentioned :)

Anyway, documenting this gotcha and a workaround would be one "fix" for most cases, since onion-style middlewares likely aren't modifying the response until on the way out anyway. Alternately, seems like adding a hasRedirect() method or similar to the OAuthException class would provide enough information to handle this cleanly for cases where you don't want to spit JSON out to the end user on an auth code or implicit grant landing page.

I don't mind PRing either of the above approaches (docs or additional method, which would constitute a minor version bump I think), but figured I'd see which one would be preferable here. Thanks!

iansltx:
Resolves #694.


tests/ExceptionTest.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace LeagueTests\Utils;
```
Sephster: I don't think the namespace is correct here

tests/ExceptionTest.php
```diff
@@ -0,0 +1,17 @@
...
+
+use League\OAuth2\Server\Exception\OAuthServerException;
+
+class ExceptionTest extends \PHPUnit_Framework_TestCase
```
Sephster: Please can we change this to extend TestCase and add in a `use PHPUnit\Framework\TestCase;` statement?

Sephster:
I think we will need to update the documentation for this to illustrate proper use cases as well.

iansltx:
I'll see what I can do re: test changes later tonight. I want to say the test differences here were due to the lib being on an older PHPUnit version at the time, so I have an excuse for the PR being outdated, but am more than happy to do those quick fixes.

Once this is merged in, I'll work on a PR into the docs branch. Sound reasonable?

Sephster:
Hi @iansltx - thanks for this. Yeah, I realised we'd updated the PHPUnit version when I made the comment. Sorry I should have highlighted this in my initial response. Apologies for taking so long to actually pick up this PR. 

I think it would be good if we could get the docs PR and the feature to be released at the same time if at all possible but let me know if you think this won't be feasible for your timescales and we can probably just add an issue to track that this needs documenting. Thanks

iansltx:
Fair enough. Ended up a little more strapped for time than expected. Will try and get this...and docs changes...handled tomorrow.

Sephster:
Thanks @iansltx, much appreciated.

PR #709, opened 2017-03-01
Corrected DateInterval from 1 min to 1 month

toby-griffiths:


SoboLAN:
Question: why was this change done? Thank you.

Sephster:
I assume this change was made because the initial commit was a mistake. If you look at the comment above it states:

> // Enable the refresh token grant on the server with a token TTL of 1 month

Either the comment or the code had to change. I think having a refresh token with a TTL of 1 month is a more realistic scenario than one min so is a better illustration for our examples.

PR #711, opened 2017-03-02
Added getScopeIdentifiers method

jimi985:
This is meant to fix a code inconsistency where AuthCodes and AccessTokens getScopes() method is assumed to be an array of ScopeEntityInterfaces and also an array of strings.

In League\OAuth2\Server\GrantAuthCodeGrant::completeAuthorizationRequest():

`$this->encrypt(
    json_encode(
    [
...
        'scopes' => $authCode->getScopes(),`
`

This is where the scopes are encrypted in the AuthCode payload.  When they are unpacked in AuthCodeGrant::respondToAccessTokenRequest(), it is assumed that the scopes in the AuthCode payload are identifiers and not instances of ScopeEntityInterface:

`
foreach ($authCodePayload->scopes as $scopeId) {
                $scope = $this->scopeRepository->getScopeEntityByIdentifier($scopeId);
`

This needs to be the case as they are being json_encoded so they wouldn't be able to be unserialized into an instance of ScopeEntityInterface without some extra work and security vulnerabilities being introduced.

I added a new method League\OAuth2\Server\Entities\TokenInterface::getScopeIdentifiers() to handle converting ScopeEntityInterface[] into string[] to be used for payloads such as the AuthCode payload.  With this change, there isn't any need for TokenInterface::getScopes(), but I am using it in the getScopeIdentifiers method and I can see a reason for it to be needed elsewhere in the future.


jimi985:
Is there anything I need to do to have this pull request reviewed? This still seems to be an issue with the current master branch.

jimi985:
I hadn't realized at first that the ScopeEntityInterface implemented \JsonSerializable.  This was because I was using a framework entity class (Cakephp) to implement the ScopeEntityInterface which has already implemented JsonSerializable.  I agree with @iliepandia (See issue: #846) that this library should not depend on a specific implementation of jsonSerialize, if in fact it needs to be in a very specific format.  This also restricts the flexibility of implementation of this interface.

I think my changes in this pull request address this in a way that @iliepandia described.  By adding a method getScopeIdentifiers, that returns an array of scope identifier, this remedies this problem and also would be backwards compatible with any implementations of ScopeEntityInterface that were abiding to the specific format of jsonSerialize required by this library.

Sephster:
Hi @jimi985 - I'm going to try and discuss this with the other maintainers to get a second opinion on whether we need a new method or whether we can just improve our documentation.

I'm not that familiar with CakePHP but if you have an example of your implementation conflict that would be really useful in helping decide what is the best course of action here. Thanks for your help.

jimi985:
As requested, here's an example implementation of the ScopeEntityInterface in the CakePHP Framework (https://cakephp.org/)

Implement the ScopeEntityInterface, using the base Entity class.  CakePHP uses Active Record paradigm, where the Entity class represents a record in the database.
```
<?php
use Cake\ORM\Entity;

// https://github.com/thephpleague/oauth2-server
use League\OAuth2\Server\Entities\ScopeEntityInterface;

/**
 * OAuthScope Entity
 *
 * @property string $id
 * @property string $description
 */
class OAuthScope extends Entity implements ScopeEntityInterface
{
}
```

The Entity class' methods are mostly present in the EntityTrait
```
class Entity implements EntityInterface, InvalidPropertyInterface
{
    use EntityTrait;
```

The trait has the jsonSerialize() method implemented by default.  However, the phpleague/oauth2-server assumes that the jsonSerialize() method be implemented in a very specific way.  So this implementation of jsonSerialize() is incompatible with the oauth2-server.
```
/**
trait EntityTrait {
     * Returns an array with the requested properties
     * stored in this entity, indexed by property name
     *
     * @param array $properties list of properties to be returned
     * @param bool $onlyDirty Return the requested property only if it is dirty
     * @return array
     */
    public function extract(array $properties, $onlyDirty = false)
    {
        $result = [];
        foreach ($properties as $property) {
            if (!$onlyDirty || $this->isDirty($property)) {
                $result[$property] = $this->get($property);
            }
        }

        return $result;
    }

    /**
     * Returns the properties that will be serialized as JSON
     *
     * @return array
     */
    public function jsonSerialize()
    {
        return $this->extract($this->visibleProperties());
    }
```

While I can override the implementation of jsonSerialize, I would be forfeiting the use of the framework's implementation of jsonSerialize, when in fact the oauth2-server could take care of this independently from jsonSerialize().

Sephster:
Thanks for the example @jimi985 - that makes things a lot clearer. Cheers for taking the time to write that

simonhamp:
>This is meant to fix a code inconsistency where AuthCodes and AccessTokens getScopes() method is assumed to be an array of ScopeEntityInterfaces and also an array of strings.

Please can you indicate where it's expected to be an array of strings?

While I really do like the addition of a good convenience method, I feel like this ultimately doesn't solve your `jsonSerialize` problem. If it's to stay, I also believe that `getScopeIdentifiers` could be simplified.

1. Your problems around the `jsonSerialize` function stem from the fact that you're mixing an implementation of it serving one purpose, with a contract for it expecting an implementation for another purpose. From what I can tell, we can't easily solve that from the `oauth2-server` side without breaking backwards-compatibility and I'm not even sure that we should.

I'd recommend your Scope entity class doesn't directly extend the Cake `Entity` class or use the `Cake\Datasource\EntityTrait` trait. Simply implement the `ScopeEntityInterface` and you will have no conflict over the `jsonSerialize` method.

Then to connect it to your database without any issues, you can create your ActiveRecord model `OAuthScope` that *doesn't* implement `ScopeEntityInterface` but instantiates and manages instances of the class that does.

I think this separation of concerns will actually help later on too.

2. Couldn't `getScopeIdentifiers` more simply `return array_keys($this->scopes)`? Where are you planning to use this in your code? If you follow the advice from (1), do you actually need `getScopeIdentifiers` any more?

jimi985:
The scopes that are defined in the authcode payload are expected to be an array of Scope identifiers (strings) in https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php `respondToAccessTokenRequest()`:
```
$scopes = [];
foreach ($authCodePayload->scopes as $scopeId) {
    $scope = $this->scopeRepository->getScopeEntityByIdentifier($scopeId);
    if ($scope instanceof ScopeEntityInterface === false) {
        // @codeCoverageIgnoreStart
        throw OAuthServerException::invalidScope($scopeId);
        // @codeCoverageIgnoreEnd
    }
    $scopes[] = $scope;
}
```

This requires a very explicit implementation of jsonSerialize() for the class that implements ScopeEntityInterface.  The `$authCodePayload` ultimately originates from these lines in https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php `completeAuthorizationRequest()`:

```
$payload = [
    'client_id'             => $authCode->getClient()->getIdentifier(),
    'redirect_uri'          => $authCode->getRedirectUri(),
    'auth_code_id'          => $authCode->getIdentifier(),
    'scopes'                => $authCode->getScopes(),
    'user_id'               => $authCode->getUserIdentifier(),
    'expire_time'           => (new \DateTime())->add($this->authCodeTTL)->format('U'),
    'code_challenge'        => $authorizationRequest->getCodeChallenge(),
    'code_challenge_method' => $authorizationRequest->getCodeChallengeMethod(),
];
$response = new RedirectResponse();
$response->setRedirectUri(
    $this->makeRedirectUri(
        $finalRedirectUri,
        [
            'code'  => $this->encrypt(
                json_encode(
                    $payload
                )
            ),
            'state' => $authorizationRequest->getState(),
        ]
    )
);
```

Since the payload life cycle of encrypting and decrypting this payload is handled entirely within the library, I believe the logic for defining the structure of `$payload['scopes']` should be defined within the library itself.  This is the reason for creating this method.  I have forked this repository for the purposes of this change, and I know that this solves my particular issue.  Whether or not I have a class that wraps the Active Record representation of the scope, does not refute the point that the library requires that the ScopeEntityInterface be jsonSerialized in a strict manner.  This is clear if you try to alter the example implementation of the ScopeEntityInterface found here:

https://github.com/thephpleague/oauth2-server/blob/master/examples/src/Entities/ScopeEntity.php

You will find that if this returns more than just the identifier string, the dependent unit tests will fail.

If, for example, I wanted to add other properties to a scope, like an added timestamp, and I wanted this property to be represented when the entity is json_encoded, I would not be able to utilize the JsonSerializable interface with this entity, because this library requires that it be serialized only as the ScopeIdentifier string.

I don't see how this creates a backwards compatibility issue, as my suggested change for `getScopeIdentifiers()` utilizes an already existing method `getScopeIdentifier()`.  Because of this, it is my understanding that any previous implementations of ScopeEntityInterface should be compatible with this.

simonhamp:
I appreciate your reasoning on this. Thanks for continuing to take the time. After a lot of further digging, I feel like I've got a much clearer understanding of the problem.

All things considered, I have to stand by my suggestion that you should consider separating your `ScopeEntity` class from your storage mechanism, but I really do appreciate your working through the problem to find a solution that could benefit everyone.

I've tried working through some ideas, but I haven't found an alternative approach which is as elegant as your solution.

There are just a couple of things...

First:

>I don't see how this creates a backwards compatibility issue

Your change to [TokenInterface](https://github.com/thephpleague/oauth2-server/pull/711/files#diff-5ecd449f761d2e05c493796abc25fcf0) will be a breaking change for anyone who _doesn't_ also rely on `TokenEntityTrait`. This is acceptable from my point of view, but it does mean that this will be a major release, so it could still be a while before it sees daylight.

Secondly:

>You will find that if this returns more than just the identifier string, the dependent unit tests will fail.

This may be an issue with my setup, but I couldn't get any tests to fail with a different implementation of `jsonSerialize` in the ScopeEntity. This is possibly down to a lack of the right tests - @Sephster we should talk about this. Or maybe I'm missing something.

Because of this though, I've not been able to quickly replicate your scenario in a predictable way, so the caveat is that I'm still basing my understanding of this on what you've written and not on code.

jimi985:
Hi @simonhamp ,

Thank you for your feedback on this.  I remember now, since I investigated this some time ago, that the unit tests aren't adequately covering this scenario as can be seen in https://github.com/thephpleague/oauth2-server/blob/master/tests/Grant/AuthCodeGrantTest.php

```
$request = new ServerRequest(
            [],
            [],
            null,
            'POST',
            'php://input',
            [],
            [],
            [],
            [
                'grant_type'   => 'authorization_code',
                'client_id'    => 'foo',
                'redirect_uri' => 'http://foo/bar',
                'code'         => $this->cryptStub->doEncrypt(
                    json_encode(
                        [
                            'auth_code_id' => uniqid(),
                            'expire_time'  => time() + 3600,
                            'client_id'    => 'foo',
                            'user_id'      => 123,
                            'scopes'       => ['foo'],
                            'redirect_uri' => 'http://foo/bar',
                        ]
                    )
                ),
            ]
        );
        /** @var StubResponseType $response */
        $response = $grant->respondToAccessTokenRequest($request, new StubResponseType(), new \DateInterval('PT10M'));
```

You will find that the `respondToAccessTokenRequest` method is being tested using a faux `ServerRequest` object, and more specifically a synthetic `$authCodePayload`, where the scopes are being defined as `['foo']` instead of being derived from an actual code generated by the Grant class.  I believe it may be worth refactoring this test to adequately cover this scenario, otherwise this critical component of the library isn't adequately tested.  I'm not a fan of dependent unit tests, but this is one example where it might make sense.

I can see why now that my proposed change might be a BC break, but I would still appreciate if this was available in a future release.  Thank you for your time you have spent investigating this.

Sephster:
@jimi985 I've been working on trying to get this into the library and it is not easy. What I am struggling with is the terminology here.

A scope is just a string identifier. The scope entity interface has just one function called `getIdentifier()` which effectively returns the scope in question.

The `jsonSerialize()` function is converting an array of scopes into their corresponding strings. The problem I see at the moment is that `getScopeIdentifiers` just gets scopes according to the OAuth2 RFC. ScopeEntity is just a wrapper around this string. At the moment we have two functions:

* getScopeIdentifiers
* getScopes

Technically `getScopeIdentifiers` is getting the scopes as expected for transport in the OAuth spec so what does `getScopes` return? We could rename `getScopes` to `getScopeEntities` but this seems convoluted to me.

The json serializable implmentation elegantly deals with this issue in my opinion so the problem is, how do we integrate this with your Cake install?

Can you not have a function in your model which converts the DB record into a compatible entity? This would be as @simonhamp suggested in his original post:

> Your problems around the jsonSerialize function stem from the fact that you're mixing an implementation of it serving one purpose, with a contract for it expecting an implementation for another purpose. From what I can tell, we can't easily solve that from the oauth2-server side without breaking backwards-compatibility and I'm not even sure that we should.

You could create a class that implements `ScopeEntityInterface` and doesn't extend Cake's Entity class. You could then return this class from a function inside your Cake Entity as follows:

```php
public function getScopeEntity()
{
    return new ScopeEntity($this->scope);
}
```

This would mean you can still get compatible entities without us needing to change/remove our implementation of `jsonSerializable` for use with the server.

I think this would be the cleanest solution as I think the changes proposed here risk making the package unnecessarily complicated.

Please let me know your thoughts on this and if there are any reasons as to why you can't implement a solution as detailed above. 

Thanks for your continued help with this.

jimi985:
> The json serializable implmentation elegantly deals with this issue in my opinion so the problem is, how do we integrate this with your Cake install?

I don't think that the jsonSerializable interface elegantly deals with this as it doesn't work unless you explicitly implement this method as seen in examples/src/Entities/ScopeEntity.php:

```
public function jsonSerialize()
    {
        return $this->getIdentifier();
    }
```

It forces the implementer of the OAuth2 server to implement this method in precisely this way and doesn't work otherwise.  If the implementer has no choice but to implement this method this way, why isn't it in the library itself?  That's what my code changes were attempting to solve.  After reviewing this, I'm not even sure how it breaks backwards compatibility, as the array of scopeidentifiers that is later json_encoded will be generated in an identical matter to the jsonSerialize method, assuming the jsonSerialize method is implemented as it is in the above snippet.

Ultimately, I think my changes remove unnecessary code that implementers of this library should not have to worry about, properly encapsulating the logic to package the different token payloads.

Sephster:
I agree that we could move `jsonSerialize` from the trait and into the class as you are right that no other implementation would work. This is probably something we will look to do as it doesn't currently hold much value being inside a trait.
If we did this, it would still not resolve your problem without you changing the way your Cake implementation works. Assuming we move the `jsonSerialize` implementation, would the implementation I outlined above resolve your issue?

jimi985:
I think relying on the interaction between PHP's `json_encode` and `jsonSerializable` interface can cause problem's for future implementers, especially because the library is not providing a concrete implementation of ScopeEntityInterface.  If they happen to implement jsonSerialize in a different way than the above cited example, they will be faced with the same problems that I had to deal with.

I also think this is more consistent with the logic that is implemented in the construction of the token payload:

```
json_encode(
    [
        'client_id'        => $this->accessToken->getClient()->getIdentifier(),
        'refresh_token_id' => $this->refreshToken->getIdentifier(),
        'access_token_id'  => $this->accessToken->getIdentifier(),
        'scopes'           => $this->accessToken->getScopes(),
        'user_id'          => $this->accessToken->getUserIdentifier(),
        'expire_time'      => $this->refreshToken->getExpiryDateTime()->getTimestamp(),
]
)
```

All of the items packaged in this payload explicitly access an identifier through a `getIdentifier` method except for the scopes property and none of them rely on an implementation of the `jsonSerializable` interface.  Using `getScopeIdentifiers` I think it cleans this up as it is a bit of a gotcha that it happens to be the only property relying on this interaction.

Sephster:
Hey @jimi985 - apologies for the length of time this has been sitting dormant. After a lot of thought, I've decided that the use case for changing the implementation here isn't strong enough so am not going to go forwards with this pull request.

I think this issue would be much better addressed by providing a scope entity trait which provides a concrete implementation of the json serializable function. I haven't tested this out yet or looked into its feasibility but I am hoping this will be a better route forwards.

Thanks for your efforts with this PR and sorry we did not end up going forwards with your implementation

PR #678, opened 2016-10-31
Added Zend diactoros library dependency to the examples

pcambra:
I was trying the examples an got a Fatal error: "Class 'Zend\Diactoros\Stream' not found in oauth2-server/examples/public/client_credentials.php on line 72.
Adding this explicit dependency and running composer update fixes it.

PR #712, opened 2017-03-14
Missing Zend library

ken-master:
Missing Zend library that causes an error below:

::1:37754 [500]: /client_credentials.php/access_token - Class 'Zend\Diactoros\Stream' not found in /data/www/html/sites/oauth2-server/examples/public/client_credentials.php on line 72


Sephster:
Thank you for your submission. Unfortunately this is a Duplicate of #678 so I will mark this a closed. Thanks for drawing this to our attention though.

PR #727, opened 2017-03-30
Update OAuthServerException.php

devonliu02:


Sephster:
Hi @devonliu02. I can't see any use case for this at the moment so I'm not seeing an immediate requirement for this change. If I have missed something, please let me know and I will take a second look but I don't think this should be merged in at this time. Thanks for your submission.

PR #731, opened 2017-04-10
Fix Segfaults: treat rsa candidate as one string

m-bymike:
- save pcre backtracking work
- prevent segfaults with long private keys

(though contributing guidelines state to open PR to dev, I was unable to find that branch)

### DIFF
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -14,7 +14,7 @@ namespace League\OAuth2\Server;
 class CryptKey
 {
     const RSA_KEY_PATTERN =
-        '/^(-----BEGIN (RSA )?(PUBLIC|PRIVATE) KEY-----\n)(.|\n)+(-----END (RSA )?(PUBLIC|PRIVATE) KEY-----)$/';
+        '/^(-----BEGIN (RSA )?(PUBLIC|PRIVATE) KEY-----\n).+(-----END (RSA )?(PUBLIC|PRIVATE) KEY-----)$/s';
 
     /**
      * @var string


m-bymike:
btw, a `\n` doesn't match CRLF line breaks, should I add that too?

alexbilbie:
Sorry for the very late response to this - how can I best test it?

m-bymike:
The segfaults with the previous expression can still be reproduced using php 5.5 / 5.6 on the command line (make sure to replace php56 w/ the appropriate executable):

**Previous version of regular expression:**
```bash
openssl genrsa 2048 | php56 -r "const RSA_KEY_PATTERN = '/^(-----BEGIN (RSA )?(PUBLIC|PRIVATE) KEY-----\n)(\n|.)+(-----END (RSA )?(PUBLIC|PRIVATE) KEY-----)$/'; echo (preg_match(RSA_KEY_PATTERN, fread(STDIN, 2048)) === 1 ? 'MATCH' : 'NO MATCH'), PHP_EOL;"
```

This is the output I get:
```
Generating RSA private key, 2048 bit long modulus
........................................................+++
.................+++
e is 65537 (0x10001)
[1]    1914 done                openssl genrsa 2048 | 
       1915 segmentation fault  php56 -r 
```

**New version of regular expression:**
```bash
openssl genrsa 2048 | php56 -r "const RSA_KEY_PATTERN = '/^(-----BEGIN (RSA )?(PUBLIC|PRIVATE) KEY-----\n).+(-----END (RSA )?(PUBLIC|PRIVATE) KEY-----)$/s'; echo (preg_match(RSA_KEY_PATTERN, fread(STDIN, 2048)) === 1 ? 'MATCH' : 'NO MATCH'), PHP_EOL;"
```
And the output here is:
```
Generating RSA private key, 2048 bit long modulus
..............................+++
...............+++
e is 65537 (0x10001)
MATCH
```

**Note:** The segfault only happens w/ php 5.5/5.6 and is not reproducible w/ php7.*.

The expression still matches the same requirements, hence no updated tests.

m-bymike:
Closing PR, changes included in #805 with commit f79d3f2.

PR #739, opened 2017-05-17
fix Stream argument 1 type is wrong

jk2K:
fix Argument 1 (stream) is string but \GuzzleHttp\Psr7\Stream::__construct() takes resource defined


Sephster:
This looks good to me. Thanks for spotting this.

PR #744, opened 2017-06-16
Fix PKCE implementation with strict verifications

erickjth:
**Summary**
The goal is to match with rfc7636 specifications strictly for AuthCode GrantType.

**Fixes**
- Verify format of the code_challenge and code_verifier for E256 method.
- Fix hash comparison. Codes must be encoded using  BASE64URL-ENCODE.
- Fix invalid `code_challenge_method` key when encrypting the JSON for a new code.
- Update Unit tests.


alexbilbie:
Thanks for this update - please could you re-review it against the current `master` branch?

erickjth:
Ok, I check it against master, the only thing missing is checking the code verifier such as code challenge check. This verifications is needed.

mojojoji:
Is this ready to be merged? Challenges generated by the client (in node js) is always failing due to this.

sgomez:
+1 to this. This feature is useless with third-party libraries until this issue is solved.

danut007ro:
+1 for this.
I also made the PKCE implementation optional in https://github.com/thephpleague/oauth2-server/pull/817

danut007ro:
Since PKCE implementation never worked, should it be enabled by default?

Sephster:
Sorry all. I will pick this up closer to Christmas for review and merge in if all is ok as I realise this is a problem that needs addressing soon. Thanks for your patience

tests/Grant/AuthCodeGrantTest.php
```diff
@@ -34,9 +34,21 @@ class AuthCodeGrantTest extends TestCase
...
+    /**
+     * @var string Valid generated code challenge using a proper code verifier.
+     */
+    protected $codeChallenge;
+
     public function setUp()
     {
         $this->cryptStub = new CryptTraitStub;
+        $this->codeVerifier = rtrim(strtr(base64_encode(random_bytes(32)), '+/', '-_'), '=');
+        $this->codeChallenge = rtrim(strtr(base64_encode(hash('sha256',$this->codeVerifier, true)), '+/', '-_'), '=');
```
Sephster: Please can this be changed to match the current implementation in master so that we hash after base 64 encoding instead of before:

`hash('sha256', strtr(rtrim(base64_encode($codeVerifier), '='), '+/', '-_'))`

Sephster:
Hi @erickjth - Thanks for your commit and apologies for the delay in getting to this. I think there only needs to be the one minor change to this listed in the review and it should be good to merge in. If you could make the change I will be happy to merge this in as soon as completed. Many thanks.

erickjth:
HI @Sephster, the change is done. Thanks for reviewing this PR.

tests/Grant/AuthCodeGrantTest.php
```diff
@@ -34,9 +34,25 @@ class AuthCodeGrantTest extends TestCase
...
+
     public function setUp()
     {
         $this->cryptStub = new CryptTraitStub;
+
+        // [RFC 7636] Appendix B.  Example for the S256 code_challenge_method
+        // $this->codeVerifier = 'dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk';
+        $this->codeVerifier = strtr(rtrim(base64_encode(random_bytes(32)), '='), '+/', '-_');
+        // $this->codeChallenge = 'E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM';
+        $this->codeChallenge = strtr(rtrim(base64_encode(hash('sha256', $this->codeVerifier, true)), '='), '+/', '-_');
```
sgomez: I disagree with this change. You are writing the algorithm that you must probe than works inside the tests. You have a know input and a know output, that's all you need to do unit test.

Sephster: I agree with @sgomez here. I will make the change

jacobweber:
Uh oh. Does this mean that PKCE won't work at all in version 6.x (and hence PHP 5.6)? Was hoping to use it in a project.

Sephster:
@jacobweber, the main reason we have dropped 5.6 support is because we are now using PHPStan. The current vesion 7 will still work fine for PHP 5.6 but we cannot guarantee future releases will run on 5.6. We won't be backporting this fix to version 6.x at this time

jacobweber:
Thanks! I'll give it a try.

Issue #748, opened 2017-06-30
Replace array_key_exists by isset on src/Grant/ImplicitGrant.php, which is faster


dmelo:
isset is faster. It might be a small change, but it would be an improvement.

alexbilbie:
Am closing this to move discussion to #749

PR #749, opened 2017-06-30
Replaces array_key_exists by isset, which is faster, on ImplicitGrant.

dmelo:
Related to issue #748 .

### DIFF
--- a/src/Grant/ImplicitGrant.php
+++ b/src/Grant/ImplicitGrant.php
@@ -96,7 +96,7 @@ class ImplicitGrant extends AbstractAuthorizeGrant
     {
         return (
             isset($request->getQueryParams()['response_type'])
-            && 'token' === $request->getQueryParams()['response_type']
+            && $request->getQueryParams()['response_type'] === 'token'
             && isset($request->getQueryParams()['client_id'])
         );
     }
--- a/src/Grant/ImplicitGrant.php
+++ b/src/Grant/ImplicitGrant.php
@@ -95,8 +95,8 @@ class ImplicitGrant extends AbstractAuthorizeGrant
     public function canRespondToAuthorizationRequest(ServerRequestInterface $request)
     {
         return (
-            array_key_exists('response_type', $request->getQueryParams())
-            && $request->getQueryParams()['response_type'] === 'token'
+            isset($request->getQueryParams()['response_type'])
+            && 'token' === $request->getQueryParams()['response_type']
             && isset($request->getQueryParams()['client_id'])
         );
     }


alexbilbie:
Thanks for your pull request.

Do you have any benchmarks to demonstrate the speed improvement?

dmelo:
Hi. I've done a benchmark a while ago http://diogomelo.net/blog/15/php-array_key_exists-vs-isset . Running the code on my machine today (with php 7.1.5) i still get isset running about 3.5x faster than array_key_exists:

```
Similar behaviour: 
/home/dmelo/v.php:6:
bool(true)
/home/dmelo/v.php:7:
bool(true)
Differences:
1. isset works with other variables, besides arrays.
/home/dmelo/v.php:12:
bool(true)
2. isset return false when value is null
/home/dmelo/v.php:17:
bool(false)
3. Speed
array_key_exists: 1.0082581043243 seconds 
isset: 0.28244996070862 seconds 
```

There are other banchmarks. In general, they come to the same conclusion: unless you have a reason to use array_key_exists you should opt for isset, which is faster.

Other posts related to the subject:

* https://ilia.ws/archives/247-Performance-Analysis-of-isset-vs-array_key_exists.html
* https://stackoverflow.com/questions/3210935/difference-between-isset-and-array-key-exists
* http://juliusbeckmann.de/blog/php-benchmark-isset-or-array_key_exists.html

Sephster:
Closing this PR to reopen and trigger a new Travis CI build as the Travis CI YAML file has changed upstream since originally posted. This PR is being built and failing against HHVM and PHP 5.5

Sephster:
Reopening to trigger a new Travis CI build

Sephster:
Thanks @dmelo - Change looks good. Would it be possible to change your use of Yoda conditions here and revert the comparison on the _token_ string? to be `$request->getQueryParams()['response_type'] === 'token'` instead? 

Convention in the package to date has not used Yoda conditions so I think it would be good to keep consistency. Many thanks

pounard:
Beware that changing array_key_exists() for isset() also changes the behavior: array_key_exists() will return true when value is null, whereas isset() returns false if value is null.

For security concerns, you should ensure it does makes the check being too loose.

dmelo:
@pounard, I was aware of that, when I proposed the change. In the context it was applied it is OK to have isset() returning false when the value for that key is null.

@Sephster , always tought Yoda condition was a good practice, but saw that PSR-2 additions advice against it. Have just reversed the condition order on commit 203be5ca20de4a0a3520513b7507d7e2dc5fc0f8 .

pounard:
@dmelo Yoda conditions are good ! I do not always respect standards when it is for readability or safety :)

sergeyklay:
👎  for Yoda conditions. Yoda conditions may make code harder to read for some developers. Part of your job as a software professional is to write quality code. You want your code to be as readable as possible.

alexbilbie:
Please remove the Yoda conditions

dmelo:
@alexbilbie I did remove it on commit 203be5ca20de4a0a3520513b7507d7e2dc5fc0f8, which is also on this PR.

Sephster:
Thanks @dmelo - cheers for removing the Yoda condition

PR #755, opened 2017-07-03
Use null comparison instead of is_null()

Lctrs:


alexbilbie:
Hello,

Thank you for your contribution - please could you explain the changes you've made?

Thank you

Lctrs:
Hi,

I replaced all the `is_null()` calls by `null ===`.
It's just what we can call a micro-optimization. Doing this comparison is always more efficient that calling the function for the same functionality.

Sephster:
Hi @Lctrs - Thanks for your pull request. After discussions with the other maintainers we've decided not to merge in this pull request. While using is_null() is slower than `=== null`, this is in terms of microseconds so we feel it would have little impact on the package. Thanks again for your time and apologies we did not go forwards with your contribution this time.

PR #768, opened 2017-07-23
Extrapolate getAccessToken and make accessTokenRepository protected

terence1990:
This makes it a lot easier to extend BearerTokenValidator without
having to write too much code, currently to extend one must declare
__construct to store accessTokenRepository as it is currently private
and the retrieve access_token from queryParams one currently has to
include the entire validateAuthorization and all of the external
dependancies all over again

### DIFF
--- a/src/AuthorizationValidators/BearerTokenValidator.php
+++ b/src/AuthorizationValidators/BearerTokenValidator.php
@@ -24,7 +24,7 @@ class BearerTokenValidator implements AuthorizationValidatorInterface
     /**
      * @var AccessTokenRepositoryInterface
      */
-    private $accessTokenRepository;
+    protected $accessTokenRepository;
 
     /**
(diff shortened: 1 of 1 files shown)


ZoozyCN:
@alexbilbie 
Retrieve access_token from other than authorization header is commonly used, can we merge this?

Sephster:
Hi @terence1990 - apologies for the long time in getting back to you regarding this. I feel that adding a `getAccessToken` method to our BearerTokenValidator class wouldn't be appropriate as it doesn't align with the purpose of this class which to simply validate the bearer token.

You had mentioned that you would like to extend this class but are having difficulties. If you could outline the purpose of the extension, we could look at this again but I feel that adding a way to directly access the JWT in this class isn't appropriate in its current form.

I will close this PR but if you have any further comments or would like to outline your motivations/intentions with extending this class I will be happy to revisit this. Thanks for your time

PR #774, opened 2017-07-29
Fix #771 - Use ? as separator for query string

### LINKED ISSUE #771: [Implicit Grant] wrong uri separator for query string?
Hi,

commit 251190d8282cda89a37f688e172776153df811f6 added a `#` as the query url separator https://github.com/thephpleague/oauth2-server/commit/251190d8282cda89a37f688e172776153df811f6#diff-4b262d46c2b822eb2ec98cd6fafaec1dR204

When redirected back to `redirect_uri`, the parameters are not accessible via `$_GET`.
Url are formed as `http://192.168.125.64/auth#access_token=eyJ0eX[...]wv5NNg&token_type=bearer&expires_in=36000`

I would have expected to have a '?' instead (which is the default value for `makeRedirectUri()`.

Is this voluntary?

kumy:


kumy:
Unrelevant.

Issue #760, opened 2017-07-07
"Ensure the server is the exclusive owner of the key" is wrong


gemal:
the implementation is wrong
if ($keyPathPerms !== '600') {

you cant say that this is the perfect permission.
you might have set up the file with permission to the group

which you then have
'660'

gemal:
I'm talking about this commit:
https://github.com/thephpleague/oauth2-server/commit/2f8de3d2302beb490abb9475cf426148801c25c4

how can I help in fixing this?

alexbilbie:
Please feel free to submit a pull request

Lavoaster:
@alexbilbie Is this supposed to be for _all_ oauth keys, or just the ones passed via a string and saved to `/tmp`

libasoles:
@gemal Have you created that PR? If not, I think I will.

andersonamuller:
Does anyone have a fix for this case?
- Container PHP-FPM runs on `www-data`;
- OAuth keychain are Kubernetes secrets mounted and owned by root;

AFAIK, you cannot change that last part.

See also https://github.com/laravel/passport/issues/418#issuecomment-312866308

alexbilbie:
@andersonamuller could you copy the keys out of Kubernetes secrets mount into a temp directory when the container starts using an entrypoint script? You might need to use UID and GID to manage privileges

andersonamuller:
`USER` defined in the container is already `www-data` so I cannot create a script to call a `chown` the user does not have permissions.

Lavoaster:
@alexbilbie This should be the job of the sysadmin, not this package.

gemal:
I still think just check for `if ($keyPathPerms !== '600') {`
is wrong

you could have a perfectly secure system without the permissions being `600`

JEDIBC:
An easy workaround is to use the values of the keys and not the path to the key files.

JEDIBC:
The recent commit 80fc8e654b6ab6ba66000ddd7b95f8d7203c2443 won't change the problem for symfony users as errors are turned into exception by the framework.

Perhaps a modification of the constructor like below would allow more flexibility in the use of the keys :

```php
/**
 * CryptKey constructor.
 *
 * @param string      $keyPath
 * @param null|string $passPhrase
 * @param bool        $checkFileMode
 * @param int         $fileMode
 */
public function __construct($keyPath, $passPhrase = null, $checkFileMode = true, $fileMode = 0600)
{
[...]

PR #776, opened 2017-08-01
Removed chmod from CryptKey and add toggle to disable checking

yannickl88:
First of all, please do not touch my files... I do not think it's the responsibility of a package to set file permissions. It is the responsibility of your deployment pipeline or developers themselves.

While I appreciate the intent that was done in 2f8de3d2302beb490abb9475cf426148801c25c4. This is breaking for us in both our development and our production environments. And according to #760 for more people.

With this PR I would like to propose two things:
- First and foremost, remove any code that modifies files. However, do trigger a `USER_NOTICE` when not set to `600`.
- Option to skip the check. This can be useful for DEV environments.

A bit of context for our case:

Our PROD setup is as follows:
We have 4 users: `deploy`, `worker`, `scheduler` and `apache`. All four are part of the `deployment` group. So our code is chowned as `deploy:deployment`. So while `600` might seems like a good assumption, it will prevent our `apache` user from reading the key, thus breaking the web requests.

Our DEV setup is a bit different:
We have our own users, `yannickl88` in my case, and `www-data` for the web user. Files are chmodded `666` so both users can read everything. (we have a umask 0000 everywhere). While not the best setup, it's just development. In this case `660` is even not enough, but a warning would be justified. However, we would like to silence this in our DEV environment since Symfony converts the `USER_NOTICE` to an exception, making the lib unusable.

### DIFF
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -29,8 +29,9 @@ class CryptKey
     /**
      * @param string      $keyPath
      * @param null|string $passPhrase
+     * @param bool        $keyPermissionsCheck
      */
-    public function __construct($keyPath, $passPhrase = null)
+    public function __construct($keyPath, $passPhrase = null, $keyPermissionsCheck = true)
     {
         if (preg_match(self::RSA_KEY_PATTERN, $keyPath)) {
             $keyPath = $this->saveKeyToFile($keyPath);
@@ -44,21 +45,15 @@ class CryptKey
             throw new \LogicException(sprintf('Key path "%s" does not exist or is not readable', $keyPath));
         }
 
-        // Verify the permissions of the key
-        $keyPathPerms = decoct(fileperms($keyPath) & 0777);
-        if ($keyPathPerms !== '600') {
-            // Attempt to correct the permissions
-            if (chmod($keyPath, 0600) === false) {
-                // @codeCoverageIgnoreStart
-                trigger_error(
-                    sprintf(
-                        'Key file "%s" permissions are not correct, should be 600 instead of %s, unable to automatically resolve the issue',
-                        $keyPath,
-                        $keyPathPerms
-                    ),
-                    E_USER_NOTICE
-                );
-                // @codeCoverageIgnoreEnd
+        if ($keyPermissionsCheck === true) {
+            // Verify the permissions of the key
+            $keyPathPerms = decoct(fileperms($keyPath) & 0777);
+            if (in_array($keyPathPerms, ['600', '660'], true) === false) {
+                trigger_error(sprintf(
+                    'Key file "%s" permissions are not correct, should be 600 or 660 instead of %s',
+                    $keyPath,
+                    $keyPathPerms
+                ), E_USER_NOTICE);
             }
         }
 
--- a/tests/AuthorizationServerTest.php
+++ b/tests/AuthorizationServerTest.php
@@ -26,6 +26,13 @@ use Zend\Diactoros\ServerRequestFactory;
 
 class AuthorizationServerTest extends \PHPUnit_Framework_TestCase
 {
+    public function setUp()
+    {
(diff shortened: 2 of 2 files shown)


yannickl88:
If okay, I would like to loosen the permissions to `660` instead of `600` to support cases like ours (and that of #760)

alexbilbie:
I'm happy to merge this if you could please remove the disabling of the check.

alexbilbie:
Actually ignore my comment about disabling the check.

Instead I'd be happy with an API that looks like this:

```php
$privateKey = new CryptKey('file://path/to/private.key', 'passphrase');
$privateKey->acknowledgeSecurityRisk(); // this disables the check
```

yannickl88:
I would like to, but that might be a bit of a bigger change than possible. The check is done in the constructor. So in order for that code to work the validation of the key needs to be moved.

This will change the behavior quite a bit, because checking so will either be done by another class or when one of the public methods is called, changing the place where the error might be triggered (if at all).

alexbilbie:
Okay I will re-evaluate the `acknowledgeIncorrectKeyPermissions`.

Please can you address the comments I've made on the commit and I will merge this PR.

yannickl88:
which comments? The one about removing the disabling of the check?

alexbilbie:
@yannickl88 see my comments in line here https://github.com/thephpleague/oauth2-server/pull/776/files

yannickl88:
I'm not seeing any comments, did you publish them?

alexbilbie:
Apparently not...

yannickl88:
np! should be fixed now.

valeryan:
So any information on how one would get laravel passport to use this new toggle for the permission check?  I just ran a composer update and now my whole dev environment has come crashing down.  I am using bash on windows and you cannot chmod files because it ignores it. https://github.com/Microsoft/BashOnWindows/issues/81

yannickl88:
There is a third constructor argument for the keys. Setting this to false will disable the check.

The [`AuthorizationServer`](https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationServer.php#L87) has 2 arguments for the public and private keys which can be both a `string` or a `CryptKey`. So the easiest thing to do is use the `CryptKey` and make sure to set the third argument to `false`.
You do the same for the [`ResourceServer`](https://github.com/thephpleague/oauth2-server/blob/master/src/ResourceServer.php#L44).

How to do this with Laravel passport I do not know, we use Symfony. I recommend making a PR there or try to extend it yourself to use this behavior.

cringer:
I don't know if this is because I use the built in PHP development server, but my files in windows always report 0666, and folder report 0777.  If specific couldn't we have the package automatically disable checking/modification using a php_uname('s') check?

valeryan:
@cringer bash on windows does not support chmod of files that are stored on the ntfs file system. see https://github.com/Microsoft/BashOnWindows/issues/81

cringer:
Yeah I get that.  My point is that the package is attempting to check and/or modify the permissions on the keys.  Since the specified values will always be reported back from a windows system the package will always crash and burn.  

While a 3rd constructor has been added to disable this check I was just wondering why it doesn't do a check based on the OS anyway.   As a cross platform package, it would seem like it would be a better solution to ignore checking and modification completely on Windows.

valeryan:
@cringer I haven't actually dug into how they are doing the permissions check, maybe it does check the OS but because you are running bash for windows the script execution is happening in a *nix environment.

cringer:
@valeryan I don't really think they are checking.  If the check was done through PHP it a check to phpuname('s') would return the Operating System of the server running the php.  It doesn't matter if the development server was started using command prompt or git bash this will always be Windows.  Specifically `php_uname('s')` would return `Windows NT`.

vibrantBits:
For backwards and crossplatform compatibility $keyPermissionsCheck should be false by default.

PR #777, opened 2017-08-02
[BC Break] Fixes invalid code challenge method payload key name

hhamon:
I guess this change might be a BC break for existing and active authorization tokens when they're validated by the server. The good thing is that an authorization token has a very short expiration time and is used once to request an access token.


alexbilbie:
Thanks for submitting this 👍

PR #790, opened 2017-09-27
Use request redirect uri on invalid scope response

benito103e:
RFC 6749, section 5.2 :
Server should use redirect_uri set on request to return error except for error concerning redirection URI

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -242,12 +242,7 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
             }
         }
 
-        $scopes = $this->validateScopes(
-            $this->getQueryStringParameter('scope', $request),
-            is_array($client->getRedirectUri())
-                ? $client->getRedirectUri()[0]
-                : $client->getRedirectUri()
-        );
+        $scopes = $this->validateScopes($this->getQueryStringParameter('scope', $request), $redirectUri);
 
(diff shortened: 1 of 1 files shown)


Sephster:
This change would force the redirect_uri to be required on the request. I don't think this has to be set as we can fall back to a client registered redirect_uri if it isn't. 

I couldn't find the section you are referring to in 5.2 about redirection. Would you be able to direct quote it here so I can find which portion of the RFC you are addressing? Many thanks for your PR.

benito103e:
1° redirect_uri : you're right, me should use redirect_uri provided or client registered redirect uris.

2° I don't find the reference to my quote, but on Authorizarion code grant, we can read : 
https://tools.ietf.org/html/rfc6749#section-4.1.2.1

> If the resource owner denies the access request or if the request
   fails for reasons other than a missing or invalid redirection URI,
   the authorization server informs the client by adding the following
   parameters to the query component of the redirection URI using the
   "application/x-www-form-urlencoded" format, per Appendix B:

Sephster:
Hi @benito103e. Thanks for the quick response. I'm sorry but I don't think redirecting to the request redirect_uri provided will always be the correct approach and I can't find a justification for this change in the RFC at the moment so I think we won't be able to merge in this request.

If you are able to justify the reason for this change with a clear example where we should use the request redirect_uri exclusively in the context of a scope validation error, I will be happy to re-review this. Thanks again for your efforts

PR #795, opened 2017-10-14
'Undefined property' in case of a wrong auth code payload

x0st:
I encountered the error `Undefined property: stdClass::$auth_code_id` when passed a refresh token to the field `code` instead of a real auth code.

image

My code checks if the auth code's payload is correct: all the necessary fields are checked for presence.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -81,6 +81,9 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
         // Validate the authorization code
         try {
             $authCodePayload = json_decode($this->decrypt($encryptedAuthCode));
+
+            $this->validateAuthCodePayload($authCodePayload);
+
             if (time() > $authCodePayload->expire_time) {
                 throw OAuthServerException::invalidRequest('code', 'Authorization code has expired');
             }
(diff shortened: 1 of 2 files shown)


src/Grant/AuthCodeGrant.php
```diff
@@ -178,6 +181,29 @@ public function respondToAccessTokenRequest(
...
+            isset($authCodePayload->client_id),
+            isset($authCodePayload->redirect_uri),
+            isset($authCodePayload->scopes) && is_array($authCodePayload->scopes),
+            isset($authCodePayload->user_id)
+        ];
+
+        if (in_array(false, $errors, true)) {
+            throw OAuthServerException::invalidRequest('code');
+        }
+    }
```
sergeyklay: Looks like a pasta code here.

1. As you may know `json_decode` may return NULL  if the cannot be decoded or if the encoded data is deeper than the recursion limit. So, why do we need type hinting here?

2. I think the code may be more simple and obviously faster:

```php
$validCodePayload = 
            isset($authCodePayload->expire_time) &
            isset($authCodePayload->auth_code_id) &
            isset($authCodePayload->client_id) &
            isset($authCodePayload->redirect_uri) &
            isset($authCodePayload->scopes) &&
            is_array($authCodePayload->scopes) &
            isset($authCodePayload->user_id);

if ($validCodePayload === false) {
    throw OAuthServerException::invalidRequest('code');
}
```

3. I wold like to get hint - what exactly is missing

4. IMO is it better to validate JSON by using JSON Schema

x0st: Thanks for reviewing. I will fix it ASAP

sergeyklay: I think that 3 and 4 aren't mandatory, although for me they are still important.

Sephster:
Thanks for this PR. Sorry that you hit this issue. I agree, the system should probably be checking for this property (and others) before using it. I have a couple of thoughts and observations that might help:

Have you considered using JSON Schema for this? I think it might be a good fit for validating the format of the payload and the [League JSON guard package](https://github.com/thephpleague/json-guard) would be suitable for this purpose.

I think using a schema to validate against would make this a cleaner and clearer implementation if it can be achieved so it would be good to explore the feasibility of this if not done already. Would you be willing to take a look at this? Thanks again for your PR

x0st:
@Sephster What do you think about [justinrainbow/json-schema](https://github.com/justinrainbow/json-schema)? I think it is easier to use and more understandable.

Issue #793, opened 2017-10-11
Return `scope` as a token response param


chervand:
According to https://tools.ietf.org/html/rfc6749#section-5.1

> scope
         OPTIONAL, if identical to the scope requested by the client;
         otherwise, REQUIRED...

and https://tools.ietf.org/html/rfc6749#section-3.3

> The authorization server MAY fully or partially ignore the scope
   requested by the client, based on the authorization server policy or
   the resource owner's instructions.  If the issued access token scope
   is different from the one requested by the client, the authorization
   server MUST include the "scope" response parameter to inform the
   client of the actual scope granted. If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope.  The authorization server SHOULD
   document its scope requirements and default value (if defined).

Is it possible to return `scope` value as a request param by default to notify clients about actually granted scopes (in case default or omitted ones) without need of parsing JWT?

DewaldBodenstein:
@chervand have you found a way to accomplish this?

chervand:
@DewaldBodenstein, I believe, adding it to a `League\OAuth2\Server\ResponseTypes\BearerTokenResponse` `$responseParams` should be enough.

DewaldBodenstein:
Thanks, got it working.

simonhamp:
@chervand That's nice. I think in most implementations, if the client requests scopes that aren't supported, the server should respond with an `invalid_scope` error. However, this is really useful for situations where a scope is valid, but for some other reason has been denied (e.g. authorisation).

And as you rightly mention, it follows spec. In fact, I'd go so far as to say that it's worth having in the response regardless.

Do you fancy making a pull request for this?

chervand:
@simonhamp yes, sure

PR #800, opened 2017-10-25
scope param added to a bearer token response

chervand:
#793 scope param added to a bearer token response

### DIFF
--- a/src/ResponseTypes/BearerTokenResponse.php
+++ b/src/ResponseTypes/BearerTokenResponse.php
@@ -13,6 +13,7 @@ namespace League\OAuth2\Server\ResponseTypes;
 
 use League\OAuth2\Server\Entities\AccessTokenEntityInterface;
 use League\OAuth2\Server\Entities\RefreshTokenEntityInterface;
+use League\OAuth2\Server\Entities\ScopeEntityInterface;
 use Psr\Http\Message\ResponseInterface;
 
 class BearerTokenResponse extends AbstractResponseType
@@ -30,6 +31,11 @@ class BearerTokenResponse extends AbstractResponseType
             'token_type'   => 'Bearer',
             'expires_in'   => $expireDateTime - (new \DateTime())->getTimestamp(),
             'access_token' => (string) $jwtAccessToken,
+            'scope' => implode(" ", array_map(
+                function (ScopeEntityInterface $scopeEntity) {
+                    return $scopeEntity->getIdentifier();
+                }, $this->accessToken->getScopes()
+            ))
         ];
 
         if ($this->refreshToken instanceof RefreshTokenEntityInterface) {
--- a/tests/ResponseTypes/BearerResponseTypeTest.php
+++ b/tests/ResponseTypes/BearerResponseTypeTest.php
@@ -28,14 +28,17 @@ class BearerResponseTypeTest extends \PHPUnit_Framework_TestCase
         $client = new ClientEntity();
         $client->setIdentifier('clientName');
 
-        $scope = new ScopeEntity();
-        $scope->setIdentifier('basic');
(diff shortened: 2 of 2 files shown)


Sephster:
As per the spec, the returning of scopes should be optional if there are no changes and mandatory if there have been alterations to the scopes requested.

This change would make the server always return the scope. I think it would be better to put this decision into the hands of the implementer instead of always returning a scope to avoid the server being opinionated on this matter. 

Is it possible to change this so the scopes will only be returned an option has been set on the server to dictate this. That would make the returning of the scopes more in line with the RFC in my opinion

chervand:
Since `AuthorizationServer`'s constructor accepts `ResponseTypeInterface` instance, it is possible to add such an option to `AbstractResponseType`. But I think it should defaults to returning scopes, because they are conditionally required and will not cause any issues in most cases. If you agree, I'll update the PR. BTW, I think such change should be documented on how to disable returning scopes.

Sephster:
Hi @chervand -  I think it would be better to have them not set by default. In most situations, if we do not return a scope, the implementer can assume they are unchanged. 

If we have these on by default, this adds overhead for each request. While probably negligible in most situations, I think it is better we err on the side of caution and only send back the scopes if they have been requested by the implementer, leaving them off by default. 

We should aim to ensure this change is also backwards compatible. Thanks for offering help with this!

chervand:
Please, review the updated PR.

Sephster:
Hi @chervand - sorry for my delayed reply. I think this is going in the right direction but I realised that we would have to set the scope policy directly on the response, where as I feel this policy should be set and controlled by the server instead. It makes sense to me that this is a server policy rather than a response type policy.

In the authenication server at the moment, we are setting up responses in the function `getResponseTypes()` by setting private keys and encryption keys on the bearer token.

I think this function would be the best place to also set the scope return policy based on the auth servers settings. That way, the server dictates this instead of the response itself.

One potential problem I can see with this is that we would need to call `setReturnScopes` from the auth server but we can pass in response types that implement the responseTypeInterface. To prevent the risk of calling `setReturnScopes` on a response type that doesn't have it, we might need to update the interface to ensure all response types implement this feature but this would be a BC breaking change and would likely need to be included in a future release.

If you can think of an alternative approach though, I'd very much welcome your thoughts.

chervand:
Hi @Sephster, sorry for the delay too. I'm still interested in this feature and I agree that this is an auth server policy rather than a response type policy.

> One potential problem I can see with this is that we would need to call setReturnScopes from the auth server but we can pass in response types that implement the responseTypeInterface. To prevent the risk of calling setReturnScopes on a response type that doesn't have it, we might need to update the interface to ensure all response types implement this feature but this would be a BC breaking change and would likely need to be included in a future release.

I can see 3 possible solutions now:
1. Update `ResponseTypeInterface` and leave it for the next possible release.
2. Add another interface declaring `setReturnScopes` which should extend `ResponseTypeInterface` and make default response type implementing this new interface instead of `ResponseTypeInterface`. Thus we could check whether response type instance implements it before calling `setReturnScopes`.
3. Add another interface which wouldn't extend `ResponseTypeInterface` and make default response type implementing this new interface as well as `ResponseTypeInterface`. With this implementation we could get rid of `setReturnScopes` and return scopes if response type implements such interface without breaking BC. However this implementation makes impossible to control it via the auth server.

Let me know what do you think please.

simonhamp:
@chervand @Sephster just been looking through comments on this issue/PR and I think extending `ResponseTypeInterface` feels like the better option at this stage.

Having said that, bundling into a new major release wouldn't be that bad - the fact is there are not thousands of people calling for this at the moment, so maybe we don't need to go to the trouble of extending and checking which interface is being implemented.

PR #801, opened 2017-10-25
Added ability to specify query delimiter character in the implicit grant

lucasantarella:
Ability to specify query delimiter, such as `?` instead of the hard-coded `#`

### DIFF
--- a/src/Grant/ImplicitGrant.php
+++ b/src/Grant/ImplicitGrant.php
@@ -31,10 +31,10 @@ class ImplicitGrant extends AbstractAuthorizeGrant
      */
     private $queryDelimiter;
 
-	/**
-	 * @param \DateInterval $accessTokenTTL
-	 * @param string $queryDelimiter
-	 */
+    /**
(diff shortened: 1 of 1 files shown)


Oxmoze:
Hi @lucasantarella 

Are you aware of the [OpenID Conenct specification](http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) and the `response_mode` parameter?

I think it would be better to support this authentication request parameter instead that solution as it may confuse developers and may not be supported by clients.

lucasantarella:
HI @Oxmoze 

Thanks for the quick response. I'm not sure how the `response_mode` parameter translates to modifying the way the auth responses are given back to the client. This is probably due to my ignorance on the subject 😄 . Can you show me how to utilize this functionality to achieve a response that gives the `access_token` as a query parameter in the callback uri?

alexbilbie:
Have merged this because the OAuth 2 spec does allow for an alternative to the fragment hash.

PR #803, opened 2017-10-28
When using ROPG with bad username/password, return 400, not 401

raarts:
When using a Resource Owner Password grant type, Oauth2 should respond with an HTTP 400 (Bad Request), not a 401 when a wrong username/password combination is supplied.

See[ this stackoverflow question](https://stackoverflow.com/questions/22586825/oauth-2-0-why-does-the-authorization-server-return-400-instead-of-401-when-the) for a good explanation.


raarts:
..aaaaand I forgot to run phpunit, and the tests still pass. Checking to see if I can add a test for this..

chervand:
Agree, this meets the RFC better.

src/Exception/OAuthServerException.php
```diff
@@ -120,7 +120,7 @@ public static function invalidScope($scope, $redirectUri = null)
      */
     public static function invalidCredentials()
     {
-        return new static('The user credentials were incorrect.', 6, 'invalid_credentials', 401);
+        return new static('The user credentials were incorrect.', 6, 'invalid_grant', 400);
```
Sephster: The 400 response can be left our here as the default HTTP code is set to 400 in the constructor.

If we are changing this to use the error code of _invalid grant_ instead of _invalid credentials_, as per the spec, there are a few scenarios that can trigger this error. I think we should perhaps look at making this error more generic to account for this. For example, the function name could be changed to `invalidGrant()` and the error message changed to use text describing this error in the OAuth spec in a similar manner to how the _invalidRequest_ exception has been implemented.

Sephster: Actually it looks like the exception is thrown but not caught so this might not be an issue for BC as it looks like there is a wider issue that needs to be addressed there

Sephster:
@raarts are you still wanting to push this PR through? Are you able to look at my initial comments and provide feedback? I would be keen to get this added to the version 8 branch if possible. Thanks

Sephster:
@raarts thank you for this PR. Because this is a breaking change, it probably should have been implemented on the 8.0.0 branch instead of the 7 branch. I tried to change this but it is proving to be difficult. In the interest of saving time, I'm going to close this PR and create a new one off the 8.0.0 branch with your changes included. I hope this is ok. Thank you so much for highlighting this to us!

PR #804, opened 2017-10-30
Accept rsa key with crlf

raarts:
I had a private/public key pair containing crlf endings. This wasn't recognized by the RSA_KEY_PATTERN in the CryptKey class. Here's a fix.


src/Exception/OAuthServerException.php
```diff
@@ -120,7 +120,7 @@ public static function invalidScope($scope, $redirectUri = null)
      */
     public static function invalidCredentials()
     {
-        return new static('The user credentials were incorrect.', 6, 'invalid_credentials', 401);
+        return new static('The user credentials were incorrect.', 6, 'invalid_grant', 400);
```
Spomky: ???

raarts:
Used wrong base branch closing and create a new PR.

PR #810, opened 2017-11-08
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

Just need to bump PHPUnit version to [4.8.36](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4836---2017-06-21) and [5.7.21](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#5721---2017-06-21) for compatibility with this `namespace`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "defuse/php-encryption": "^2.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8 || ^5.0",
+        "phpunit/phpunit": "^4.8.38 || ^5.7.21",
         "zendframework/zend-diactoros": "^1.0"
     },
     "repositories": [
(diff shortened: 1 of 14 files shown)


composer.json
```diff
@@ -13,7 +13,7 @@
         "defuse/php-encryption": "^2.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8 || ^5.0",
+        "phpunit/phpunit": "^4.8.38 || ^5.7.21",
```
Sephster: I think we are safe to drop support for PHPUnit 4x now. I will merge this in and do this change separately. Thanks for your contribution

Issue #677, opened 2016-10-29
Failure without 'scope' in BearerToken


everflux:
The 'scope' parameter is optional according to RFC 6749, but the `BearerTokenValidator` fails without a scope in the token due to

```
return $request
                ->withAttribute('oauth_access_token_id', $token->getClaim('jti'))
                ->withAttribute('oauth_client_id', $token->getClaim('aud'))
                ->withAttribute('oauth_user_id', $token->getClaim('sub'))
                ->withAttribute('oauth_scopes', $token->getClaim('scopes'));
```

I assume a possible fix would be to provide a default value

```
->withAttribute('oauth_scopes', $token->getClaim('scopes', [] ));
```

alexbilbie:
Thanks; will look into it

Sephster:
Thanks for this @everflux - I'm working on a fix at the moment to allow devs to set a default scope which I will add here shortly. If no scope is provided and no default scope is present, an exception will be thrown. Thanks for reporting this

PR #811, opened 2017-11-14
Add default scopes to authentication server

Sephster:
This change adds default scopes to the authentication server which can be used when no scope has been specified in a request.

In the event of no scope being provided for a refresh token, the system will revert to the original scope requested for the old token that is being refreshed. This fixes issue #677 by enforcing the specification of a scope.

If no default scope is set, an exception is thrown.

### DIFF
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -57,11 +57,6 @@ abstract class AbstractGrant implements GrantTypeInterface
     protected $scopeRepository;
 
     /**
-     * @var string
-     */
-    protected $defaultScope = '';
-
-    /**
      * @var AuthCodeRepositoryInterface
      */
     protected $authCodeRepository;
@@ -108,14 +103,6 @@ abstract class AbstractGrant implements GrantTypeInterface
     public function setScopeRepository(ScopeRepositoryInterface $scopeRepository)
     {
         $this->scopeRepository = $scopeRepository;
-    }
-
-    /**
-     * @param string $scope
-     */
-    public function setDefaultScope($scope)
-    {
-        $this->defaultScope = $scope;
     }
 
     /**
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -57,6 +57,11 @@ abstract class AbstractGrant implements GrantTypeInterface
     protected $scopeRepository;
 
     /**
+     * @var string
+     */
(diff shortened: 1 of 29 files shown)


Sephster:
Awesome. Thanks guys.  I will merge this in. As per @bretterer's suggestion, I will also get documentation updated but will submit this as a separate PR. Cheers guys

src/Grant/RefreshTokenGrant.php
```diff
@@ -44,28 +44,17 @@ public function respondToAccessTokenRequest(
...
-            // the request doesn't include any new scopes
-            foreach ($scopes as $scope) {
-                if (in_array($scope->getIdentifier(), $oldRefreshToken['scopes']) === false) {
-                    throw OAuthServerException::invalidScope($scope->getIdentifier());
-                }
+        $scopes = $this->validateScopes($this->getRequestParameter(
+            'scope',
+            $request,
+            implode(self::SCOPE_DELIMITER_STRING, $oldRefreshToken['scopes']))
+        );
```
sergeyklay: @Sephster how about `isset( $oldRefreshToken['scopes'])`?

src/Grant/AbstractGrant.php
```diff
@@ -211,29 +224,29 @@ protected function validateClient(ServerRequestInterface $request)
...
             if ($scope instanceof ScopeEntityInterface === false) {
                 throw OAuthServerException::invalidScope($scopeItem, $redirectUri);
             }
 
-            $scopes[] = $scope;
+            $validScopes[] = $scope;
+        }
+
+        if (empty($validScopes)) {
+            throw OAuthServerException::invalidScope($redirectUri);
```
sergeyklay: @Sephster
```diff
if (empty($validScopes)) {
--    throw OAuthServerException::invalidScope($redirectUri);
++    throw OAuthServerException::invalidScope($this->defaultScope ?: 'null', $redirectUri);
}
```

Sephster: Thanks for spotting this bug. I will commit a fix up shortly. Cheers

src/Grant/RefreshTokenGrant.php
Sephster: The changes around the default scope should mean that we don't issue a refresh token without a scope whether that be default or ones specified in the original request. I don't think this will be required unless there is a scenario I'm missing. Apologies if this is the case.

Issue #812, opened 2017-11-16
Encrypting/decrypting very slow due to use of de-/encryptWithPassword()


SunMar:
Hi,

During tests we discovered that encrypting and decrypting is very slow. This is caused by the use of the [PBKDF2 algorithm](https://github.com/defuse/php-encryption/blob/5176f5abb38d3ea8a6e3ac6cd3bbb54d8185a689/src/KeyOrPassword.php#L83) in [`KeyOrPassword`](https://github.com/defuse/php-encryption/blob/5176f5abb38d3ea8a6e3ac6cd3bbb54d8185a689/src/KeyOrPassword.php#L7) (from the [`defuse/php-encryption`](https://github.com/defuse/php-encryption) dependency), which can be traced back to [`CryptTrait`](https://github.com/thephpleague/oauth2-server/blob/925776958fc3f5278e74363663c20147af32b668/src/CryptTrait.php#L14) where the [`encryptWithPassword()`](https://github.com/thephpleague/oauth2-server/blob/925776958fc3f5278e74363663c20147af32b668/src/CryptTrait.php#L32) and [`decryptWithPassword()`](https://github.com/thephpleague/oauth2-server/blob/925776958fc3f5278e74363663c20147af32b668/src/CryptTrait.php#L49) methods are used. The use of this algorithm is secure if you're unsure about how good your encryption key is, however if you know you have a secure encryption key, then PBKDF2 is not needed.

[...]

PR #814, opened 2017-11-16
Allow CryptTrait to accept a \Defuse\Crypto\Key as encryption key #812

SunMar:
See #812 for details. This allows a Key object to be used as encryption key over an alphanumeric string which improves performance of encryption and decryption.


SunMar:
@Sephster Ah yes, sorry, forgot about those interfaces. Phpdoc updates.

SunMar:
I also updated the PR to remove some code duplication in the `CryptTraitTest`.

SunMar:
@simonhamp `setEncryptionKey()` is called by `AuthorizationServer` and is basically a pass-through for whatever encryption key is passed when doing a `new AuthorizationServer`, which is done by whoever is implementing this package and is not done by the package itself.

So I'm not really sure how to get rid of the `if`, because the library doesn't have control over whether the encryption is a `string` or a `Key`. This way it stays backwards compatible, if you pass a `string` it does the current implementation of `encryptWithPassword()` and only if you pass a `Key` it uses `encrypt()`.

Note that in the tests `doEncrypt()` and `doDecrypt()` are just pass-throughs in the stub to `encrypt()` and `decrypt()` because they are `protected` and can't be called directly. See [CryptTraitStub](https://github.com/thephpleague/oauth2-server/blob/8c93fd74c9efa2d81b20dc150bf278f6f62094c0/tests/Stubs/CryptTraitStub.php#L21).

SunMar:
@Sephster Is the PR ok for you too? I updated the change you requested.

Sephster:
Sorry for the delay in getting this reviewed @SunMar. I've made some minor changes but all looks good so merging in. Thanks for your patience and your contribution

PR #822, opened 2017-12-01
Test against PHP 7.2

carusogabriel:
With the release of [`PHP 7.2`](php.net/archive/2017.php#id2017-11-30-1), would be nice to test it.


Sephster:
Thanks. I'd been meaning to make this change. Cheers for being more on the ball than me :+1:

PR #823, opened 2017-12-05
Exchangeable token encrpytion

Kingson-de:
First of all thanks for providing this awesome package!
We want to use it in production in our company but we have the problem that we have no benefit to use JWT.
Our access tokens have their own storage so we don't need transport all the information with the token itself.
But the encryption was not exchangeable so I created a PR for it.
I don't know if you like the way or have a better idea how to make the token transportation more flexible. I hope we can discuss it and find a solution, because we really want to use your package :)


bauer01:
any news pls? its critical for us..

Sephster:
Hi @Kingson-de  - thanks for this PR. There is quite a lot of code to go through here so I was hoping you could assist by provinding a few more details about the PR to speed things along.

I was wondering if you could outline what alternative you are using to JWT's so I can understand the problem better? Ultimately it doesn't matter but it would be nice to have some contect here.

 Also which specific function are you having issues extending/overriding? I am presuming it is `convertToJwt()` which wouldn't be appropriate for a non-JWT bearer token. If you could expand on the specific issue you are trying to resolve and why you cannot do it given the current state of the library that would be much appreciated.

@bauer01 if you want to outline the issues you are having as well that would be much appreciated

lookyman:
I have an alternative here with a little less code 😄 https://github.com/thephpleague/oauth2-server/pull/874

Kingson-de:
Hey @Sephster,
the main problem right now is that you can't easily override the encryption of the tokens.
So my approach is that you only rely on the `AccessTokenEntityInterface` and `RefreshTokenEntityInterface`.
That way you can implement your own encryption of the tokens.
Currently if you have look on e.g. `BearerTokenResponse` there it is impossible to change the encryption.
Also if you compare the size of the Access/Refresh Tokens with Facebook or Google, they are more than double the size.
If you need more explanation, I am more than happy to help :)

Sephster:
To clarify, is it the JWT that you are wanting to replace or the encryption mechanism? I thought that initially you wanted to replace the JWT with your own bearer format but in your last post you mentioned that:

> That way you can implement your _own encryption_ of the tokens.

I am assuming you want to replace the JWT with some other format rather than alter the encryption mechanisms but just wanted to confirm this is the case.

Kingson-de:
@Sephster For us we want to implement our own bearer format. But with this implementation you can also adjust the information you put into the JWT. Maybe you need more or less data in the JWT, which can't be changed right now.
To be fair with the PR I provided you could not really replace the encryption mechanism, there would be a lot more changes involved.
That is why I proposed to discuss a solution which is beneficial for this package in my first comment :)

Kingson-de:
Hey @Sephster,
thanks for taking the time to review.
Sadly #874 is just renaming `convertToJWT` to `convertToAccessToken` in `AccessTokenEntityInterface` which is a BC nonetheless.
So if you go for a BC then I would take the more flexible solution.
Still the hardcoded generating of the refresh token in `BearerTokenResponse` is not fixed there.
Maybe I will take another approach when I got time.
But thanks anyway for responding and providing this OAuth solution for PHP!

lookyman:
@Kingson-de Not just renaming, mainly it changes the return type from JWT's token to a simple string, so users can define their own format.

Sephster:
@Kingson-de - if you can detail how you would change the refresh token, we can maybe look at putting a PR together. At the moment the refresh token is fairly basic, containing the following:
* client ID
* refresh token ID
* access token ID
* original scopes
* user ID
* expiry date

This is `json_encoded` and then encrypted for transport. I guess you could move some of the items (such as initial scopes into the DB but what you can in small refresh tokens, you lose in time querying the database so it would be hard to justify making such a change to the main codebase.

The easiest thing to do would probably be to just create your own response type that extends `AbstractResponseType` which is already within the scope of the current codebase.

Kingson-de:
@Sephster In my opinion it would be nice if the refresh/access token entity implements how it is converted for transport and which data it contains. Similar to the JsonSerializable interface.
For simple tokens this should be enough.
If the transport format is more complex like JWT, an exchangeable/optional serializer could do the final transformation.
So in the JWT case the token entity will produce an assoc array and the JWT serializer will encode the array to JSON and encrypt it afterwards.
We could have a token manager that is able to convert the token to the transport format using token and serializer.

In the special case of refresh tokens they actually could be short because when the client wants to exchange it for a new access token, the client needs to send it to the Auth Server which should validate the refresh token. So the Auth Server needs to have a storage for the refresh tokens to validate them. If you only rely on decrypting the JWT then you have a security issue because your refresh tokens can't be revoked.
And if your storage is for example Redis with twemproxy the querying is very fast.

Sephster:
Yep, this all sounds good to me. It would be nice to have a separate function where we can specifically target the way the refresh token is created to make  the library more flexible.

PR #824, opened 2017-12-06
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertInstanceOf` instead of `instanceof` operator;
- `assertInternalType` instead of `is_*` functions;
- `assertObjectHasAttribute` instead of `isset` function;
- `assertAttributeEquals` instead of `assertEquals`.


Sephster:
Thanks for your contribution!

PR #828, opened 2017-12-23
Fixed ordering so we only hash after base64 encoding

Sephster:
As per the examples in Apendix A of [RFC 7636](https://tools.ietf.org/html/rfc7636), we should be hashing the verifier code _after_ we have base 64 encoded, and not prior to. This pull request fixes an outstanding issue with our PKCE implementation.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -144,7 +144,7 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
                 case 'S256':
                     if (
                         hash_equals(
-                            strtr(rtrim(base64_encode(hash('sha256', $codeVerifier)), '='), '+/', '-_'),
+                            hash('sha256', strtr(rtrim(base64_encode($codeVerifier), '='), '+/', '-_')),
                             $authCodePayload->code_challenge
                         ) === false
                     ) {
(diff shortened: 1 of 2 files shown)


sgomez:
This is wrong, `base64_encode` must be done over the binary output of hash('sha256'). The process should be this:

`strtr(rtrim(base64_encode(hash('sha256', $codeVerifier, true)), '='), '+/', '-_')` 

This can be checked with the example described in [RFC7636 (Appendix B)](https://tools.ietf.org/html/rfc7636#appendix-B):

```
<?php
$codeVerifier = 'dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk';
$codeChallenge = 'E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM';

echo hash_equals(
  strtr(rtrim(base64_encode(hash('sha256', $codeVerifier, true)), '='), '+/', '-_'),
  $codeChallenge
) ? 'True' : 'False';
```

This returns `True`.

With the implementation of this PR:

```
<?php
$codeVerifier = 'dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk';
$codeChallenge = 'E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM';

echo hash_equals(
  hash('sha256', strtr(rtrim(base64_encode($codeVerifier), '='), '+/', '-_')),
  $codeChallenge
) ? 'True' : 'False';
```

This return `False`. 

hash('sha256', ..., false) returns a hexadecimal string and there are no way than match with _code challenge_ that it is a base64 string.

PR #839, opened 2018-01-13
Allow other secure key permissions.

simesy:
I'm working with a hosting provider that sets the key file to be 440, which is secure IMO.


Sephster:
Thanks @simesy - I've updated the error message to recommend 600/660 but see no issue with having read only be accepted

PR #840, opened 2018-01-13
BUGFIX: Wrong redirect uri.

liverbool:
This's bugfix when redirect on error.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -249,9 +249,9 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
 
         $scopes = $this->validateScopes(
             $this->getQueryStringParameter('scope', $request, $this->defaultScope),
-            is_array($client->getRedirectUri())
+            $redirectUri ?: (is_array($client->getRedirectUri())
                 ? $client->getRedirectUri()[0]
-                : $client->getRedirectUri()
+                : $client->getRedirectUri())
         );
 
         $stateParameter = $this->getQueryStringParameter('state', $request);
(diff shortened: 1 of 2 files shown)


Sephster:
Hi @liverbool - please can you provide some details on what bug this is fixing and how to reproduce the error as it is not clear to me from this PR. Thanks for your help

liverbool:
@Sephster, When redirect back to client site (on error) it not use `redirect_uri` from query string but from database config instead. This's wrong behavior if i have added multiple redirect Uris in my database config (client config).

src/Grant/AuthCodeGrant.php
```diff
@@ -249,9 +249,9 @@ public function validateAuthorizationRequest(ServerRequestInterface $request)
 
         $scopes = $this->validateScopes(
             $this->getQueryStringParameter('scope', $request, $this->defaultScope),
-            is_array($client->getRedirectUri())
+            $redirectUri ?: (is_array($client->getRedirectUri())
```
Sephster: The logic is becoming a bit unwieldy here. I think it would be easier to just use `$redirectUri` here. There is an if statement above which checks if `$redirectUri` is not null. We could set the variable to the client redirect Uri in a new else block to make the code cleaner.

src/Grant/ImplicitGrant.php
```diff
@@ -158,9 +158,9 @@ public function validateAuthorizationRequest(ServerRequestInterface $request)
 
         $scopes = $this->validateScopes(
             $this->getQueryStringParameter('scope', $request, $this->defaultScope),
-            is_array($client->getRedirectUri())
+            $redirectUri ?: (is_array($client->getRedirectUri())
```
Sephster: Same goes for this line here. We should just pass in `$redirectUri` and set the variable in an else block in the preceding if statement.

Sephster:
I reinstated the check on the client redirect uri to ensure we only use this is one redirect uri has been registered with the client as we don't want to always grab the first redirect uri if multiple have been registered against the client. In situations where multiple redirect uris have been listed, we should enforce that the specific redirect uri is passed along with the auth request

PR #842, opened 2018-01-18
Fix S256 code challenge method

sgomez:
This PR fix the problem described in #828 ([see comment](https://github.com/thephpleague/oauth2-server/pull/828#issuecomment-358408459)).

According to [RFC7636#section-4.3](https://tools.ietf.org/html/rfc7636#section-4.3):

    If the "code_challenge_method" from Section 4.3 was "S256", the
    received "code_verifier" is hashed by SHA-256, base64url-encoded, and
    then compared to the "code_challenge", i.e.:

    BASE64URL-ENCODE(SHA256(ASCII(code_verifier))) == code_challenge

So, the `hash('sha256',...)` must be done before the `base64_encode`.

The tests are modified to use example data from the [RFC7636#appendix-B](https://tools.ietf.org/html/rfc7636#appendix-B).

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -144,7 +144,7 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
                 case 'S256':
                     if (
                         hash_equals(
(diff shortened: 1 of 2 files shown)


Sephster:
Thanks for spotting this and providing a fix. Much appreciated

PR #844, opened 2018-01-29
Fix highlighting newlines.

frankdejonge:
The highlighting was messed up. By using this notation the newlines are preserved so the code samples show up correctly.

Before

screen shot 2018-01-29 at 10 31 39

After
screen shot 2018-01-29 at 10 31 22


frankdejonge:
Before

screen shot 2018-01-29 at 10 31 39

After
screen shot 2018-01-29 at 10 31 22

Issue #833, opened 2018-01-02
Writing key to file even if it exists


kpn13:
Hi,

In the saveKeyToFile method form the CryptKey class, the key is always wrote into the temporary key file even if it is unchanged.

https://github.com/thephpleague/oauth2-server/blob/57ca83a8baadd5b5a6c1d82da7d261089e55585c/src/CryptKey.php#L71-L86

Would it not be better if we skip the file_put_contents() if the file already exists for performances improvement ? Because it file exists before the "touch()" function call, it means the key is the same.

Am I wrong ?

Best regards.

alexbilbie:
Valid point 👍

Sephster:
Merged PR in as of commit 3999c41

PR #845, opened 2018-01-29
Do not create key file if it already exists and it is the same

kpn13:
See #833. Just a little improvement to skip the key file creation if it exists and it has no change.

Best regards.


Sephster:
This looks good to me. Thanks for your contribution @kpn13

PR #848, opened 2018-02-11
Static analysis with PHPStan

lookyman:
I intend to make the static analysis much more strict, but this is a good start..


alexbilbie:
Thanks!

PR #849, opened 2018-02-11
Update PHPUnit, run static analysis on tests

lookyman:


Sephster:
I've removed the temp vars and `@var` comments. I don't think these are needed as the variable types are already type hinted in the methods argument list. Other than that, looks great. Thanks for the submission. I will get this included in the release that is going out shortly.

lookyman:
They are needed when you use `composer udate --prefer-lowest`, but I can address that in a separate PR. Thanks!

Sephster:
Ah of course. Sorry, I didn't spot this. It probably is best to have these in then although long term it would be good to remove them. I will reinstate them and get this merged in. Thanks again

PR #852, opened 2018-02-17
Updated dependencies, more strict static analysis

lookyman:


.travis.yml
```diff
@@ -4,15 +4,19 @@ sudo: false
...
+  - DEPENDENCIES="--prefer-lowest --prefer-stable"
 
 php:
   - 7.0
   - 7.1
   - 7.2
 
 install:
-  - travis_retry composer install --no-interaction --prefer-source
+  - travis_retry composer update --no-interaction --prefer-dist $DEPENDENCIES
```
Sephster: I don't think we need the `travis_retry` call here. From the Travis docs, it looks like the retry function only works in the script step. If you could check/change this that would be much appreciated

lookyman: Sure, no problem.

lookyman: Done.

PR #856, opened 2018-02-18
PHPStan level 7

lookyman:
This fixes few issues with nullables, and adds custom extension for PHPStan so it can correctly resolve return types of some methods in `AbstractGrant`.

Level 7 is currently the most strict one, so we can't do any better than this. 🥇

### DIFF
--- a/src/Entities/AuthCodeEntityInterface.php
+++ b/src/Entities/AuthCodeEntityInterface.php
@@ -17,7 +17,7 @@ interface AuthCodeEntityInterface extends TokenInterface
     public function getRedirectUri();
 
     /**
-     * @param string|null $uri
+     * @param string $uri
      */
     public function setRedirectUri($uri);
 }
--- a/src/Entities/AuthCodeEntityInterface.php
+++ b/src/Entities/AuthCodeEntityInterface.php
@@ -12,12 +12,12 @@ namespace League\OAuth2\Server\Entities;
 interface AuthCodeEntityInterface extends TokenInterface
 {
     /**
-     * @return string
+     * @return string|null
      */
     public function getRedirectUri();
 
     /**
-     * @param string $uri
+     * @param string|null $uri
      */
     public function setRedirectUri($uri);
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
 
 script:
   - vendor/bin/phpunit --coverage-clover=coverage.clover
-  - vendor/bin/phpstan analyse -l 6 -c phpstan.neon src tests
+  - vendor/bin/phpstan analyse -l 7 -c phpstan.neon src tests
 
 after_script:
     - wget https://scrutinizer-ci.com/ocular.phar
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
-- Upgrade PHPStan checks to level 7 (PR #893)
+- Upgrade PHPStan checks to level 7 (PR #856)
 - Added event emitters for issued access and refresh tokens (PR #860)
 - Can now use Defuse\Crypto\Key for encryption/decryption of keys which is faster than the Cryto class (PR #812)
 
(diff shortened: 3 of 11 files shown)


src/Entities/AuthCodeEntityInterface.php
```diff
@@ -12,12 +12,12 @@
...
 {
     /**
-     * @return string
+     * @return string|null
      */
     public function getRedirectUri();
 
     /**
-     * @param string $uri
+     * @param string|null $uri
```
Sephster: hi @lookyman - thanks for your efforts to get PHP Stan up to level 7. It really is appreciated.

I'm concerned about some of the changes we've made to our interfaces here specifying that it is ok to pass null into methods like `setRedirectUri()`. I can't think of an instance where it would be useful to make a call like this so I would have thought we would only ever want to call this with a string.

The same goes for some of the other calls where we had added null either as a return type or a param call. I was wondering if you could provide some more information on the needs for this change? Thank you

lookyman: Well, PHPStan on level 7 (with the custom AbstractGrantExtension) reports these three errors:

```
 ------ --------------------------------------------------------------------
  Line   src\Grant\AuthCodeGrant.php
 ------ --------------------------------------------------------------------
  273    Parameter #1 $state of method
         League\OAuth2\Server\RequestTypes\AuthorizationRequest::setState()
         expects string, string|null given.
  323    Parameter #4 $redirectUri of method
         League\OAuth2\Server\Grant\AbstractGrant::issueAuthCode() expects
         string, string|null given.
 ------ --------------------------------------------------------------------

 ------ --------------------------------------------------------------------
  Line   src\Grant\ImplicitGrant.php
 ------ --------------------------------------------------------------------
  180    Parameter #1 $state of method
         League\OAuth2\Server\RequestTypes\AuthorizationRequest::setState()
         expects string, string|null given.
 ------ --------------------------------------------------------------------
```

The first one occures [here](https://github.com/lookyman/oauth2-server/blob/143afc9561ecb417d6560c1bd5331f5ee83d7f70/src/Grant/AuthCodeGrant.php#L273). As you can see, few lines above, we get the `$stateParameter` from a call to `AbstractGrant::getQueryStringParameter()` with a `null` default, which means it could technically result in a `null` value, because there is no validation that the server request really contains the `state` parameter.

The second one occures [here](https://github.com/lookyman/oauth2-server/blob/143afc9561ecb417d6560c1bd5331f5ee83d7f70/src/Grant/AuthCodeGrant.php#L327). The `AuthorizationRequest::getRedirectUri()`'s return type is `string|null`, and again, we cannot be sure the `null` value won't be returned.

The third one occures [here](https://github.com/lookyman/oauth2-server/blob/143afc9561ecb417d6560c1bd5331f5ee83d7f70/src/Grant/ImplicitGrant.php#L180), and is identical to the first one.

I didn't dig too deep in the code to figure out the necessary validations tht would have to be implemented. The easiest fix was to allow nullability in those places, which kept the current behavior intact (since it's all only in phpdocs), and postpone the issue to a later time when it could be dealt with properly.

Is this sufficient info? I would be happy to improve this PR according to any suggestions you might have.

Sephster: Thanks for your detailed reply @lookyman. Apologies for the delay in getting back to you. I think there probably are some issues here which haven't surfaced until we've implemented PHPStan.

We should probably try and address these issues as I feel changing our PHPDoc to match the current behaviour, but potentially confuse end users is not something we should do.

Looking at the `state` for example, this should only every consist of printable ascii characters. Setting it to null won't have any impact in this instance but it still seems somewhat misleading to me to call `setState` if we don't need to. As you suggested, I think this will need to be tidied up with some validation. I haven't looked at the other two instances yet but I presume they will be of a similar state.

If we accept this PR in its current form, it will ultimately mask the underlying issues which need to be addressed. Running PHPStan at level 7 has highlighted these issues so I think it would be appropriate to deal with them in this issue.

If I get time soon, I will look at rectifying these problems and putting in correct validation but feel free to address these if you also have a spare minute. As soon as these have been addressed I think this will be good to merge in.

Thanks again for your efforts with this!

I appreciate there is no behaviour change happening here but our PHPDoc

Sephster:
Merged in. Thanks for the efforts here @lookyman

Issue #641, opened 2016-08-18
Validate token


nrwiersma:
Hi,

Was looking for a way to manually validate a token. It seems to only be possible to through a bearer token validator. There are a fair number use cases for this, the main one in my case is allowing consumers to validate a token and its scopes.

As a reference, GitHub does this, the documentation can be found [here](https://developer.github.com/v3/oauth_authorizations/#check-an-authorization).

alexbilbie:
You'll need to implement this manually for now, but V6 will support "token
introspection" which is an RFC that I've been working on.

nrwiersma:
Awesome, great to hear. Thanks.

simonhamp:
@alexbilbie did Introspection make it's way into v6?

Issue #579, opened 2016-05-13
Apart OAuth2 server - Validate on other API's


bobmulder:
Hey all,

For a project I am planning to build 3 API's, and all of them are going to use OAuth2.
I've read a lot about it, and talked a lot about it and I am starting to understand it....

I decided to build one extra API, which is the OAuth2 API. I guess I should use this awesome library for that.

But, what about securing my 3 other API's? I would love to use the middleware (https://oauth2.thephpleague.com/resource-server/securing-your-api/), but I can't figure out how the middleware should validate it via the OAuth2 API. Any hints? 

Thank you very much!

Greetz,

Bob

alexbilbie:
Hi Bob,

So your auth server (i.e. the implementation of this library) will issue access tokens (specifically JWTs signed using the auth servers private key).

The API clients then use the access tokens as authorisation when calling your other APIs.

Each of those APIs implement the resource server middleware that you linked to in your question. When the access token in the request is parsed by the middleware a few things will happen:

1. The middleware will check the access token is signed by the Auth server's public key
2. The middleware will then check the access token hasn't expired
3. Finally the `isAccessTokenRevoked` method on [`AccessTokenRepositoryInterface`](https://oauth2.thephpleague.com/access-token-repository-interface/) (that was passed into the `ResourceServer` class when you initialised it) will be called to verify the access token hasn't been revoked since it was issued.

If the access token is valid the following attributes will be set on the HTTP request that was passed into the middleware:

* `oauth_access_token_id` - the access token identifier
* `oauth_client_id` - the client identifier
* `oauth_user_id` - the user identifier represented by the access token
* `oauth_scopes` - an array of string scope identifiers

If the authorization is invalid an instance of `OAuthServerException::accessDenied` will be thrown.

Does that answer your question?

bobmulder:
Hi Alex,

Thank you for taking time to answer my strugge ;)

> So your auth server (i.e. the implementation of this library) will issue access tokens (specifically JWTs signed using the auth servers private key).

> The API clients then use the access tokens as authorisation when calling your other APIs.

Completely agree, I already got that...

> Each of those APIs implement the resource server middleware that you linked to in your question. When the access token in the request is parsed by the middleware a few things will happen:

> - The middleware will check the access token is signed by the Auth server's public key
- The middleware will then check the access token hasn't expired
[...]

PR #869, opened 2018-03-02
Add introspection implementation

fetzi:
according to RFC 7662 the introspection mechanism is implemented.


simonhamp:
Thanks for your work on this @fetzi! I am not yet up to speed with RFC7662, so will need a bit more time to digest that thoroughly before I can make a full review of this PR.

But on first pass, this looks like a good start. I have already spotted a couple of things that I think would need to be addressed to make this solution more complete and ready for release though:

As you're primarily reading the data stored in the token, we're missing the opportunity to do some more interesting introspection directly from the repository, which would give us the ability to include `username` (or other more opaque in the response, for example.

Also, it's limited just to access tokens at the moment, but the RFC allows for introspection on all token types as well as allowing the client to hint at the token type (via the `token_type_hint` parameter) it is trying to inspect.

We should also think about documentation, as this is a useful feature that I'm sure many implementors will want to deploy and it requires some extra considerations.

I'm also marking this for Discussion because I think that there's a really interesting twist on this that's not really explored in the RFC: why not use the Token being inspected itself as the Bearer token instead of it being in the body of the request? The authorization server will then be able to tell whether or not the token is valid using already-implemented methods.

This also solves a potential issue I see with this RFC. The ability/desire to allow for _en masse_ introspection of many tokens is a bit of a weak link, which the authors sort of acknowledge. If the credentials of any client that has the ability to do this introspection is compromised, it would be open season on finding active tokens, potentially even across many clients if not implemented securely - a nasty single point of failure. I think forcing clients to authenticate with the token they're attempting to inspect goes some way to deflecting this.

fetzi:
Thanks for your feedback @simonhamp.

I tried to avoid breaking changes in the introspection implementation (for a . But yes additional data like the username would be great to be included. What else are you thinking of?

The usage of the `token_type_hint` parameter is a good idea, I will try to implement this functionality as soon as possible.

According to your security concerns: the RFC says that you should protect your introspection endpoint in some way except if its only an internal resource. Therefore I would leave the implementation of a security mechanism (oAuth2 Access Token, HTTP Basic Auth, ...) up to the implementor. This is a very important and sensible point that should be highlighted in the introspection documentation as well.

The usage of the acutal token to introspect is not allowed because the RFC states in section 2.1 that the introspection request is a HTTP POST request that sends the input parameters as `application/x-www-form-urlencoded` data. Therefore moving the token into the Header is not a good idea IMHO.

And yes when the feature is completed I will send an additional PR for a detailed documentation update.

What do you think about the points mentioned? Maybe we should split the points up and define them in detail what should be included in the auth server and how?

simonhamp:
I would hold fire until we've had chance to discuss this a bit further.

> Therefore I would leave the implementation of a security mechanism (oAuth2 Access Token, HTTP Basic Auth, ...) up to the implementor.

I agree. This just needs to be well covered in documentation - we'd like to help people avoid building naïve implementations that are open to abuse, if possible.

> The usage of the acutal token to introspect is not allowed because the RFC states in section 2.1 that the introspection request is a HTTP POST request that sends the input parameters as application/x-www-form-urlencoded data. Therefore moving the token into the Header is not a good idea IMHO.

I fully appreciate this point, but the fact is this is not a ratified standard as yet, 'only' a "Proposed Standard" RFC (like most of the OAuth2 RFCs) and so it's an issue that is still open for discussion among all of us who will be using/implementing against this standard in the coming years, which is especially likely if it is bundled into packages such as this.

My point here is for us to engage critical thinking, not simply to implement any and every related RFC that pops up verbatim. Standards should tend to follow implementations and accepted best practice, not the other way around. So we are kind of in the driving seat on stuff like this.

This is just my opinion though, weakly held. Happy to see it implemented in whatever way suits and works and ultimately is the safest :)

I'll pick this up with the other maintainers (in case they haven't already seen this) and hopefully we'll get some discussion going on these points.

simonhamp:
Closing some loops:

https://github.com/laravel/passport/issues/491
https://github.com/thephpleague/oauth2-server/issues/579
https://github.com/thephpleague/oauth2-server/issues/641

fetzi:
@simonhamp any update on this PR?

fetzi:
Closing this for now, implementation done by @steveporter92 in #925.

Issue #870, opened 2018-03-06
Authorization Code flow: conflict between an authorization code custom redirect_uri (saved in the auth code grant entity) and what's saved in the client entity


jackdpeterson:
A little background: I'm implementing a light version of Open ID Connect to support two providers (utilizing the League-Oauth2-Client provider adapters).

I have an action, FacebookAction that 
1. Receives a GET request from a given client (http://sso/login/facebook/{clientId})
2. Redirects to FB Authorization Dialog
3. A. convert the FB code to FB access token, fetches FB identity.
    B. Auto-creates a new user or merges an existing user with what's already in the DB by the e-mail address. 
4. Create a new AuthorizationRequest() with the appropriate parameters, mainly differing the redirect_uri. Finally, pass that to the authServer::completeAuthorizationRequest(). This redirects the user back to the Client's configured redirect URI (a microservice's login app).

A current assumption in the existing AbstractGrant code is that the client.redirect_uri and authorization_code.redirect_uri are always identical. To me, this seems to make that field (in the authCode entity) redundant ... For the sake of security verificaiton I believe all that ought to be compared against would be the HTTP_HOST portion, port, and protocol (http / https). The other details ought to be able to be overridden ... otherwise we should always refer to the client.redirect_uri when validating things.

Effectively, when one wants to override the redirect_uri in the authorizationCode entity there is a conflict between AbstractGrant::validateClient()
```
// If a redirect URI is provided ensure it matches what is pre-registered
        $redirectUri = $this->getRequestParameter('redirect_uri', $request, null);
        if ($redirectUri !== null) {
            if (
                is_string($client->getRedirectUri())
                && (strcmp($client->getRedirectUri(), $redirectUri) !== 0)
            ) {
                $this->getEmitter()->emit(new RequestEvent(RequestEvent::CLIENT_AUTHENTICATION_FAILED, $request));
                throw OAuthServerException::invalidClient();
            } elseif (
                is_array($client->getRedirectUri())
                && in_array($redirectUri, $client->getRedirectUri()) === false
            ) {
                $this->getEmitter()->emit(new RequestEvent(RequestEvent::CLIENT_AUTHENTICATION_FAILED, $request));
                throw OAuthServerException::invalidClient();
            }
        }
```
and 

AuthCodeGrant::respondToAccessTokenRequest
```
// The redirect URI is required in this request
            $redirectUri = $this->getRequestParameter('redirect_uri', $request, null);
[...]

PR #871, opened 2018-03-06
comment ensure redirect_uri is present

jackdpeterson:
#870 -- The assumption that this change would make is that the redirect_uri supplied by a client redeeming an authorization_code for an access token is that they supply just what's configured in the client. No verification happens inside the grant for the redirect_uri ... rather the stored redirect_uri is used purely for redirecting a user to a client with whatever parameters need to be in place ... and the verification happens at a code level.

Of course, the other assumption is that the way the authorization_code redirect_uri is generated is secure server-side and doesn't allow manipulation of the host portion or otherwise.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -70,6 +70,12 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
         ResponseTypeInterface $responseType,
         \DateInterval $accessTokenTTL
     ) {
+        
+        // The redirect URI is required in this request.
+            $redirectUri = $this->getRequestParameter('redirect_uri', $request, null);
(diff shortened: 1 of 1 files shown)


Sephster:
Hi @jackdpeterson. Thank you for your PR. It isn't clear to me what issue this is addressing. The grant first grabs the redirect_uri from the request and then compares it against the redirect uri stored in the authorisation code.

This is in line with the following section of RF6749:

> ensure that the "redirect_uri" parameter is present if the "redirect_uri" parameter was included in the initial authorization request as described in Section 4.1.1, and if included ensure that their values are identical.

The change proposed in this PR would break this mechanism so I don't believe it is something we should be merging in. For this reason, I'm going to close this PR but if you do have further comments please feel free to respond and I will look at reopening this if necessary.

PR #873, opened 2018-03-13
Response converts access token format

lookyman:
This allows for moving away from JWT or simply adding new token formats.

### DIFF
--- a/src/AuthorizationServer.php
+++ b/src/AuthorizationServer.php
@@ -163,7 +163,7 @@ class AuthorizationServer implements EmitterAwareInterface
     public function completeAuthorizationRequest(AuthorizationRequest $authRequest, ResponseInterface $response)
     {
         return $this->enabledGrantTypes[$authRequest->getGrantTypeId()]
-            ->completeAuthorizationRequest($authRequest)
+            ->completeAuthorizationRequest($authRequest, $this->getResponseType())
             ->generateHttpResponse($response);
     }
 
--- a/src/Entities/AccessTokenEntityInterface.php
+++ b/src/Entities/AccessTokenEntityInterface.php
@@ -9,17 +9,6 @@
 
 namespace League\OAuth2\Server\Entities;
 
-use Lcobucci\JWT\Token;
-use League\OAuth2\Server\CryptKey;
-
 interface AccessTokenEntityInterface extends TokenInterface
 {
-    /**
-     * Generate a JWT from the access token
-     *
-     * @param CryptKey $privateKey
-     *
-     * @return Token
-     */
-    public function convertToJWT(CryptKey $privateKey);
 }
--- a/src/Entities/Traits/AccessTokenTrait.php
+++ b/src/Entities/Traits/AccessTokenTrait.php
@@ -9,37 +9,11 @@
 
 namespace League\OAuth2\Server\Entities\Traits;
 
-use Lcobucci\JWT\Builder;
-use Lcobucci\JWT\Signer\Key;
-use Lcobucci\JWT\Signer\Rsa\Sha256;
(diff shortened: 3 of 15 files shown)


simonhamp:
@lookyman this looks very interesting, thanks for your time and effort in preparing and sharing this with us. 🙏 

Feels like there's quite a lot changing here, so it may take some time to review. Please bear with us.

lookyman:
@simonhamp No problem at all.

There really aren't that many changes, the idea is to move the `convertToJWT` method from `AccessToken` to `ResponseType`. Everything else is basically fixing tests. 😄

Sephster:
Thanks for this @lookyman. I've had a look through this and had a few thoughts. If this was to be accepted it would need to be a major release as it is changing so many interfaces. I expect version 8 would be some way off at the moment as we only released version 7 recently.

I think it is great that we make the library as flexible as possible so any PRs to support this are appreciated. I've had a quick look through this and was wondering if you could outline your reasoning for moving the conversion from the access token to the responseType.

My initial thoughts are that we could just make the accessToken implement a more generic conversion function. It seems appropriate to me that the conversion is sitting within the access token because in the case of the JWT conversion, we rely on many of the internal functions within the access token itself to make the conversion e.g. `getScopes`, `getClient` etc. 

Moving the conversion to the response type means we need to pass the access token and call the access tokens functions which just seems a little messier to me although there might be good reason for this.

If you could outline what issues there are with keeping the conversion in the access token for you, so I can better understand the motivations for this change, that would be much appreciated. Thanks for your time

lookyman:
Your idea of AccessToken implementing a more generic conversion function would work too, you are right. I missed that solution. However, that would mean that the interface would have to change anyway, so we are still breaking BC.

My reasoning for moving the conversion to ResponseType was a case of separation of concerns. The token should IMO be basically a value object holding some data, but a format in which that data appear in the response should be a concern of the response, not the token. I admit that this might be more of an academic reason than a practical one.

In any case, the JWT aspect is currently hardcoded in, and I'd like to change that one way or the other. What do you think?

Sephster:
Okies I see the logic there. Thanks for clarifying. I don't see the function as that different from a `_toString` function or someting in Java like where you can convert a string to an int or float with inbuilt functions. The object itself both stores the information and has the ability to output itself in a different format. 

Given that, and the fact that we are relying on many of the functions already within the access token object to make the conversion, I think it is probably best we still keep the conversion function within the access token itself.

The main problem I see is the naming of the function itself. If, as you've noted, you wanted to use a different mechanism for transporting the access token, the name `convertToJWT` doesn't lend itself well for this purpose as it is too specific.

I think the best approach with this would be to just rename the function from `convertToJWT` to something more generic like `getResponseString` or something similar. I'm sure there are probably better names but that is the first one off the top of my head.

This would still be a breaking change so would have to wait until version eight but would let this classes interfaces be overridden more easily, with less change in the codebase overall.

WRT implementing Pasteto, I think this is something that would be desirable for implementers but I'm unsure we'd bring this into the core of the package unless there were clear benefits such as reduction in size of the payload which some people have complained about with the current JWT setup. I haven't looked into this much though and this might  be something that Pasteto offers but we must ensure we are catering for the desires of most users without increasing the size of the package unnecessarily. 

By making the access token more generic, we can then leave it up to the implementer if they want to move away from JWTs.

lookyman:
Thank you for your comments and ideas, @Sephster. I will close this PR and prepare another one.

(As for the Paseto thing, I was thinking of a separate package on top of this one. But we can maybe implement it at some point in the future, possibly after @sarciszewski submits a proper RFC to IETF - I recall him mentioning somewhere that he wants to eventually do that. In the meantime, we can stick with JWT.)

PR #876, opened 2018-03-17
Fix fatal error in examples caused by ClientRepositoryInterface change

steverhoades:
The following fatal is thrown when attempting to execute the examples.

```
Fatal error: Declaration of OAuth2ServerExamples\Repositories\ClientRepository::getClientEntity($clientIdentifier, $grantType, $clientSecret = NULL, $mustValidateSecret = true) must be compatible with League\OAuth2\Server\Repositories\ClientRepositoryInterface::getClientEntity($clientIdentifier, $grantType = NULL, $clientSecret = NULL, $mustValidateSecret = true)
```

### DIFF
--- a/examples/src/Repositories/ClientRepository.php
+++ b/examples/src/Repositories/ClientRepository.php
@@ -17,7 +17,7 @@ class ClientRepository implements ClientRepositoryInterface
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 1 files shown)


Sephster:
Thanks for the fix @steverhoades

Issue #886, opened 2018-04-10
Adding support for Client scopes


leroy0211:
It would be nice to also limit Clients to specific scopes. I'm not sure if it's part of the Oauth2 Spec though. 

For example you might have 3 scopes `read`, `write` and `delete`. 

In most cases, every client may request the `read` scope. But not all clients may request the `write` or `delete` scopes. Especially when using the client credentials grant.

simonhamp:
@leroy0211 This is left up to your implementation. The logic for achieving this shouldn't be too complicated, but it is very dependent on your specific circumstances as it won't be needed/wanted in all implementations.

Having said that, if you do come up with an implementation that you feel is quite generic and can be integrated without adding any extra setup on behalf of other implementors, if you can share it, please feel free to submit a PR :)

Sephster:
The finalizeScope method is probably best used for this purpose. It will modify the requested scopes by the client so would be perfect for this. As @simonhamp  says, this is an implementation detail so would be unlikely to be included in the core library

PR #887, opened 2018-04-10
Made the Client be aware of scopes

leroy0211:
but only when the Client implements the ScopeInterface and the ScopesTrait. So it's like optional. 

Moved Scope related code to it's own interface and trait, otherwise it would end up with code duplication.

This relates to #886

### DIFF
--- a/src/Entities/ScopeInterface.php
+++ b/src/Entities/ScopeInterface.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * @author      Alex Bilbie <hello@alexbilbie.com>
+ * @copyright   Copyright (c) Alex Bilbie
+ * @license     http://mit-license.org/
+ *
+ * @link        https://github.com/thephpleague/oauth2-server
+ */
+
+namespace League\OAuth2\Server\Entities;
+
+interface ScopeInterface
+{
+    /**
+     * Associate a scope with the entity.
+     *
+     * @param ScopeEntityInterface $scope
+     */
+    public function addScope(ScopeEntityInterface $scope);
+
+    /**
+     * Return an array of scopes associated with the entity.
+     *
+     * @return ScopeEntityInterface[]
+     */
+    public function getScopes();
+}
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -268,6 +268,7 @@ abstract class AbstractGrant implements GrantTypeInterface
         foreach ($entity->getScopes() as $scope) {
             $scopesList[] = $scope->getIdentifier();
         }
+
         return $scopesList;
     }
 
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -222,8 +222,9 @@ abstract class AbstractGrant implements GrantTypeInterface
      * @param string                $scopes
      * @param string                $redirectUri
      *
-     * @return ScopeEntityInterface[]
      * @throws OAuthServerException
+     *
+     * @return ScopeEntityInterface[]
      */
     public function validateScopes(ClientEntityInterface $client, $scopes, $redirectUri = null)
     {
@@ -257,6 +258,7 @@ abstract class AbstractGrant implements GrantTypeInterface
      * Retrieve the scope identifiers from an Entity.
      *
      * @param ScopeInterface $entity
+     *
      * @return string[]
      */
     private function getScopeIdentifiers(ScopeInterface $entity)
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -234,7 +234,8 @@ abstract class AbstractGrant implements GrantTypeInterface
         $validScopes = [];
 
         if ($client instanceof ScopeInterface) {
-            if ($clientScopes = $this->getScopeIdentifiers($client)) {
+            $clientScopes = $this->getScopeIdentifiers($client);
+            if (count($clientScopes) > 0) {
                 $scopesList = array_intersect($clientScopes, $scopesList);
             }
         }
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -18,6 +18,7 @@ use League\OAuth2\Server\Entities\AuthCodeEntityInterface;
 use League\OAuth2\Server\Entities\ClientEntityInterface;
 use League\OAuth2\Server\Entities\RefreshTokenEntityInterface;
 use League\OAuth2\Server\Entities\ScopeEntityInterface;
+use League\OAuth2\Server\Entities\ScopeInterface;
 use League\OAuth2\Server\Exception\OAuthServerException;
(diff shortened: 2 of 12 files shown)


src/Entities/ScopeInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ * @author      Alex Bilbie <hello@alexbilbie.com>
+ * @copyright   Copyright (c) Alex Bilbie
+ * @license     http://mit-license.org/
+ *
+ * @link        https://github.com/thephpleague/oauth2-server
+ */
+
+namespace League\OAuth2\Server\Entities;
+
+interface ScopeInterface
```
Sephster: I think the naming of this interface is confusing. A scope shouldn't be able to add itself. I thought that perhaps you were trying to associate scopes with clients and that this would be a _client_ interface but I see it is being used by the token interface. It makes sense that a scope can be added to a token I guess but we already had this functionality in the token interface, which has been deleted and moved here instead.

src/Grant/AbstractGrant.php
```diff
@@ -217,21 +218,29 @@ protected function validateClient(ServerRequestInterface $request)
...
-    public function validateScopes($scopes, $redirectUri = null)
+    public function validateScopes(ClientEntityInterface $client, $scopes, $redirectUri = null)
     {
         $scopesList = array_filter(explode(self::SCOPE_DELIMITER_STRING, trim($scopes)), function ($scope) {
             return !empty($scope);
         });
 
         $validScopes = [];
 
+        if ($client instanceof ScopeInterface) {
```
Sephster: I would not expect a client to be an instance of a scope interface.

tests/Stubs/ClientScopeEntity.php
```diff
@@ -0,0 +1,24 @@
...
+use League\OAuth2\Server\Entities\ScopeInterface;
+use League\OAuth2\Server\Entities\Traits\ClientTrait;
+use League\OAuth2\Server\Entities\Traits\EntityTrait;
+use League\OAuth2\Server\Entities\Traits\ScopeEntityTrait;
+
+class ClientScopeEntity implements ClientEntityInterface, ScopeInterface
+{
+    use EntityTrait, ClientTrait, ScopeEntityTrait;
+
+    public function setRedirectUri($uri)
```
Sephster: I would expect this to be a class for scopes associated with clients given the naming. The function suggests this is actually for a client instead as we wouldn't set a redirect URI on a scope.

Sephster:
Hi @leroy0211 - thank you for your submission. I don't think this PR will be merged in as I think there are some design decisions which make the library more complicated. Some things that stood out for me were:

`if ($client instanceof ScopeInterface)` 

This see

leroy0211:
Sad to hear, but based on your previous suggestion I also created a PR in Passport, because it's also achievable with `finalizeScopes()`. They however also declined the PR, so I'm forced to use some other library for Oauth2 in our Microservice Architecture. 

But I'm still a bit confused about the use of scopes in `thephpleague/oauth2-server` (and Passport). Scopes are there to protect access to resources. But there is nothing that restricts a client to request access to all available scopes, and therefor could have access to all your API's. So why would anyone implement scopes if every client can request access to all of them?
 
Based on the Oauth2 spec, there is actually nothing that suggests a Client must first be assigned to one or more scopes. So I totally understand why thephpleague won't add it to the core.

bitgandtter:
Scopes are intended for human users, so they can allow/disallow access to api on their behalf, not on behalf of a client AFAIK

Sephster:
Yep, @bitgandtter is correct. The scopes requested by the client are displayed to the user and it is up to the user to accept or decline. If you, as the owner of the auth server, want to restrict certain clients to specific scopes, you can do this using finaizeScopes but as this is something that would not be required by most users, it is you as the implementer who should implement this.

When a client registers with an OAuth server, they only need to provide information on whether they are public or confidential and their redirect URIs. There is nothing in the specification that stipulates a client should also provide the scopes they intend to use.

The use case you are describing is not really described in the spec as as @bitgandtter explains, it is the user and not the server which will typically dictate what scopes the client has access to. I'm sorry that your use case isn't catered to out of the box in this library but I would be surprised if you find any OAuth implementation that does this as it is not a scenario outlined in the OAuth spec.

leroy0211:
But what you describe only applies on the `Auth code` grant (with user interaction), and maybe on the `password` grant. 
So what's implied is, that in case of the `client credentials` grant none of the scopes matter, because it's without interaction of the user and without any user-id. But there is no way to access any user-resource with the `client credentials` grant since there is no user. So in what kind of use-case would you use the `client credentials` grant?


Maybe I am misusing Oauth2 (and scopes) to limit complete access to specific API's, and maybe I need to implement another layer of security, something like ACL's for clients.

simonhamp:
@leroy0211 first of all, it's worth noting that there's no silver bullet here... no one solution will safely guard all of the different areas of your application appropriately - only you are in the best position to determine what's needed, when and for whom.

Having simple pieces that function in predictable ways with clear APIs is much easier to get a grasp of. It's also simpler to find pieces that do/don not fit together when you are fully conversant with their interfaces and interactions. OAuth 2 Server needs to stay flexible in this regard.

Finally, in your specific case, it sounds like something which I've also worked on in Passport which limits certain scopes to specific grant types. I've achieved this with some config and simple middleware:

```php
<?php

namespace App\Http\Middleware;

use Closure;
use Laravel\Passport\Passport;

class EnableAppropriateScopes
{
    /**
     * Dynamically enable scopes based on the grant type.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  \Closure  $next
     * @return mixed
     */
    public function handle($request, Closure $next)
    {
        $enabledScopes = config('passport.scopes.global', []);

        // Special case for the Authorization Code Grant flow, which requires scopes in Step 1
        if ($request->request->get('response_type') == 'code') {
            $enabledScopes = array_merge($enabledScopes, config('passport.scopes.authorization_code', []));
        } elseif ($grantType = $request->request->get('grant_type')) {
            $grantSpecificScopes = config("passport.scopes.{$grantType}", []);
            $enabledScopes = array_merge($enabledScopes, $grantSpecificScopes);
        }

        Passport::tokensCan($enabledScopes);

        return $next($request);
    }
}
```

Register and use this middleware and create the appropriate config (or hardcode your settings in the middleware itself). You will also probably need to remove any calls to `Passport::tokensCan()` in your `AuthServiceProvider` or whichever provider you've been putting your Passport bindings in.

Hope this helps!

leroy0211:
@simonhamp Thanks, I didn't thought of altering the `tokensCan` in a middleware. My use-case however does not work completely like that.

I currently kind of fixed my use-case with a new Layer of ACL (on the client) just like the `$user->can([role], [object])` of the laravel `Auth`. 
Instead I created a middleware which does something like `$client->can('product:write')`, and limited the client's access like that. All configuration is stored in another database table, so it can be maintained by other teams.

But thanks everybody for this discussion, I really learned a bit more about Oauth2.

PR #897, opened 2018-04-26
Add method to return token data and modify validateAuthorization() to…

marekk:
This is alternative to https://github.com/thephpleague/oauth2-server/pull/851

### DIFF
--- a/src/AuthorizationValidators/AuthorizationValidatorInterface.php
+++ b/src/AuthorizationValidators/AuthorizationValidatorInterface.php
@@ -14,6 +14,14 @@ use Psr\Http\Message\ServerRequestInterface;
 interface AuthorizationValidatorInterface
 {
     /**
+     * Determine the access token in the authorization header and return Token object with configured claims
+     *
+     * @param ServerRequestInterface $request
+     * @return \Lcobucci\JWT\Token
+     */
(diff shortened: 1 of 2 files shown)


simonhamp:
Thanks for taking the time to submit this PR @marekk, but I can't see the value that this adds over #851.

In fact, I think this is actually a slightly more dangerous implementation as it's even easier to override core validation functionality (esp. without some code duplication) and moves the really critical stuff further away from where it's being called.

It also causes some potential confusion, as the naming of the methods is very similar while their purpose remains distinct. This feels like a side-effect of trying to work within the bounds of the existing  contract whilst still splitting the code to make it more easily overridable. Your `validateAuthorizationHeader` is clearly named (if a bit more verbose than necessary), but your amended `validateAuthorization` is actually just adding claims to the token, not really validating anything.

marekk:
My problem is that I need to remove that validation because we don't use the registered claim names here, and they are all optional.

`validateAuthorization`  is not amended but the original kept for backward compatibility.

simonhamp:
Understood. From what you've said it seems like this is very specific to your use-case.

I believe that you would be able to implement an easier fix for your needs if/when #851 does settle on improving how validation and claim extraction is done.

Issue #779, opened 2017-08-08
Public/Private Key Permissions on Windows


johannesschobel:
Dear @alexbilbie ,

as you have requested in #778 , I can give some more details about my current development machine.
* I am currently running Windows 8.1 (latest patches installed).
* As a "development environment" I use XAMPP ( https://www.apachefriends.org/de/index.html ) for Windows 7.1.7 (with PHP 7.1.7)
* I work with `laravel 5.4`
* I use `laravel/passport 3.*`

and then I ran into this error
> "Key file \"file:\/\/C:\\Path\\To\\My\\Larvel\\Project\\storage\\oauth-private.key\" permissions are not correct, should be 600 or 660 instead of 666",

The problem is, that Windows Users do not have that access/right management that you know from Linux. So we (Windows users) are not able to `chmod 660` any specific file. If you try to get the permission when running a PHP script, it tells you that the permission is 666 to somehow simulate this behaviour.

Therefore, the mentioned PR / commit 2aca909d203e8a925da8c3a3f16a803683cecf04 causes heavy problems on windows machines..

Any ideas on this?!
Cheers

vvasilev-:
Hey @johannesschobel,

According to [these lines](https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationServer.php#L95-L97), you can pass a `CryptKey` instance. So you can do the following: 

```php
<?php 

use League\OAuth2\Server\CryptKey;
use League\OAuth2\Server\AuthorizationServer;

// rest of your code...

$privateKey = new CryptKey($keyPath, null, false);
$server = new AuthorizationServer(
    $clientRepository,
    $accessTokenRepository,
    $scopeRepository,
    $privateKey,
    $encryptiongKey
);
```

I tried this on my Windows machine and everything works well.

vvasilev-:
Regarding to `laravel/passport`, they're already doing the same thing [here](https://github.com/laravel/passport/blob/3.0/src/PassportServiceProvider.php#L231-L235). Just make sure that you're using `v3.0.1` of `laravel/passport`.

johannesschobel:
Yes, i am using `laravel/passport` so the problem is fixed for me.. However, other users, that are not using `laravel/passport`, struggle with this issue..

Sephster:
I don't believe there is an issue here as I don't think the library is forcing this permission check to occur. If a developer does not want the permissions to be checked, they can create a CryptKey instance with the third argument set to false as stated above. 

[...]

PR #901, opened 2018-05-17
Update CryptKey.php

ShyZhen:
当我在windows环境下测试的时候，它报错，虽然我知道他不应该在windows下跑。
这是我的原始问题：https://github.com/laravel/passport/issues/712
我需要一些建议，谢谢。

-----
When I tested in the windows environment, it was wrong, though I knew he should not run under windows.
This is my original question: https://github.com/laravel/passport/issues/712
I need some advice, thank you.

### DIFF
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -48,7 +48,7 @@ class CryptKey
         if ($keyPermissionsCheck === true) {
             // Verify the permissions of the key
             $keyPathPerms = decoct(fileperms($keyPath) & 0777);
-            if (in_array($keyPathPerms, ['400', '440', '600', '660'], true) === false) {
+            if (in_array($keyPathPerms, ['400', '440', '600', '660'], true) === false && strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN') {
                 trigger_error(sprintf(
                     'Key file "%s" permissions are not correct, recommend changing to 600 or 660 instead of %s',
                     $keyPath,


Sephster:
I don't think `fileperms()` is reliable on Windows and is only really viable on Unix systems. This is one of the reasons we only issue a notice. I must admit though, I haven't been able to find much documentation on this and don't have a Windows machine to hand at the moment to test this.

If it is the case that Windows servers can't be relied upon for `fileperms()` as I suspect, we might need to see if we can have a work around or exemption.

This was discussed in [another issue](https://github.com/thephpleague/oauth2-server/issues/779). It was noted there that if you use an instance of  `CryptKey`, you shouldn't need to worry about this. 

This information might be of use  to you @ShyZhen so I would recommend looking through the previous issue raised.

I will try and have a look at this properly tonight but we might need to reopen this issue as multiple people have reported problems with Windows servers and this check.

If anyone has a reliable source for how `fileperms()` works on Windows it would be appreciated if you could share.

ShyZhen:
Hi, 谢谢您的答复，我认为用Windows进行权限管理其实是很难的一件事，我一般都使用Windows进行开发，最终程序跑在Linux；我一直在尝试，怎样把我的文件权限改成600，但是，我试过了很多方法，找了很多资料，她依然是666，在我眼里Windows并不适合做权限管理。不过这只是我的一面之词，因为我也很无知，我也在追求真理。谢谢阅读。

------
Hi, thanks for your reply, I think it's very difficult to use Windows to manage permission, I usually use Windows for development, and the final program runs in Linux; I've been trying to change my file rights to 600, but I tried a lot of methods and found a lot of information, she was still 666, In my opinion
 Windows is not suitable for permission management. It's statement of only one of the parties, because I am also ignorant, and I am also seeking truth. Thank you for reading.

Sephster:
Reopening this issue. I've not had chance to test this yet but suspect that @ShyZhen is right, and we need to re-evaluate this for Windows users

marcharding:
Hey, i just stumbled upon this very problem.

If oauth2-server is not used directly but as a dependency, it is not going to work on windows (if used directly one can just the the $keyPermissionsCheck flag, so that is not a problem).

I dug a little bit deeper and there is no way to get a windows php to return 600, so it would be great if this could be merged.

Why will windows never return 600?

Take a look here:

https://github.com/php/php-src/blob/master/ext/standard/php_filestat.h#L23

The  S_**USR, S_**GRP and S_**OTH all get the same value S_IREAD, S_IWRITE and S_IEXEC which is what you get on windows. So you will always get 666, 444 etc. but never 600.

marcharding:
Any news here?

Sephster:
No update on this yet. I am leaning towards removing these checks but am torn. I need to take a look at the history of this being introduced. We will either implement this PR or remove the checks entirely. Thank you for the offer of a new PR but at this time, it is not required. Cheers

PR #904, opened 2018-05-22
Fix uncaught exception produced by unsigned token

dzibma:
Fixes uncaught BadMethodCallException caused by unsigned access token validation.


Sephster:
Hi @dzibma. Which method is causing the exception? Is it the `verify()` method?

Sephster:
Thanks for finding this @dzibma

PR #913, opened 2018-06-16
Reword the error message for invalid clients

brentgaither:
The error message for invalid clients is vague, adding some more information to check on an invalid client


brentgaither:
Thanks for the feedback. I thought that might be giving away too much information, but I was not sure and figured I would give the suggestion.

Sephster:
No problem. Thank you for the suggestion and sorry we didn't decide to go forwards with implementing this this time.

Issue #914, opened 2018-06-18
Probable BC break between 4.1.3 and 4.1.6


rmasclef:
Hello :)

We just updated some old php 5.4 (#shame) app deps and we encountered an error upgrading v4.1.3 to v4.1.6 ^^

Here is some log that might help:

`request.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\FatalErrorException: "Compile Error: Can't use method return value in write context" at [...]/vendor/league/oauth2-server/src/ResourceServer.php line 141 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalErrorException(code: 0): Compile Error: Can't use method return value in write context at [..]/vendor/league/oauth2-server/src/ResourceServer.php:141)"} []`

Downgrading from v4.1.6 to v4.1.3 "fixes" the problem for now.

FYI: this error seams to appear only on php5.4 (it works great on php5.5 and 5.6).

I'd be curious to know what the problem is here :+1: 

Thanks for your help !

Sephster:
It looks like the issue is here:

https://github.com/thephpleague/oauth2-server/blob/3bec591393b1cc6a3fc02c94d18262e4ffad2e4f/src/ResourceServer.php#L141

Prior to PHP 5.5, you could only put variables inside the `empty()` function so the function call is causing the error. I will raise a fix for this as version 4 of the package should support PHP 5.4.

I would strongly encourage you to upgrade to a later version of the package though. Thanks for raising this.

Sephster:
@rmasclef I believe the issue is resolved now. I've release version 4.1.7 of the package with the bug fix. Please let me know how you get on. Thanks for reporting this.

rmasclef:
@sephter thanks a lot for your reactivity! i'll check by the end the week :)

Sephster:
Excellent. Please do let me know how you get on. Thanks

On 24 June 2018 22:16:53 BST, Romain Masclef <notifications@github.com> wrote:
>@sephter thanks a lot for your reactivity! i'll check by the end the
>week :)
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/thephpleague/oauth2-server/issues/914#issuecomment-399788289

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Sephster:
@rmasclef did the fix work as expected?

rmasclef:
[...]

PR #916, opened 2018-06-21
Adding validation to Refresh Token Grant ensuring old access token is valid

melanef:
Adding validation to Refresh Token Grant ensuring refresh token has a valid access token

When Access Token Storage is unable to find the Access Token referred by the Refresh Token (by it being deleted), we should throw an Invalid Refresh Token Exception. If we don't throw such exception, $oldAccessToken has null as value and then $oldAccessToken->getSession() fails with an "Call to a member function getSession() on null" error.

### DIFF
--- a/src/Grant/RefreshTokenGrant.php
+++ b/src/Grant/RefreshTokenGrant.php
@@ -161,6 +161,11 @@ class RefreshTokenGrant extends AbstractGrant
 
         $oldAccessToken = $oldRefreshToken->getAccessToken();
 
+        // Ensure the old refresh token has a valid access token referred by it
+        if (($oldAccessToken instanceof AccessTokenEntity) === false) {
+            throw new Exception\InvalidRefreshException();
+        }
+
         // Get the scopes for the original session
         $session = $oldAccessToken->getSession();
         $scopes = $this->formatScopes($session->getScopes());


Sephster:
HI @melanef. Thank you for your PR. I see the value in this but I also believe this could be solved by the library implementer through a simple change to their deletion policy. 

Why would someone delete their access tokens if they know they are associated with refresh tokens? It isn't an impossible scenario but I was wondering if you could provide some background as to your own circumstance so I can gain a better understanding.

Is was also wondering if this related to issue #914 that you posted? I notice this is targeting branch 4 at the moment which is not under active development any more but that isn't too much of a problem if this is indeed a bug or warrants an update to that branch. I hope with a bit more background we will be able to decide the best place for this code together. Many thanks and I look forward to hearing from you.

melanef:
Hi Andrew, thanks for taking the time to analyze and reply to my PR.

No, I'm not the one who opened issue #914, and it's totally not what I have been experiencing. 

I agree with you that it can (maybe should) be fixed on my side (and I believe it will as soon as my partners review it) as it is somewhat of an inconsistency, but at the same time I also think the oAuth server should fail gracefully throwing an exception instead of issuing a "Call to member function on null" error.

My issue was caught on Sentry (a tool to monitor and track errors and events) and I did some analysis on the whole flow to understand where would be the problem. I even considered adding the exception throw on my Access Token Storage method get, where we check that the access token we are looking for does not exist, but it wouldn't be the best approach.

On my platform we do have an area where our users can check which API clients are authorized and, if he wishes, delete the authorization, where is the only part of the system where an access token can be removed. The only way to reproduce this error is to get a new access token (by the authorization flow) and then removing the access token.

My pull request was issued against branch 4.1.x because it's the one we use (4.1.6) on our platform and to be honest, I don't know if the problem exists on the other branches.

Once again, thank you very much for your attention.

Sephster:
Thanks for the reply @melanef. Apologies for the mixup with the issue. I misread the username. Looking at later versions of the package, we have a function called `validateOldRefreshToken()` which handles the problem you are trying to address.

Are you unable to upgrade to a later version of the package? What version of PHP are you running? 

Thanks!

Sephster:
@melanef I've had a think about this and I don't think this warrants being added to the version 4 branch. It isn't actively under development at the moment. I released a bug fix for it today but that was because the bug in question breaks the entire code base if hit and the only way to resolve it is to modify the code.

In this instance, you could change your token policy to prevent this from happening. In addition, this issue has been resolved in later versions of the library. I would _strongly_ recommend anyone using this version of the branch upgrades to a version that at a minimum, is compliant with PHP 7. 

Thank you for raising this and I hope you understand my decision. All the best.

PR #917, opened 2018-06-21
Allow 640 as key file permisions

Erikvv:
Don't have any special reason for this, infra at my company set the permissions like this for everything.


Sephster:
Thanks @Erikvv - this looks grand.

PR #920, opened 2018-06-26
PHPStan 0.10

lookyman:


Sephster:
Thanks @lookyman. There are _a lot_ of changes here so this will take some time to process. I also note that this is currently failing our build as PHPStan 0.10 doesn't support PHP 7.0. 

I don't think this change would currently warrant a version upgrade in itself so is probably better targetted for the 8.0 branch which won't be released for some time yet.

lookyman:
Oh right, I forgot that we still support PHP 7.0. 😢

Sephster:
Aye, it isn't ideal to be honest. I would like to move off 7.0 pretty soon. Thinking on it though, it might not require a major bump in this package as fundamentally, nothing has changing in our API and composer should handle anyone who isn't running 7.1 or above.

I will have  a wee think but will probably just include this in the next version on the 7 branch. Will need some time to do some testing first though. Thanks for the submission!

lookyman:
Also the weird thing is that I don't get those reported errors locally. I will work on it some more as soon as I have some more time.

PR #925, opened 2018-07-20
Add introspection implementation

StevePorter92:
Off the back of the pull request comments laid out here https://github.com/thephpleague/oauth2-server/pull/869. I thought I would take the comments on board, as I am keen to start using the introspection features already started by @fetzi.


Sephster:
Thanks for this @StevePorter92 and apologies for not picking this up yet. I plan on setting some time aside this weekend to pick up some of the pending PRs. Thanks for your patience and thank you for your contribution

StevePorter92:
Hey @Sephster, don't suppose you have had a chance to cast your eye of this?

Sephster:
Sorry @StevePorter92 I have not yet. I have been working on numerous things for the library over the past week or so. I'm focussing on getting the PKCE changes into version 8 at the moment as there was at one time 3 separate pull requests trying to address some implemenation concerns around that. Once I've finished this (which is now fairly close to be honest), I will look at getting this reviewed. Thanks for your patience

src/AuthorizationServer.php
```diff
@@ -197,6 +209,73 @@ public function respondToAccessTokenRequest(ServerRequestInterface $request, Res
...
+
+    /**
+     * Get the introspection response
+     *
+     * @return IntrospectionResponse
+     */
+    protected function getIntrospectionResponseType()
+    {
+        if ($this->introspectionResponseType instanceof IntrospectionResponse === false) {
+            $this->introspectionResponseType = new IntrospectionResponse;
```
Sephster: We should add brackets to the class e.g. `new IntrospectionResponse()`

src/AuthorizationServer.php
```diff
@@ -197,6 +209,73 @@ public function respondToAccessTokenRequest(ServerRequestInterface $request, Res
...
+        $introspectionResponse = $introspector->respondToIntrospectionRequest(
+            $request,
+            $this->getIntrospectionResponseType()
+        );
+
+        return $introspectionResponse->generateHttpResponse($response);
+    }
+
+    /**
+     * Return an introspection response.
```
Sephster: The description here doesn't match the function. Can this be updated? Thank you

src/AuthorizationServer.php
```diff
@@ -197,6 +209,73 @@ public function respondToAccessTokenRequest(ServerRequestInterface $request, Res
...
+     * @param ServerRequestInterface $request
+     */
+    public function validateIntrospectionRequest(ServerRequestInterface $request)
+    {
+        $introspector = $this->getIntrospector();
+        $introspector->validateIntrospectionRequest($request);
+    }
+
+    /**
+     * Returns the introspector
```
Sephster: Please add a full stop at the end of the description here. Apologies as I thought this would have been picked up by Travis CI. I will need to look at the rules

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,102 @@
...
+     * Set the token against the response
+     *
+     * @param Token $token
+     */
+    public function setToken(Token $token)
+    {
+        $this->token = $token;
+    }
+
+    private function hasToken()
```
Sephster: Need a docblock for this function

Sephster:
Hi @StevePorter92. Thanks for your patience here. A couple of things spring to mind at the moment which I think are missing from this at the moment:

* The token doesn't _have_ to be a JWT. Although this server provides JWT bearer tokens out the box, implementers are free to use other formats. I don't think this is handled at the moment so will need to be addressed.
* I don't believe the introspection end point is protected. We should be protecting this with some form of authentication as it should not be available to public clients. Note that some implementations will put this end point behind a firewall internally so we have to take this into consideration as well.
* I don't see how the resource server is using this end point. Can you clarify this.

Thanks for your efforts to date. I hope these comments are useful. Cheers!

StevePorter92:
Hey @Sephster! Really appreciate the feedback, will get to your comments as soon as possible

Sephster:
Hi @StevePorter92 - any updates on this PR? I'm happy to pick this up in your absence if you're snowed under.

StevePorter92:
Hey @Sephster. Sorry for the delay. Little snowed under recently yes. I've got a few updates I'll check-in hopefully this evening to address your comments. Thanks for the support!

StevePorter92:
Added some changes to address your comments @Sephster.

The Introspector now validates through a IntrospectionValidator interface. I've moved all of the token logic to a BearerTokenValidator class. This should give implementers the freedom to validate introspection requests for other formats.

I'm a little unsure as to the best way to get the resource server to use our new endpoint so could do with being pointed in the right direction. My current line of thought is to create a `AuthorizationValidators\IntrospectionValidator` class. I think this will just need the URL of the introspection endpoint as well as any information required for authentication?

Also a little unsure on the best way to protect this new endpoint. What are your thoughts on this?

Thanks again.

Sephster:
It is awaiting review from myself. I will get to it eventually but don't have a firm timeline for this at present sorry.

StevePorter92:
As @sephster said, there isn't currently a timeline for this feature @dvdbot.

Introspection can be achieved with laravel through passport and https://github.com/designmynight/laravel-oauth-introspect-middleware.

Currently what's outstanding is finding a sensible way for the resource server to use introspection. We also need to consider authentication. 

@sephster, when you come to review this, let me know if there's anything I can pick up.

Sephster:
Will do, thanks @StevePorter92. I should have some free time this evening. Unfortunately @dvdbot, this isn't just a case of reviewing the code. I also need to familiarise myself with the RFC to ensure it is conforming to spec which is why PRs such as this take longer than a normal bug fix. I won't be drawn on timescales as I find they are nearly always optimistic. Hope you understand and sorry I can't give a more definitive answer than that

src/AuthorizationServer.php
```diff
@@ -197,6 +216,103 @@ public function respondToAccessTokenRequest(ServerRequestInterface $request, Res
...
+    /**
+     * Get the introspection response
+     *
+     * @return IntrospectionValidatorInterface
+     */
+    protected function getIntrospectionValidator()
+    {
+        if ($this->introspectionValidator instanceof IntrospectionValidatorInterface === false) {
+            $this->introspectionValidator = new BearerTokenValidator($this->accessTokenRepository);
+        }
```
ralfbecker: The BearerTokenValidator has no validation-key set, unless you also call here
```
$this->introspectionValidator->setPrivateKey($this->privateKey);
```
But that also requires that the public-key is also added to the private-key file, as Lcobucci\JWT\Signer\Rsa::doVerify($expected, $payload, Key $key) calls open_ssl_publickey($key->getContent())!

With that change I was able to verify/introspect access_keys generated by my server.

I'm currently implementing an [OpenID Connect / OAuth2 server for EGroupware](https://github.com/EGroupware/openid)

canfone:
I'm just wondering why you put the introspection logic in the authorization server? Shouldn't it be in the resource server? Then create a new validator `IntrospectionValidator` maybe then apply same logic as the `BearerTokenValidator` but except it doesn't require to be have `Authorization` header?

ArondeParon:
Is this feature still being actively considered for development?

Sephster:
Yes. Rough plan is to get Device Code in place and then this. January is super busy for me as I need to do tax returns etc but will be picking up speed on this soon. Cheers

kgrosvenor:
This would be very nice, be cool to be able to authenticate other server projects as if it was apart of the same, for e.g to be able to validate scopes against another API

Veitor:
How is the progress?

ThaDaVos:
Sorry to comment, but any ETA?

mikados:
Sorry to comment. How is the progress?

abbluiz:
How can I help to make this happen? I think this feature would be very useful.

bram-pkg:
I need this to create separate auth and resource servers in a project at work. Any updates? How can I help?

Sephster:
I'm working on the 9.0 release right now so have added this to the milestone but I don't expect this to be out for at least a month, maybe more sorry

saade:
Will this ever be implemented?

hafezdivandari:
#1473

PR #932, opened 2018-08-19
Added docbloc to UniqueTokenIdentifierConstraintViolationException

JasonTheAdams:
PhpStorm was throwing a warning since the UniqueTokenIdentifierConstraintViolationException::create method doesn't explicitly return an instance of itself and doesn't document it for the function.

I just added the docblock to make things happy. Not a big deal if it isn't merged, but it made my life easier. 👍


Sephster:
Thanks @JasonTheAdams

PR #933, opened 2018-08-21
Update RefreshTokenGrant.php

jeremyj11:
function validateOldRefreshToken(ServerRequestInterface $request, $clientId) requires $clientId as a string, but $refreshTokenData['client_id'] is an integer.
See https://stackoverflow.com/questions/51907450/laravel-passport-invalid-refresh-token-token-is-not-linked-to-client/51952225

### DIFF
--- a/src/Grant/RefreshTokenGrant.php
+++ b/src/Grant/RefreshTokenGrant.php
@@ -99,7 +99,7 @@ class RefreshTokenGrant extends AbstractGrant
         }
 
         $refreshTokenData = json_decode($refreshToken, true);
-        if ($refreshTokenData['client_id'] !== $clientId) {
+        if ($refreshTokenData['client_id'] != $clientId) {
             $this->getEmitter()->emit(new RequestEvent(RequestEvent::REFRESH_TOKEN_CLIENT_FAILED, $request));
             throw OAuthServerException::invalidRefreshToken('Token is not linked to client');
         }


Sephster:
Hi @jeremyj11 - Can you provide some more information for this? The client ID should be set from your Client Entity function _getIdentifier()_ which should return a string.

Please see https://github.com/thephpleague/oauth2-server/blob/27b956c1499f7fcfbad3a7c071cb4f48e4271e9d/src/ResponseTypes/BearerTokenResponse.php#L35-L47

I'm interested to know if your implementation does return a string at this point or not. If you could share your implmentation of the BearerTokenResponse that would be great thanks.

JasonTheAdams:
I was confused by this at first, thinking there was an integer id (such as a typical auto-incrementing primary key) as well as the long identifier. I realized, though, that there really only needs to be the identifier. So in my case it's always a string and it works fine.

Sephster:
@jeremyj11 are you still wanting to merge this? If you could respond to my previous comment so we can get this progressed, that would be much appreciated. Thanks for your help

Sephster:
Hi @jeremyj11 - I'm closing this PR as we have not heard back from you. If you'd like to have this PR progressed, please respond to my initial question and I will be happy to reopen this and take a look. Thank you.

trevorgehman:
@Sephster Just an FYI, I also ran into this issue because older versions of [laravel/passport](https://github.com/laravel/passport) did not explicitly cast the identifier as a string. You can see when they made the change: https://github.com/laravel/passport/pull/550/files.

So my problem is I have a lot of `client_id` values that are integers. So basically that means in order to upgrade, I'll have to invalidate all the existing logins, which I'd like to avoid doing. I'm looking for workarounds.

Is there any downside to doing a loose comparison of the client_id like this PR suggests?

PR #941, opened 2018-09-13
Introducing backwards compatibility in CryptTrait for decrypting password-encrypted data when a \Defuse\Crypto\Key is used

EricTendian:
At the moment, in order for someone using this package to adopt the significant performance improvement made in https://github.com/thephpleague/oauth2-server/pull/814, they must have all their refresh tokens invalidated, as old refresh tokens encrypted with a password encryption key cannot be decrypted with the Key object, even if the underlying data is the same, due to how the key is created.

In order to make switching from the old encryption method to the new one non-disruptive, this change will first try to decrypt data with a Key object if one is specified, otherwise it will pull the raw bytes out of the encryption key and act as if it the raw key bytes are a password.

In practice, this enables https://github.com/laravel/passport/pull/820 to be done in a non-disruptive fashion, allowing the library to gain the performance benefit.

### DIFF
--- a/src/CryptTrait.php
+++ b/src/CryptTrait.php
@@ -56,7 +56,15 @@ trait CryptTrait
     {
         try {
             if ($this->encryptionKey instanceof Key) {
-                return Crypto::decrypt($encryptedData, $this->encryptionKey);
+                try {
+                    return Crypto::decrypt($encryptedData, $this->encryptionKey);
+                } catch (\Exception $e) {
+                    // This may have been encrypted with encryptWithPassword(), pull
+                    // the data out of the key and use that to decrypt.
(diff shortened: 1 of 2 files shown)


Sephster:
Thanks for this pull request @EricTendian. Unfortunately I think this might introduce a security concern. If you use a string password, your chosen password might be weak. To combat this, the encryption library applies key stretching to your password to make it less susceptible to brute force attacks.

By contrast, the key object is known to be secure by default (as you aren't setting the password used) so it can skip the password stretching making it a lot quicker to use.

The only way this backwards compatibility check can work is if we set the password on the Key object to one we had previously chosen, which mitigates the benefits of using the Key object.

Thank you for your efforts with this but I believe there is no BC solution to this problem.

PR #959, opened 2018-10-31
Fix missing return typehints

mabar:
getClientEntity
https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AbstractGrant.php#L184-L191

getScopeEntityByIdentifier
https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AbstractGrant.php#L252-L254

getUserEntityByUserCredentials
https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/PasswordGrant.php#L93-L99

### DIFF
--- a/src/Repositories/ClientRepositoryInterface.php
+++ b/src/Repositories/ClientRepositoryInterface.php
@@ -25,7 +25,7 @@ interface ClientRepositoryInterface extends RepositoryInterface
      * @param bool        $mustValidateSecret If true the client must attempt to validate the secret if the client
      *                                        is confidential
      *
-     * @return ClientEntityInterface
+     * @return ClientEntityInterface|null
      */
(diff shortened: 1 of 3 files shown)


Sephster:
Thanks for your pull request @mabar. I don't believe these functions you've edited will ever explicitly return null so I don't think this is the correct change to make. We won't merge in your PR this time but thank you for your submission.

mabar:
It is not required to return null, I agree. Currently it is 'anything else than entity'. But I expect that you will in future add native typehints, so it will be only entity or null.
If you dont like Entity|null so it should be at least Entity|mixed as current phpdoc tells repository must return entity which is definitely incorrect.

Sephster:
I think it would be Entity or void rather than null. We don't explicitly state void in our doc blocks and I don't think this would change in the future.

PR #960, opened 2018-11-02
Make AuthorizationServer stateless

marc-mabe:
I want to use this library by `zend-expressive-authentication-oauth2` and together with `zend-expressive-swoole`.

The problem I have now is that the `AuthenticationServer` is state full by (in my opinion bad deisgn):
1. The injected object `$responseType` gets modified by the server by setting the keys
2. The same object gets modified again by the authentication grants to inject tokens

What happens by me together with swoole is that an access and refresh token get injected on the first successful password authentication and after that the same swoole worker serves a successful client credentials authentication and the response contains the refresh token on the first password authentication. 👎 

To quickly solve that without introducing BC break I have changed the injected `responseType` to act as a prototype. Means it gets cloned on first injection to not modify the injected objected and secondly clone again on authentication response.

In general I thing this specific handling should be refactored by a better architecture but that for sure introduces a BC break.

### DIFF
--- a/src/AuthorizationServer.php
+++ b/src/AuthorizationServer.php
@@ -49,9 +49,9 @@ class AuthorizationServer implements EmitterAwareInterface
     protected $publicKey;
 
     /**
-     * @var null|ResponseTypeInterface
+     * @var ResponseTypeInterface
      */
-    protected $responseType;
+    protected $responseTypePrototype;
 
     /**
      * @var ClientRepositoryInterface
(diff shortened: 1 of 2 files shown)


Sephster:
Thanks @marc-mabe for spotting this and providing a fix. Tis very much appreciated. If you want a stab at the BC breaking refactor you had hinted at, we have a version 8 branch that I plan on releasing towards the end of the year. Thanks again for your PR.

ceeram:
So it was not without breaking BC: https://github.com/thephpleague/oauth2-server/issues/968

sunspikes:
Would it also be possible to make a release for 6.x series with this fix because the `zendframework/zend-expressive-authentication-oauth2` uses that.

PR #963, opened 2018-11-08
BC-Break: cleanup DateTime handline for 8.0.0

marc-mabe:
* `DateTime` -> `DateTimeImmutable`
* `DateTime::format('U')` -> `DateTime::getTimestamp()`
* `(new DateTime())->getTimestamp()` -> `time()`
* added return type-hints for `DateTimeImmutable`

Open Question:
I'm not sure if the current timezone is is a problem but something like that `(new DateTimeImmutable())->add($accessTokenTTL)` depends on the current default timezone where an interval of e.g. `P1D` results into a different time then `PT24H` if there is a Daylight-saving-time switch in between.

### DIFF
--- a/src/Entities/RefreshTokenEntityInterface.php
+++ b/src/Entities/RefreshTokenEntityInterface.php
@@ -32,7 +32,7 @@ interface RefreshTokenEntityInterface
      *
      * @return DateTimeImmutable
      */
-    public function getExpiryDateTime(): DateTimeImmutable;
+    public function getExpiryDateTime();
 
     /**
      * Set the date time when the token expires.
--- a/src/Entities/RefreshTokenEntityInterface.php
+++ b/src/Entities/RefreshTokenEntityInterface.php
@@ -9,6 +9,8 @@
 
 namespace League\OAuth2\Server\Entities;
 
+use DateTimeImmutable;
+
 interface RefreshTokenEntityInterface
 {
     /**
@@ -28,16 +30,16 @@ interface RefreshTokenEntityInterface
     /**
      * Get the token's expiry date time.
      *
-     * @return \DateTime
+     * @return DateTimeImmutable
      */
-    public function getExpiryDateTime();
+    public function getExpiryDateTime(): DateTimeImmutable;
 
     /**
      * Set the date time when the token expires.
      *
-     * @param \DateTime $dateTime
+     * @param DateTimeImmutable $dateTime
      */
-    public function setExpiryDateTime(\DateTime $dateTime);
+    public function setExpiryDateTime(DateTimeImmutable $dateTime);
 
     /**
      * Set the access token that the refresh token was associated with.
--- a/src/Entities/TokenInterface.php
+++ b/src/Entities/TokenInterface.php
@@ -32,7 +32,7 @@ interface TokenInterface
      *
      * @return DateTimeImmutable
      */
-    public function getExpiryDateTime(): DateTimeImmutable;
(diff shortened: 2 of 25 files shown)


src/Entities/RefreshTokenEntityInterface.php
```diff
@@ -28,16 +30,16 @@ public function setIdentifier($identifier);
     /**
      * Get the token's expiry date time.
      *
-     * @return \DateTime
+     * @return DateTimeImmutable
      */
-    public function getExpiryDateTime();
+    public function getExpiryDateTime(): DateTimeImmutable;
```
Sephster: Revoking this change as we aren't using return types at this time.

src/Entities/TokenInterface.php
```diff
@@ -28,16 +30,16 @@ public function setIdentifier($identifier);
     /**
      * Get the token's expiry date time.
      *
-     * @return \DateTime
+     * @return DateTimeImmutable
      */
-    public function getExpiryDateTime();
+    public function getExpiryDateTime(): DateTimeImmutable;
```
Sephster: Revoking this change as we aren't using return types at this time.

src/Entities/Traits/AccessTokenTrait.php
```diff
@@ -67,9 +68,9 @@ public function __toString()
     abstract public function getClient();
 
     /**
-     * @return \DateTime
+     * @return DateTimeImmutable
      */
-    abstract public function getExpiryDateTime();
+    abstract public function getExpiryDateTime(): DateTimeImmutable;
```
Sephster: Revoking as not using return types at this time.

src/Entities/Traits/RefreshTokenTrait.php
```diff
@@ -42,19 +43,19 @@ public function getAccessToken()
     /**
      * Get the token's expiry date time.
      *
-     * @return \DateTime
+     * @return DateTimeImmutable
      */
-    public function getExpiryDateTime()
+    public function getExpiryDateTime(): DateTimeImmutable
```
Sephster: Revoking as not using return types at this time.

src/Entities/Traits/TokenEntityTrait.php
```diff
@@ -57,19 +58,19 @@ public function getScopes()
     /**
      * Get the token's expiry date time.
      *
-     * @return \DateTime
+     * @return DateTimeImmutable
      */
-    public function getExpiryDateTime()
+    public function getExpiryDateTime(): DateTimeImmutable
```
Sephster: Revoking as not using return types at this time.

src/Entities/RefreshTokenEntityInterface.php
Sephster: Revoked

src/Entities/TokenInterface.php
Sephster: Revoked

src/Entities/Traits/AccessTokenTrait.php
Sephster: Revoked

src/Entities/Traits/RefreshTokenTrait.php
Sephster: Revoked

src/Entities/Traits/TokenEntityTrait.php
Sephster: Revoked

PR #965, opened 2018-11-12
Include previous exception in catch and throw

ceeram:


Sephster:
Thanks @ceeram - this looks good to me. I've updated the changelog. I also wanted to add in some basic tests and where possible, remove the backslashes from `\Exception` and `\LogicException` where applicable but I need to head out.

I will probably pick this up tomorrow and make the changes. If you have time to make them in the meantime, that would be much appreciated but no worries if not. Thanks so much for submitting this!

ceeram:
You want use import statements add throughout the whole project? i can do that, although unrelated to this PR. I can make a new PR for that if you like. I can also add some tests for testing that previous exceptions are passed to exceptions to the current PR

Sephster:
Thanks so much @ceeram. It doesn't need to be the whole project, I think we are currently a bit patchy on our approach with this. I think if you just alter the files you are editing here that would be grand thanks. Thanks for your efforts!

Sephster:
Thanks @ceeram

ceeram:
Thanks for the quick merge. Any plans for a new tagged release?

PR #966, opened 2018-11-13
Replace fqn with unqualified name

ceeram:
Replace fqn with unqualified name, refs https://github.com/thephpleague/oauth2-server/pull/965#issuecomment-438011695


Sephster:
Thanks so much @ceeram. This is awesome. Merging in. Cheers for your work!

PR #969, opened 2018-11-15
Fix bc breaking change

ceeram:


Sephster:
@ceeram thank you for reverting the function names. The only thing I wanted to raise for discussion here is the moving of the setting of keys. The reason this had been moved to the prototype was so that the keys weren't set over and over again, I think this change means that the keys might be set again when they are already present.

Please can you explain why the moving of the key setting to the constructor has caused you issues?

Sephster:
@ceeram I went ahead and reverted the interface changes. I've then merged master into this branch so we can discuss the reverted key setting location in this PR. Hope this is ok. Many thanks

ceeram:
i moved it to ensure privateKey was set to the instance created in getResponseType, since it was only being set in constructor now. People need to set those themselves when creating a new instance in this method when overriding, while this wasnt needed when just calling parent. As said, we have solved it differently now. So if your solution works as well, im fine with it. Just fyio, only reverting the methodname change as you did now will still not be 100% BC with original code.

ceeram:
You removed the tests that were added in the PR #960 . If you would add them back and they still pass they wont at leats totally break the changes for stateless, just that privatekey might not be set in responseTypes created in overridden methods.

ceeram:
Re-added tests from #960 in https://github.com/thephpleague/oauth2-server/pull/971

The move of setting keys is up to you if you move that back or not.

ceeram:
Added a test to prove the BC break concerning the move of key setting.

ceeram:
This was our original code that was broken, and would still be broken by current master:
```
class AuthorizationServer extends BaseAuthorizationServer
{
    protected function getResponseType()
    {
        $this->responseType = new OpenIdBearerTokenResponse();

        return parent::getResponseType();
    }
}
```

Sephster:
Thanks @ceeram - I see now why this is causing issues. Apologies I haven't had time to progress this yet as I was working over the weekend.

Have you looked at changing your implementation to pass the `OpenIdBearerTokenResponse()` in via the constructor instead? If so, did you encounter any blockers?

ceeram:
Indeed thats the way we worked around it now, passing the object as contructor argument. Still the move of the key setting is a braking change. People can update their code to set them themselves ofcourse, but that would not be desired in a minor version upgrade.

Example fix for users:
     
```
class AuthorizationServer extends BaseAuthorizationServer
{
    protected function getResponseType()
    {
        $this->responseType = new OpenIdBearerTokenResponse();
        $this->responseType->setPrivateKey($this->privateKey);
        $this->responseType->setEncryptionKey($this->encryptionKey);

        return parent::getResponseType();
    }
}```

Above fix would in userland code is not needed when merging in this PR. That will make the stateless server change backwards compatible with implementations that extend AuthorizationServer and create new responseType objects in their overrided method

PR #971, opened 2018-11-16
Re-add tests from #960

ceeram:


ceeram:
Added back the tests from #960 to ensure this behavior is unchanged after the method name revert in #970

tests/AuthorizationServerTest.php
```diff
@@ -107,9 +107,58 @@ public function testGetResponseType()
...
 
+        $responseTypeA = $method->invoke($server);
+        $responseTypeB = $method->invoke($server);
+        $this->assertInstanceOf(BearerTokenResponse::class, $responseTypeA);
+        $this->assertInstanceOf(BearerTokenResponse::class, $responseTypeB);
+        $this->assertNotSame($responseTypeA, $responseTypeB);
         $this->assertInstanceOf(BearerTokenResponse::class, $method->invoke($server));
     }
 
+    public function testGetResponseTypeStateless()
```
Sephster: This is a replication of the test below. I just changed the name

tests/AuthorizationServerTest.php
```diff
@@ -107,9 +107,58 @@ public function testGetResponseType()
         $method = $abstractGrantReflection->getMethod('getResponseType');
         $method->setAccessible(true);
 
+        $responseTypeA = $method->invoke($server);
```
Sephster: I removed this section of the tests as it is replicated in the test below as far as I can tell.

ceeram:
In this pr some asserts were removed, https://github.com/thephpleague/oauth2-server/pull/970/files#diff-257ce39e10ed4b9dcbbf1a46a2b4dab6L110

indeed not all of it, but some.

Issue #974, opened 2018-11-22
Dangerously incorrect documentation (re: public/private keys)


krozinov:
Hi team,

There is a critical mistake in the wording on the Installation page (https://oauth2.thephpleague.com/installation/) under the section "Generating public and private keys"

It says:
> The public/private key pair is used to sign and verify JWTs transmitted. The Authorization Server possesses the public key to sign tokens and the Resource Server possesses the corresponding private key to verify the signatures. To generate the private key run this command on the terminal

It should say:
> The public/private key pair is used to sign and verify JWTs transmitted. The Authorization Server possesses the **private** key to sign tokens and the Resource Server possesses the corresponding **public** key to verify the signatures. To generate the private key run this command on the terminal

The current documentation is dangerously wrong - instructing users to commit a potentially serious security issue.

Sephster:
Merged. Thanks for spotting

PR #976, opened 2018-11-22
Fix the location of public and private keys

krozinov:
This is described in Issue #974


Sephster:
Thanks for spotting this. Merged

Issue #961, opened 2018-11-06
In the 8.0.0 branch the AuthCodeGrant does not accept client_id and client_secret through HTTP Basic Authentication


Devristo:
First of all thanks for this great library! I have been playing around with the 8.0.0 branch because we need the `isConfidential` / PKCE functionality. However we stumbled upon a regression: Clients can no longer authenticate themselves using HTTP Basic authentication. 

This seem to have happened in https://github.com/thephpleague/oauth2-server/pull/938. More specifically: The `$this->validateClient()` on https://github.com/thephpleague/oauth2-server/pull/938/files#diff-4390e0360afea418f1505565796e2fb9L73 was checking for HTTP Basic Auth  but it has been moved down the method.

A fix would be very welcome. Although we work around this for now in our middleware by setting client_id and client_secret in the request when we find them through HTTP Basic auth.

Sephster:
Hi @Devristo - thanks for getting in touch and thanks for the kind words regarding the library.

Is your code calling the `validateClient()` function at all? There was a change in the 8.0.0 branch where clients must now identify themselves as confidential. If they aren't confidential, then no client validation takes place as the client cannot be trusted to keep its credentials secret.

Devristo:
In this case I am not using the PKCE flow but a confidential client. However the 8.0.0 branch changed the way the AuthCodeGrant is implemented.

`src/Grant/AuthCodeGrant.php`
![image](https://user-images.githubusercontent.com/1564257/48115647-e8318800-e263-11e8-91bd-c8fd1e359bab.png)

Previously the `validateClient` call would check the request params AND the headers using `getBasicAuthCredentials`. However now the `respondToAccessTokenRequest` only checks the request params and will fail if the client is using HTTP Basic authentication.

If you want I could create an  test that would succeed on 7.x and fail on 8.x to illustrate the issue?

Sephster:
Ahh I think I see. I presume it is issuing an _invalid request_ exception because you aren't passing a client ID in as a request parameter, correct? 

I will look at getting a fix sorted for this soon and double checking everything. Thanks for reporting this.

Devristo:
Exactly! Thanks for picking this up

Sephster:
Closing as issue now resolved by PR #978

PR #978, opened 2018-11-29
 Fixed respondToAccessTokenRequest using Http Basic Auth

Devristo:
** Note ** This PR targets 8.0.0-branch instead of the master branch. The Contribution Guidelines don't allow it. But this issue does not exist in the master-branch.

Resolves https://github.com/thephpleague/oauth2-server/issues/961

Changes:

 - Fixes 8.0.0 regression by re-accepting `client_id` from `Authorization` header in `AuthCodeGrant` `respondToAccessTokenRequest`.
 - Added regression test

### DIFF
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -171,15 +171,7 @@ abstract class AbstractGrant implements GrantTypeInterface
      */
     protected function validateClient(ServerRequestInterface $request)
     {
-        list($basicAuthUser, $basicAuthPassword) = $this->getBasicAuthCredentials($request);
-
-        $clientId = $this->getRequestParameter('client_id', $request, $basicAuthUser);
-
-        if (is_null($clientId)) {
-            throw OAuthServerException::invalidRequest('client_id');
-        }
-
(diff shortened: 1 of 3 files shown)


Sephster:
This looks good to me @Devristo. Many thanks for picking this up. I like the solution you have come up with. The only point I wanted to raise is around the new `getClientCredentials()` function. Would it be possible to change this to be private? I can't think of many reasons to override this at present but if you can think of any please jump in. Cheers again!

Sephster:
If you could, that would be great cheers @Devristo

Devristo:
Hmm it cannot be `private` because it is being called from its subclass `AuthCodeGrant`.

```
1) LeagueTests\Grant\AuthCodeGrantTest::testRespondToAccessTokenRequest
Error: Call to private method League\OAuth2\Server\Grant\AbstractGrant::getClientCredentials() from context 'League\OAuth2\Server\Grant\AuthCodeGrant'
```

My suggestion is to leave it `protected`.

Sephster:
Thanks for the PR @Devristo - I deleted some of the methods from the test as they weren't needed but otherwise looks good! Thanks for your contribution and spotting this

Issue #715, opened 2017-03-17
AccessTokenEntityInterface::getNewToken() and scopes parameter.


Danack:
Hi,

Please could you clarify for me what an implementation of AccessTokenEntityInterface should do with the scopes parameter in getNewToken() ? or any of the parameters actually.

The getNewToken method is called, and then immediately scopes are added to the token again, which lead to the scopes being double entered on my implementation. 

https://github.com/thephpleague/oauth2-server/blob/bf7084a147e8072b889347f072a081530b7e0956/src/Grant/AbstractGrant.php#L335-L342

cheers
Dan

Sephster:
Hi @Danack - Apologies for the delayed response. This looks like a bug to me as you are correct in that we are likely double entering scopes. This shouldn't created duplicates against the access token though as these are indexed on the scope ID:

https://github.com/thephpleague/oauth2-server/blob/9fc288ce53bbd198ceb7be604ca00dc3910a7d82/src/Entities/Traits/TokenEntityTrait.php#L42-L45

The easiest fix would be to remove the duplicate code from the Abstract Grant but I will need to do a thorough check to make sure this won't break anything. Thanks for reporting this

lordrhodos:
I hit the same issue as well and was wondering why the `AbstractGrant` adds the scopes again after they had been passed to the `getNewToken` method of the repository. I was not using indexed entries in the scopes collection (common doctrine `ArrayCollection`). Changed that for now, but would love to see a fix in the future.

Sephster:
This issue was originally tackled in https://github.com/thephpleague/oauth2-server/pull/553 but reverted as it introduced a breaking change. I think it should be reinstated for version 8.

crtl:
I wanted to ask the same question.
Either `getNewToken` or `AbstractGrant` should setup the token.
Why call a factory when I do everything myself?

If it is to execute additional logic, why dont you use an event system?

PR #983, opened 2019-01-02
Bugfix: prevent double entering parameters passed to AccessTokenRepositoryInterface::getNewToken() in AbstractGrant::issueAccessToken function

lordrhodos:
In `AbstractGrant::issueAccessToken` the client, scope and user identifier are set manually after they have been passed to the `AccessTokenRepositoryInterface::getNewToken` function as parameters (suggesting they should be set via the concrete implementation).

This PR removes the explicit `set*` and `add*` calls to prevent double setting / entering of data. Especially for scope entries not stored in an indexed array this can become buggy otherwise.

https://github.com/thephpleague/oauth2-server/blob/2a16dbeb7f83135db25a1a2b416f6696e5906f5a/src/Grant/AbstractGrant.php#L396-L403

Related issue:
https://github.com/thephpleague/oauth2-server/issues/715

### DIFF
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -394,13 +394,7 @@ abstract class AbstractGrant implements GrantTypeInterface
         $maxGenerationAttempts = self::MAX_RANDOM_TOKEN_GENERATION_ATTEMPTS;
 
         $accessToken = $this->accessTokenRepository->getNewToken($client, $scopes, $userIdentifier);
-        $accessToken->setClient($client);
-        $accessToken->setUserIdentifier($userIdentifier);
         $accessToken->setExpiryDateTime((new DateTime())->add($accessTokenTTL));
-
-        foreach ($scopes as $scope) {
-            $accessToken->addScope($scope);
-        }
 
         while ($maxGenerationAttempts-- > 0) {
             $accessToken->setIdentifier($this->generateUniqueIdentifier());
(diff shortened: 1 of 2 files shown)


Sephster:
I think this would need to be a change for version 8. There is a risk that some implementations will do nothing with the constructor and rely on the setting functions to set their clients etc.

Although I would hope that most people would have set the client etc in the constructor, we cannot guarantee this. Accepting this pull request would risk introducing a breaking change for some users.

When considering the wider package, we need to ensure we are implementing consistent behaviour as well. At the moment, authCodes can also have clients set against them. Like access tokens, they have a `setClient()` method, but the main difference is that you cannot set a client via a constructor for the authcode.

It looks like the change I am suggesting was originally implemented in https://github.com/thephpleague/oauth2-server/pull/553 but removed as it was a breaking change. It was never reimplemented but I think this would be the best course of action to take.

I will flag the corresponding issue for version 8. Thanks for taking a look at this and sorry I am not merging this in at this time.

lordrhodos:
> When considering the wider package, we need to ensure we are implementing consistent behaviour as well. At the moment, authCodes can also have clients set against them. Like access tokens, they have a setClient() method, but the main difference is that you cannot set a client via a constructor for the authcode.

@Sephster 

Indeed, there are a couple of general / global issues regading DI for this library. 

Is there an agreement amongst core members / maintainers if the library should favour constructor injection over setter injection? 

I mean in the [AbstractGrant](https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AbstractGrant.php) there is no constructor injection, instead all depencies are injected via setter injection in

https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationServer.php#L126-L142

so there is no real dependecy coupling for each grant type that is actually needed. What are your thoughts on this? Wouldn't it be best to just inject the services (e.g. repositories) needed into each dedicated grant type instead or am I missing something here?

PR #988, opened 2019-01-20
Cleanup: remove unused local variable $scopeEntity from ImplicitGrantTest

lordrhodos:
I recognized that the `ImplicitGrantTest` defines an unused `$scopeEntity` variable in serveral tests. This maybe a leftover from a copy and paste action somewhen in the past.

This PR

* removes the unused local variables `$scopeEntity` in `ImplicitGrantTest`


Sephster:
Thanks for spotting and fixing! Much appreciated.

Issue #806, opened 2017-10-31
Token Revocation question


chervand:
Are you going to include something [RFC7009](https://tools.ietf.org/html/rfc7009) compliant to the lib that could be used in a straightforward way like `$server->respondToRevokeTokenRequest($request, $response);`?

Sephster:
I think it would be good to work towards this so would hope for it to be included in the future. Thanks for the suggestion

jacobweber:
I might try working on this for my own purposes. I can try to submit a PR if you want.

I could add a `respondToRevokeTokenRequest ` method to `AuthorizationServer`. It would need to duplicate some of the code in `AbstractGrant->validateClient`, since that's not available to `AuthorizationServer`.

Then, depending on the `token_type_hint`, it would call either `accessTokenRepository->revokeAccessToken` or `refreshTokenRepository->revokeRefreshToken`. You'd probably need to pass the refresh token repository into the method, since it's not otherwise available to `AuthorizationServer`.

I think we'd also want to pass in a flag for whether to revoke the access token, when `token_type_hint=refresh_token`, since this is optional.

jacobweber:
Sure, if you don’t mind me moving that out of the Grant classes, and maybe into its own class, I can take that approach.

jacobweber:
Before I go ahead with writing tests and submitting a pull request, could I ask if this is something you're interested in at all, and if my approach in [this commit](https://github.com/jacobweber/oauth2-server/commit/54c879aa1aad6d03c12538d562a562d573d0aaeb) is reasonable? It's based on master.

Basically I did this:
- Copied the client validation code from `AbstractGrant` into a `ClientValidator` class (later I can separate this into its own PR, and change the Grants to use this same class).
- Added a `RevokeTokenHandler` class to handle revoke requests. It checks to see if the token is an access or refresh token, and revokes it if possible. If `$canRevokeAccessTokens` is true, it will allow you to revoke access tokens, and it will revoke the associated access token when you revoke a refresh token.
- Added a `enableRevokeTokenHandler` method to `AuthorizationServer`, which can be used like this:
```
$refreshTokenRepository = [...];
$authServer = new \League\OAuth2\Server\AuthorizationServer([...]);
$handler = new \League\OAuth2\Server\RevokeTokenHandler($refreshTokenRepository);
$authServer->enableRevokeTokenHandler($handler);
```
[...]

PR #995, opened 2019-02-08
Add token revocation support

jacobweber:
My attempt to implement a fix for #806.

- Moved some of the validation methods from `AbstractGrant` into a `RequestValidatorTrait` trait, so they can be used by non-grant classes. This trait includes some abstract methods, to get the client repository and the grant identifier. Those could be refactored into arguments to `validateClient`, if that's preferable.
- Added `RevokeTokenHandler` class to handle revocation. The constructor requires the refresh token repository and the public key as arguments. Uses existing repository methods for revocation.
- Added `enableRevokeTokenHandler` method to `AuthorizationServer` to be used during setup.
- Added `respondToRevokeTokenRequest` method to `AuthorizationServer`, to be used in a POST request, similar to `respondToAccessTokenRequest`. CORS support is up to the application.
- Example setup (based on [this](https://oauth2.thephpleague.com/authorization-server/refresh-token-grant/)):
```
// Init our repositories
$clientRepository = new ClientRepository();
$accessTokenRepository = new AccessTokenRepository();
$scopeRepository = new ScopeRepository();
$refreshTokenRepository = new RefreshTokenRepository();

// Path to public and private keys
$privateKey = 'file://path/to/private.key';
//$privateKey = new CryptKey('file://path/to/private.key', 'passphrase'); // if private key has a pass phrase
$encryptionKey = 'lxZFUEsBCJ2Yb14IF2ygAHI5N4+ZAUXXaSeeJm6+twsUmIen'; // generate using base64_encode(random_bytes(32))
$publicKey = 'file://path/to/public.key';

// Setup the authorization server
$server = new \League\OAuth2\Server\AuthorizationServer(
    $clientRepository,
    $accessTokenRepository,
    $scopeRepository,
    $privateKey,
    $encryptionKey
);

$handler = new \League\OAuth2\Server\RevokeTokenHandler($refreshTokenRepository, $publicKey);

// Enable the revoke token handler on the server
$server->enableRevokeTokenHandler($handler);
```
- Example implementation:
```
$app->post('/revoke', function (ServerRequestInterface $request, ResponseInterface $response) use ($app) {

    /* @var \League\OAuth2\Server\AuthorizationServer $server */
    $server = $app->getContainer()->get(AuthorizationServer::class);

    // Try to respond to the request
    try {
        return $server->respondToRevokeTokenRequest($request, $response);

    } catch (\League\OAuth2\Server\Exception\OAuthServerException $exception) {
        return $exception->generateHttpResponse($response);

    } catch (\Exception $exception) {
        $body = new Stream('php://temp', 'r+');
        $body->write($exception->getMessage());
        return $response->withStatus(500)->withBody($body);
    }
});
```
- You can set `$canRevokeAccessTokens` in the `RevokeTokenHandler` constructor if you want to allow access tokens to be revoked, since the spec describes this as optional.
- If you revoke a refresh token, and the above setting is true, it will also revoke the associated access token.
- If you revoke an access token, it will **not** revoke the associated refresh token, since I don't think this is possible without adding a new repository method.

### DIFF
--- a/src/AuthorizationServer.php
+++ b/src/AuthorizationServer.php
@@ -13,6 +13,7 @@ use DateInterval;
 use Defuse\Crypto\Key;
 use League\Event\EmitterAwareInterface;
 use League\Event\EmitterAwareTrait;
+use League\OAuth2\Server\RevokeTokenHandler;
 use League\OAuth2\Server\Exception\OAuthServerException;
 use League\OAuth2\Server\Grant\GrantTypeInterface;
 use League\OAuth2\Server\Repositories\AccessTokenRepositoryInterface;
@@ -33,6 +34,11 @@ class AuthorizationServer implements EmitterAwareInterface
      * @var GrantTypeInterface[]
      */
     protected $enabledGrantTypes = [];
+
+    /**
+     * @var RevokeTokenHandler
+     */
+    protected $revokeTokenHandler = null;
 
     /**
      * @var DateInterval[]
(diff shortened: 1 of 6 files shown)


jacobweber:
BTW, I couldn't figure out how to get it to pass both tests, if I have a function that sometimes returns null. The style checker wants me to change "return null" to "return", but if I do that, I get a CI error.

Sephster:
Is there any alternative approach you can take? I am not a fan of blank returns and would prefer the code is more explicit.

For example, if token verification fails, could we use exceptions to issue an error rather than aborting the execution?

jacobweber:
Sure, I refactored it to avoid returning null, and just made it revoke the token right away.

neodc:
Is there any reason this PR is not merged? I'd like to use the revoke in one of my project

simonhamp:
@neodc right now it looks like there are merge conflicts

jacobweber:
I think this may need some re-working – it was written for 7.x, and 8.0.0 has come out since then. Not sure I'll have time right now to do it, but if you guys are still interested in merging it, I can try.

PR #1000, opened 2019-03-08
Optional Refresh Tokens

### LINKED ISSUE #649: Making refresh tokens optional
I'm working through the V4->V5 upgrade (we're using Diactoros to work around the PSR-7 requirement for now) and one thing that has me stumped is how to make refresh tokens optional. This was very easy to do in V4, but it isn't clear to me how to do it in V5.

I realize it's a good idea to use them (not using them when we first implemented OAuth using V4 has bitten us in a number of ways) and we're in the process of migrating our existing production clients to use them, but not all of them are ready for that transition yet, and we're contractually obligated to respect that.

Looking at how `RefreshTokenRepository` is set up, I originally thought I could just have `getNewRefreshToken` return `null`, but that breaks everything in `AbstractGrant::issueRefreshToken()`. I suppose I could just have `persistNewRefreshToken` not do anything (i.e. not actually save the token to the database), but the token will still get included in the response, which we're trying to avoid because it could potentially break a client application that isn't expecting that field to be in the response (or, conversely, someone using a standard OAuth client that can parse it, and that then may try to use a refresh token that was never saved to the database).

Any guidance? Am I missing something?

filecage:
This PR resolves #649 by allowing a `RefreshTokenRepository` to return `null` on `getNewRefreshToken()`.

I chose this way of implementing the optional Refresh Token because from what I've seen it's the only way of doing it without breaking BC - I'd actually prefer not passing a `RefreshTokenRepository` at all when I don't want to issue a refresh token, but that would mean having a nullable type hint and that would break support for PHP 7.0.

I'm looking forward to your feedback!

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -41,7 +41,6 @@ enabled:
   - print_to_echo
   - short_array_syntax
   - short_scalar_cast
-  - simplified_null_return
   - single_quote
   - spaces_cast
   - standardize_not_equal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Refresh Tokens are now optional (#649)
+
 ## [7.3.2] - released 2018-11-21
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Changed
-- Refresh Tokens are now optional (#649)
+- RefreshTokenRepository can now return null, allowing refresh tokens to be optional. (PR #649)
 
 ## [7.3.3] - released 2019-03-29
 ### Added
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -478,7 +478,7 @@ abstract class AbstractGrant implements GrantTypeInterface
     {
         $refreshToken = $this->refreshTokenRepository->getNewRefreshToken();
         if ($refreshToken === null) {
-            return null;
+            return;
         }
 
         $refreshToken->setExpiryDateTime((new DateTime())->add($this->refreshTokenTTL));
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -477,6 +477,7 @@ abstract class AbstractGrant implements GrantTypeInterface
     protected function issueRefreshToken(AccessTokenEntityInterface $accessToken)
     {
         $refreshToken = $this->refreshTokenRepository->getNewRefreshToken();
+
         if ($refreshToken === null) {
             return null;
         }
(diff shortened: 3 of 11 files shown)


filecage:
@Sephster The PHPStan and Style CI checks have incosistent behaviour. Declaring `return null;` makes Style CI remove the `null` and making it a void return which is then failing in Travis CI due to PHPStan.

Please advise.

Sephster:
I think it is better to be explicit so I would favour removing the StyleCI check here.

filecage:
I agree, thank you! Changed and pushed.

filecage:
@Sephster please tell me whether you'd like me to remove the `simplified_null_return` style CI fixer to make all the checks pass. I have no experience with style CI and cannot tell whether removing it will break anything else.

filecage:
@Sephster short reminder that this PR can not pass the checks if you don't give me advice whether to change the style CI settings.

Your feedback is very much appreciated.

Sephster:
Sorry @filecage I've been busy on another project which is now coming to a close so will have some time to look at this this weekend. I had a quick glance to see if it was obvious what needs to be changed but I can't see the rule that is being triggered. I don't think it is the `simplified_null_return` as in the docs it states that this rule:

> A return statement wishing to return void should not return null.

The function in question isn't explicitly returning void so I'm not clear if this is linked. It definitely seems to be the most likely candidate from naming alone though. 

If you are happy to, go ahead and remove this in your branch as it shouldn't cause any issues and we will the definitely know if it is this rule triggering the conflict. Cheers

filecage:
@Sephster StyleCI seems to focus on return statements and is not aware of the function or method context. Removing the option made the check pass.

If there is nothing else left that you would like to change, feel free to merge.

Sephster:
Thanks @filecage - I've made some minor adjustments to the changelog and formatting but otherwise looks good. I think long term this will be a stop gap and likely reverted when version 8 comes along as I'd like to eventually implement Alex's proposed plan of just not passing the refresh repo to grants if you don't want to issue a refresh token.

At the moment though, this is a good solution for people that need to have optional refresh tokens now.

filecage:
@Sephster thank you for your feedback. I also think that the proposal from #1005 is a cleaner way of implementing this but in terms of BC this one is the safer solution. So it makes absolute sense to me to revert this in a `8.x` release.

Sephster:
No worries. Thanks for your contribution and patience while I reviewed this.

ambrous:
Hi there!

If I understand right Optional Refresh Token is only when you try to get Access Token second and next times. How it is noted on the scheme here https://tools.ietf.org/html/rfc6749#page-10 you obtain Refresh Token always when you do the first request to get Access Token, however, then you are able to get a new Refresh Token or not.

Moreover, I have tested the version (7.4.0) with that fix () and it looks like it does not work with https://github.com/thephpleague/oauth2-client (2.4.1) . I get an error of **Required parameter not passed: "refresh_token"**

ambrous:
I think the best way to do support of Optional Refresh Token is to change src/Grant/RefreshTokenGrant.php so that we are able to send a flag to the constructor to have capacity to set that behavior up and change method respondToAccessTokenRequest so that we are able to return old refresh token instead of issue of a new one. However, in that case we have to change Access Token of the old our Refresh Token, and because of this we will have orphan access tokens.

But, to do thiswe have to add getByIdentifier method into RefreshTokenRepository including adding that method into the RefreshTokenRepositoryInterface what breaks backward compatibility.

PR #1003, opened 2019-03-22
Moved creation of AuthorizationRequest into separate function

iNviNho:
#### Background
In our OAUTH server we were required to implement handling of `nonce` parameter which can be given as a part of authorization request.

For this requirement, we had to override some functionality and add a `nonce` as another property of `AuthorizationRequest`. 

It would be easier to override one simple function and return our own instance of `AuthorizationRequest` then overriding a whole `validateAuthorizationRequest` function.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -268,7 +268,8 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
 
         $stateParameter = $this->getQueryStringParameter('state', $request);
 
-        $authorizationRequest = new AuthorizationRequest();
+        $authorizationRequest = $this->createAuthorizationRequest();
+
         $authorizationRequest->setGrantTypeId($this->getIdentifier());
         $authorizationRequest->setClient($client);
         $authorizationRequest->setRedirectUri($redirectUri);
@@ -308,6 +309,14 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
         }
 
         return $authorizationRequest;
+    }
+
+    /**
+     * @return AuthorizationRequest
+     */
+    private function createAuthorizationRequest() {
+
+        return new AuthorizationRequest();
     }
 
     /**
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -314,7 +314,7 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
     /**
      * @return AuthorizationRequest
      */
-    private function createAuthorizationRequest()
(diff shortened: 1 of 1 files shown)


src/Grant/AuthCodeGrant.php
```diff
@@ -310,6 +311,14 @@ public function validateAuthorizationRequest(ServerRequestInterface $request)
         return $authorizationRequest;
     }
 
+    /**
+     * @return AuthorizationRequest
+     */
+    private function createAuthorizationRequest()
```
Sephster: I think this should be protected rather than private or you won't be able to override it.

Sephster:
The same issue also applies to the Implicit Grant. If both grants need this feature, it would probably be better to put this in the AbstractAuthorizeGrant but I think this would be a breaking change.

Our version 8 is already pretty full at the moment so I'm inclined to not accept this change at this time although I can see the value in it.

I will leave this open for a week to see if there is any appetite to introduce this and if so, we can look at possibly including it in version 8 although there would need to be compelling reasons to do so as I don't want to introduce too many more breaking changes.

Thanks very much for raising this though.

src/Grant/AuthCodeGrant.php
iNviNho: Changed, thank you.

iNviNho:
@Sephster thank you very much for your review, i understand your concerns. 

If you decide to accept this feature, i can move that function to AbstractAuthorizeGrant so other grants can benefit as well.

TheLevti:
Bump

lordrhodos:
> The same issue also applies to the Implicit Grant. If both grants need this feature, it would probably be better to put this in the AbstractAuthorizeGrant but I think this would be a breaking change.

@Sephster why would this be a breaking change? If it is, could this go into `9.x`?

Sephster:
I think that @lordrhodos's solution in PR #1110 which has just been merged into the v9 branch now makes this change redundant. As such, I'm going to close this PR. Apologies for not progressing this sooner @iNviNho - I hope @lordrhodos's solution will serve your needs. Many thanks

lordrhodos:
> It would be easier to override one simple function and return our own instance of AuthorizationRequest then overriding a whole validateAuthorizationRequest function.

@Sephster I think @iNviNho has a point here in easing the pain. Although #1110 lays the foundation for more flexibility I still think the creation of the `AuthorizationRequestInterface` implementation should be separated from the functions where it is created at the moment. 

I will come up with a PR ;-)

Sephster:
Yeah I think I was a bit hasty closing this. Shouldn't review PRs at 1am :) Will take a look at your solution. Thanks

Issue #958, opened 2018-10-30
Error Response is not OAuth2 compliant


thuethe:
On Issuing an Access Token the OAuth 2.0 Server produces an error response like:
```json
{
    "error": "invalid_client",
    "message": "Client authentication failed"
}
```
with optional `hint` in some cases.

Specs compliant would be an error response like:
```json
{
    "error": "invalid_client",
    "error_description": "Client authentication failed",
    "error_uri": "..."
}
```
`error_uri` and `error_description` are optional.

From the specs (https://tools.ietf.org/html/rfc6749#section-5.2):
> The authorization server responds with an HTTP 400 (Bad Request)
   status code (unless specified otherwise) and includes the following
   parameters with the response:
   error
         REQUIRED.  [...]
   error_description
         OPTIONAL.  Human-readable ASCII [USASCII] text providing
         additional information, used to assist the client developer in
         understanding the error that occurred.
         Values for the "error_description" parameter MUST NOT include
         characters outside the set %x20-21 / %x23-5B / %x5D-7E.
   error_uri
         OPTIONAL.  A URI identifying a human-readable web page with
         information about the error, used to provide the client
         developer with additional information about the error.
         Values for the "error_uri" parameter MUST conform to the
         URI-reference syntax and thus MUST NOT include characters
         outside the set %x21 / %x23-5B / %x5D-7E.

Sephster:
Thanks both. I will aim to look at this this evening.

lordrhodos:
> Also, the **state** query parameter ist missing in error responses. The RFC states:
>
> ```
>state
>         REQUIRED if a "state" parameter was present in the client
>        authorization request.  The exact value received from the
>         client.
>```

just to clarify, this seems to apply only to the following grant types:

* Authorization Code Grant (https://tools.ietf.org/html/rfc6749#section-4.1.2.1)
* Implicit Grant (https://tools.ietf.org/html/rfc6749#section-4.2.2.1)

The error responses for the following grant types *do not require* the `state`parameter:

* Resource Owner Password Credentials Grant
* Client Credentials Grant

(both link to [section 5.2](https://tools.ietf.org/html/rfc6749#section-5.2) in the spec for the error response)

Spomky:
@lordrhodos tha’s right. Broadly speaking, the `state` parameter is should be part of any response (success or failure) from the authorization endpoint.
[...]

PR #1006, opened 2019-03-29
spec compliant 'error_description' but keep 'message' for BC

marc-mabe:
see #958 

This is only fixing one part of the problem explained in #958.
I think the other problem with missing `state` parameter should be done separately.

### DIFF
--- a/src/Exception/OAuthServerException.php
+++ b/src/Exception/OAuthServerException.php
@@ -59,8 +59,8 @@ class OAuthServerException extends Exception
         $this->hint = $hint;
         $this->redirectUri = $redirectUri;
         $this->payload = [
-            'error'   => $errorType,
-            'message' => $message,
+            'error'             => $errorType,
+            'error_description' => $message,
         ];
         if ($hint !== null) {
             $this->payload['hint'] = $hint;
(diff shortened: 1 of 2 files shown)


Sephster:
Thanks @marc-mabe - although the duplication isn't ideal it is a way forwards without introducing a breaking change in the version 7 branch. Happy to merge this in. 

Will look to make a full change to use `error_description` in version 8.

Sephster:
@marc-mabe I was going to create a release for this but I'm now thinking this isn't required.

The `error_description` is applicable for authorization requests from the implicit grant and auth code grant.

At the moment I think we do need to have a way to optionally specify the `error_description` but if we want to make this spec compliant, I don't think moving to replace the existing `message` is the way to go about this.

On the surface, it looked like someone had put the property in as a wrong name but on closer inspection, I think this is something custom to this library to assist developers.

I would appreciate your thoughts on this but given this, I will likely revert this PR.

driesvints:
This seems to have broken the Passport build but I can see that this isn't a breaking change so we should just fix our build for this: https://travis-ci.org/laravel/passport/builds/519352496

Sephster:
Thanks @driesvints - sorry this caused you issues. We made every effort to ensure it didn't break functionality but looks like this did break your test. I will try and remember to check changes like this against Passport in future and be proactive with a fix.

driesvints:
No worries at all. I don't consider this to be a breaking change. In the end it only broke our test suite because we're doing some a pretty aggressive assertion there. It's probably better if we used a "contains" assertion instead.

PR #1009, opened 2019-05-11
Skip SHA256 verifier if system doesn't support sha256

iansltx:


Sephster:
Thanks @iansltx - I never finished this feature as I couldn't find the `hash_algos` algorithm. Thanks for sorting this

PR #1010, opened 2019-05-11
Ensure unvalidated ClientEntity gets throw/emit if they return null

iansltx:
In many cases, we validate client info before pulling from client itself
from the repository, in which case it's safe to assume that you can grab
the client once validation passes. However on implicit/auth code grants
we don't have this guarantee due to non-confidential clients that just
reference the client ID. In those cases the client may supply a client
ID that doesn't exist, and we don't do a validation step before pulling
it from the repo.

The issue with that is that ClientRepository doesn't actually enforce
returning a ClientInterface via typehint, nor does it even suggest an
exception to throw if the client doesn't exist. So in most places we
do an instanceof check after the repository returns and throw/emit an
error event if the client doesn't exist.

This approach ends up being a bit error-prone; we missed one case where
we should've been doing this check: in the access token request on an
auth code grant. We don't do enough validation beforehand to assume that
the incoming request has an accurate client ID, so L96 could absolutely
be a method call on a non-object.

This commit centralizes the return-check-emit-throw logic so it's a
one-liner for wherever we need it, including the access token request
processor for auth code grants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Public clients can now use the Auth Code Grant (PR #938)
 - `isConfidential` getter added to `ClientEntity` to identify type of client (PR #938)
 - Function `validateClient()` added to validate clients which was previously performed by the `getClientEntity()` function (PR #938)
+- Add a new function to the AbstractGrant class called `getClientEntityOrFail()`. This is a wrapper around the `getClientEntity()` function that ensures we emit and throw an exception if the repo doesn't return a client entity. (PR #1010)
 
 ### Changed
 - Replace `convertToJWT()` interface with a more generic `__toString()` to improve extensibility; AccessTokenEntityInterface now requires `setPrivateKey(CryptKey $privateKey)` so `__toString()` has everything it needs to work (PR #874)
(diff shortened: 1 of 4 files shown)


iansltx:
Not sure what you're referring to on `getClient()`.

If you're thinking of rolling it into `getClientEntity()` (on the repository interface), my concern is that we would have to revise that interface to switch from returning a nullable `ClientInterface` to forcing a `ClientInterface` return and throwing a specific exception class if no client is found. Since this is a behavior change from how the interface currently works, we would want to explicitly/loudly break BC, probably by enforcing the return type via a typehint (vs. the current setup that only mentions the return type in the docblock).

One additional snag is that PHP doesn't allow enforcement of exceptions thrown by an interface, so the absolute most we can do if we went down that route would be to list the exception in `@throws` in the docblock. Whereas with the nullable return option we could actually enforce via a return typehint for PHP 7.1+ (which would be reasonable since 7.0 and below are now EOL).

I'd be less worried about this stuff, but given that this is an auth library it feels like defensive coding is warranted. Which this current PR ensures. Pushing that responsibility to the interface implementor certainly makes the code in this library less complex (and removes instanceof/null checks if we enforce the typehint on the interface), but I'll defer on whether that makes it too easy for interface implementors to get it wrong.

Sephster:
Reading back on my comment I'm not sure what I was referring to either :) Looks good to me so happy to merge in. Thanks @iansltx

Issue #798, opened 2017-10-25
Issued time claim validation is not tolerant of timing drift


soundsgoodsofar:
We had very difficult to diagnose problems with valid oauth tokens being rejected immediately after being issued, and we tracked it down to a claims enforcement issue.

It turns out that `iat` and `nbf` timestamps are being set in `AccessTokenTrait` to now().  Then in `BearerTokenValidator` a ValidationData object is created, also with time now().

The problem is when these oauth tokens can be generated on different servers.  We have a pool of backend servers and, like most real world environments, their system times are susceptible to drift, even when using NTP.  A second or even fractional second of drift is occasionally enough to have one server issue an oauth token with a timestamp that fails to validate if the token is immediately used by the client and a different server is hit.

Personally I don't see any real value in using these and would prefer to drop them since all they really do right now is cause potential errors.  The "not valid before" timestamp is hardcoded, so users can't use it to issue future tokens even if they wanted to.

The only other option would be to add some arbitrary amount of wiggle-room to the values either on setting or validating, but then you're picking an arbitrary amount of allowable drift which may not be reasonable in different environments.

sergeyklay:
I propose to consider such a scenario:

1. User should explicitly set a timezone, this is not a responsibility oauth2-server
2. The oauth2-server should use `DateTime` and `DateTimeZone` objects instead of `time()` function

```php
interface DateTimeAwareInterface {
    public function getTimestamp(): int;

    public function getDateTime(string $format = 'Y-m-d H:i:s'): string;
}

trait DateTimeAwareTrait {
    protected $dateTime;

    public function __construct(DateTimeZone $timezone = null) {
        $this->dateTime = new DateTime('now', $timezone);
    }

    public function getTimestamp(): int {
        return $this->dateTime->getTimestamp();
    }

    public function getDateTime(string $format = 'Y-m-d H:i:s'): string {
        return $this->dateTime->format($format);
    }
}


$dc1 = new class(new DateTimeZone("Europe/London")) implements DateTimeAwareInterface {
    use DateTimeAwareTrait;
};

$dc2 = new class(new DateTimeZone("UTC")) implements DateTimeAwareinterface {
    use DateTimeAwareTrait;
};

var_dump(
    [
        $dc1->getDateTime() => $dc1->getTimestamp(),
        $dc2->getDateTime() => $dc2->getTimestamp()
    ]
);
```

```
array(2) {
  '2017-10-25 09:25:59' =>
  int(1508919959)
[...]

PR #1017, opened 2019-06-09
Issue #798: Allow leeway to handle clock skew

m777z:
Adding support and tests for time claim leeway. Support for this was added to the lcobucci/jwt repository to fix https://github.com/lcobucci/jwt/issues/191 ; this pull request adds that leeway implementation here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": ">=7.0.0",
         "ext-openssl": "*",
         "league/event": "^2.1",
-        "lcobucci/jwt": "^3.2.2",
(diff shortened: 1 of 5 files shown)


Sephster:
Thank you for your pull request @m777z. As this will be changing interfaces, it would need to be added to a major version. At this time, I don't think we would proceed with this support call due to a number of other additions I need to push through for an already pending new major. 

I've raised an issue to track this feature and add it in at a later date. Thank you very much for suggesting this though and apologies we are not going to implement it at this time.

PR #1023, opened 2019-06-25
Fix BasicAuth issues with Password Grant type

alshf89:
Currently we should send __username__ & __password__ in __POST__ request parameters, but when we send them via __BasicAuth__ it won't validate in __respondToAccessTokenRequest__ method in __League\OAuth2\Server\Grant\PasswordGrant__ Class.

__validateClient__ method which run in __respondToAccessTokenRequest__ method in __League\OAuth2\Server\Grant\PasswordGrant__ Class validate it but __validateUser__ don't.

So I add __BasicAuth__ validation in __validateUser__ method same as __validateClient__ in __League\OAuth2\Server\Grant\AbstractGrant__ Class.

### DIFF
--- a/src/Grant/PasswordGrant.php
+++ b/src/Grant/PasswordGrant.php
@@ -82,12 +82,14 @@ class PasswordGrant extends AbstractGrant
      */
     protected function validateUser(ServerRequestInterface $request, ClientEntityInterface $client)
     {
-        $username = $this->getRequestParameter('username', $request);
+        list($basicAuthUser, $basicAuthPassword) = $this->getBasicAuthCredentials($request);
+
+        $username = $this->getRequestParameter('username', $request, $basicAuthUser);
         if (is_null($username)) {
             throw OAuthServerException::invalidRequest('username');
         }
 
-        $password = $this->getRequestParameter('password', $request);
+        $password = $this->getRequestParameter('password', $request, $basicAuthPassword);
         if (is_null($password)) {
             throw OAuthServerException::invalidRequest('password');
         }


Sephster:
The OAuth spec for the password grant requires the user's name and password to be passed via the request. There is nothing that I can see in the spec about using basic auth for a username and password.

Typically it would be the client that would use basic auth to authenticate with the server. 

Apologies if I'm missing a section in the spec but I don't believe we should be supporting basic auth for username and password out the box.

alshf89:
Please check OAuth 2.0 specifications for password grant type here:
https://tools.ietf.org/html/rfc6749#section-1.3.3

And specially this one:
https://tools.ietf.org/html/rfc6749#section-2.3.1

It says:

The authorization server MUST support the HTTP Basic authentication scheme for authenticating clients that were issued a client password.

Including the client credentials in the request-body using the two parameters is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme (or other password-based HTTP authentication schemes).  The parameters can only be transmitted in the request-body and MUST NOT be included in the request URI.


Bshaffer OAuth2.0 package currently supports BasicAuth in Password Grant type.

http://bshaffer.github.io/oauth2-server-php-docs/grant-types/user-credentials/

Sephster:
Sorry again if I'm confusing issues here. The sections you have posted from the spec are relating to the client credentials, not the user credentials.

The library supports basic auth for client credentials but the username and password for the user should be passed via the body of the request as far as I'm aware.

alshf89:
Sorry if I sent unrelated section from spec...

Currently we migrate from __bshaffer__ package which we implemented in __codeigniter__ to __laravel passport__ so we used basic auth for password grant type but laravel passport use __oauth2-server__ package so we can’t use that perk anymore.

Anyway I will use middlewares and replace basic auth  authorization header with POST params before request reach Password grant endpoint. 

Tnx

alshf89:
Sorry, but spec say password grant must use basic auth too.

According to section 4.3.2 that talk about password grant access token request:

It says:

If the client type is confidential or the client was issued client credentials (or assigned other authentication requirements), the client MUST authenticate with the authorization server as described in [Section 3.2.1](https://tools.ietf.org/html/rfc6749#section-3.2.1).

In section 3.2.1:

Confidential clients or other clients issued client credentials MUST authenticate with the authorization server as described in [Section 2.3](https://tools.ietf.org/html/rfc6749#section-2.3.1) when making requests to the token endpoint.

In section 2.3.1:

Clients in possession of a client password MAY use the HTTP Basic authentication scheme as defined in [RFC2617] to authenticate with the authorization server.

The section I have posted in this [comment](https://github.com/thephpleague/oauth2-server/pull/1023#issuecomment-505475163) is exactly related to client credentials and user credentials topics in spec bcuz both of them follow rules in section 2.3.1.

If you don’t have plan to add this feature its OK. But its take my time to fix it and I really need it.

Anyway, Thanks for your time. 😉

Sephster:
Sorry but I don't think this is something we would support as I don't feel this is covered by the spec. Sorry I am not proceeding with this PR at this time. Thank you very much for your contribution though.

PR #1028, opened 2019-07-03
PSR-14 compliant version

datapp:


Sephster:
Thank you for your pull request and apologies for the delay in responding. At the moment, because PSR-14 is only supported in PHP 7.2, we would not implement this change.

I'm still unsure about the level of support in the PHP community for PSR-14, as many of the frameworks out there haven't officially adopted it. As such, I'm unclear of the demand for such a change. If we do include this in a later version, it is likely we will keep both the league events and the PSR-14 events, allows implementers to choose which event dispatch mechanism they want to use.

Sorry we aren't going to go ahead with this PR at this time and thank you for your efforts on this.

Issue #1034, opened 2019-07-22
Example client repository allows public clients when it shouldn't


matt-allan:
The client credentials grant should only be used by confidential clients:

> The client MUST authenticate with the authorization server as described in Section 3.2.1.

https://tools.ietf.org/html/rfc6749#section-4.4.2

The `ClientCredentialsGrant` class does not validate that the client is confidential, but instead calls `ClientRepository::validateClient`.

To prevent a public client from using the client credentials grant type you would need to check the `$grantType` and fail validation if it's client credentials and the client is public.

However, the example ClientRepository [does not do this](https://github.com/thephpleague/oauth2-server/blob/e1dc4d708c56fcfa205be4bb1862b6d525b4baac/examples/src/Repositories/ClientRepository.php#L54) and if you follow the example public clients would be allowed to use the client credentials grant type.

You can illustrate the problem by adding a public client to the example and making a curl request:

```diff
diff --git a/examples/src/Repositories/ClientRepository.php b/examples/src/Repositories/ClientRepository.php
index 47db508c..9339a971 100644
--- a/examples/src/Repositories/ClientRepository.php
+++ b/examples/src/Repositories/ClientRepository.php
@@ -43,6 +43,12 @@ class ClientRepository implements ClientRepositoryInterface
                 'redirect_uri'    => self::REDIRECT_URI,
                 'is_confidential' => true,
             ],
+            'mypublicapp' => [
+                'secret'          => null,
+                'name'            => 'My Other app',
+                'redirect_uri'    => self::REDIRECT_URI,
+                'is_confidential' => false,
+            ],
         ];
 
         // Check if client is registered
```

```
curl -X "POST" "http://localhost:4444/client_credentials.php/access_token" \
        -H "Content-Type: application/x-www-form-urlencoded" \
        -H "Accept: 1.0" \
        --data-urlencode "grant_type=client_credentials" \
        --data-urlencode "client_id=myothereapp" \ 
        --data-urlencode "scope=basic email"
{"token_type":"Bearer","expires_in":3600,"access_token":"..."}
```

Should the example be updated, or should we update the `ClientCredentialsGrant` to check `Client::isConfidential` so it's not necessary to check in `validateClient`?

matt-allan:
[...]

PR #1035, opened 2019-07-22
Prevent public clients from using the client_credentials grant type

matt-allan:
This PR updates the `ClientCredentialsGrant` to fail if the client is not confidential.

According to the OAuth specification the client credentials grant type should only be used with confidential clients.

> The client MUST authenticate with the authorization server as described in Section 3.2.1.

https://tools.ietf.org/html/rfc6749#section-4.4.2

While it's possible to check this in the `validateClient` method it's possible the user failed to do so. As explained in #1034 the example repository is not checking this and the docs don't tell you to check this either so it's likely other developers have missed this.

By checking it in the `ClientCredentialsGrant` we don't need to require the user to implement the check themselves.

I had to update a few tests that were relying on this behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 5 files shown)


Sephster:
Thanks @matt-allan - looks good. I made some minor adjustments to formatting and added a note in the changelog but otherwise is good to go. Thanks for submitting this!

PR #1037, opened 2019-07-22
Update the client interface docs

matt-allan:
This PR updates the docs for the client entity interface to correspond to the methods you need to implement as of the 8.0 release.

- Correct the getClientEntity method documentation
- Document the validateClient method


Sephster:
Thanks @matt-allan

Issue #993, opened 2019-02-05
Spec compliance: invalid_request returned when refresh token invalid


jacobweber:
According to [the OAuth2 spec](https://tools.ietf.org/html/rfc6749#section-5.2), when you're using the `refresh_token` grant and the refresh token has expired, it should return an `invalid_grant` error. But this server actually returns a `invalid_request` error.

 ```
        invalid_grant
               The provided authorization grant (e.g., authorization
               code, resource owner credentials) or refresh token is
               invalid, expired, revoked, does not match the redirection
               URI used in the authorization request, or was issued to
               another client.
```

Sephster:
Well spotted. Thanks for finding this issue @jacobweber

Sephster:
Fixed in PR #1042

PR #1042, opened 2019-07-26
change errorType to invalid_grant on invalid refresh_token

omdmhd:
Fix Spec compliance according to [this issue](https://github.com/thephpleague/oauth2-server/issues/993).
Now invalid_grant is returned instead of invalid_request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,3 +30,4 @@ after_script:
 branches:
   only:
     - master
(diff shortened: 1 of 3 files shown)


Sephster:
Thanks @omdmhd - Changing this to the version 9 branch as this is BC but otherwise looks good. Cheers!

PR #1043, opened 2019-08-13
Update which grant graphics

### LINKED ISSUE #1038: The 'Which Grant' docs are outdated and don't reflect best practices
The ['Which Grant?'](https://oauth2.thephpleague.com/authorization-server/which-grant/) page of the documentation is outdated and the recommendations don't match up with current best practices. Specifically:

- The password grant is recommended for first party clients, but the [OAuth 2.0 for Browser-Based Apps IETF draft recommends the authorization code grant](https://tools.ietf.org/html/draft-parecki-oauth-browser-based-apps-02#section-5).
- The implicit grant is recommended for third party clients but the [IETF recommends the authorization code flow](https://tools.ietf.org/html/draft-parecki-oauth-browser-based-apps-02#section-9.8.5).

I'm not sure if I can edit the flowchart without a Mac with Sketch installed. If not it might be a good idea to save a SVG that can be edited with Inkscape next time it's edited. :grinning:

matt-allan:
I finally got around to updating the 'which grant' flowchart.

which grant flowchart

The flowchart no longer recommends the password grant for 1st party apps because the [IETF doesn't recommend it](https://tools.ietf.org/html/draft-parecki-oauth-browser-based-apps-02#section-5). The IETF [doesn't recommend the implicit grant either](https://tools.ietf.org/html/draft-parecki-oauth-browser-based-apps-02#section-9.8.5).

This makes things a lot simpler because your only choices are really client credentials, auth code, or auth code with PKCE.

I opened the SVG from Sketch and modified it in Inkscape. I changed all of the text elements because the original diagram used Helvetica which is not a free font and the SVG renders poorly on systems that don't include it (on my Linux laptop it renders in serif).

I included the original SVG (so it can be edited with Inkscape) as well as a version that was minified with [SVGOMG](https://jakearchibald.github.io/svgomg/).

Closes #1038.


Sephster:
Thanks for this @matt-allan. I will take a look at this this evening but all looks good. I'm keen to retain the old documentation on the page in some capacity. Probably just leaving it there with a note saying that this is historical/no longer recommended.

The reason for this is that we won't be removing the old grants from the server so I'd like to have some documentation as to why it is there and leave it up to the implementer whether they want to use them or not, so have a kind of recommended/not recommended (historical) section.

Hope that makes sense. I will take a proper look at this this evening but thank you so much for making this change. It is really appreciated and I'm not sure how I'd have done this without access to a mac!

Sephster:
The RFC linked is still a work in progress at present but I'm no averse to updating the docs to reflect its recommendations. Reading through the docs, I think we should still have the password grant listed as per the quote below:

> To conform to this best practice, first-party applications using OAuth or OpenID Connect MUST use an OAuth Authorization Code flow as described later in this document or use the OAuth Password grant.
>
> It is strongly RECOMMENDED that applications use the Authorization Code flow over the Password grant for several reasons...

I'd be keen to recommend the auth code grant over the password grant but not dismiss the password grant entirely. We could update the text below to provide the reasons for this recommendation as per the RFC.

Happy to remove the implicit grant though as there really is no need to use it. As long as we have a section explaining the reason for not needing to use it, I think we are good.

I'd welcome your thoughts on these proposals. I am happy to put them in place once I remember how to use Inkscape :D Thanks for your time on this @matt-allan

Sephster:
I couldn't figure out a way to nicely integrate the password grant into the diagram so have left it as is. I've updated the docs to add in a bit about why we won't recommend the password grant or implicit grant in certain scenarios. 

Other than that, looks good to go. Thanks for updating this Matt. It is much appreciated.

Spomky:
Hi all,

Sorry I missed that conversation and I post here after the PR is merged.
Anyway, I think it is relevant to share another diagram with you.

I don't expect it to be included in the documentation. It is a more complete one that includes flows not supported by the library. However may be useful to have a wider view of possibilities offered by the OAuth2 framework. It also uses terms that or described in the specifications (e.g. Resource Owner instead, Client, End-User instead of Machine, Access Token Owner...).

Hope it helps.

GT

Issue #1007, opened 2019-05-09
Accept key content directly instead of require a key file


marc-mabe:
We are running the server in a docker environment with read-only filesystem and we use the same docker image for staging and production.

That's why we configure everything (including keys) using environment variables but this is impossible because the keys are required to be available as files :(

Furthermore you already accept the keys as content but in this case you write them to the filesystem (fails in my case because of r/o filesystem) just to get read again by `Lcobucci\JWT\Signer\Key`.

Interestingly `Lcobucci\JWT\Signer\Key` does accept the key content directly and only tries to read it as file if it starts with `file://`.

As a simple test I have updated `League\OAuth2\Server\CryptKey` constructor to just take the given key as is in case of matching `RSA_KEY_PATTERN` and it worked fine for me.
... It's already a bit misleading that `$keyPath` accepts key content (was the case before) but now `getKeyPath()` also returns the key content in this case.

I would suggest to:
* Change it the way I described it
  * rename constructor argument `$keyPath` to `$key` and document it accepts both, a file path and key conent
  * deprecate `getKeyPath`
  * add new method `getKey()` documented as either key content or file path

Thoughts?

sg3s:
We are also running a non-root containerized environment with read-only file systems. 
Our containers are also never rebuilt between promotions to different environments.
We worked around the issue by always specifying the keys as a file path reference. 

But I agree, the library should just take a full key and not try to save it as a file. It tries to coerce you to handle the key safely but the method it uses does not play well with containerized environments. Personally I would ditch all code trying to manage how the key is delivered to the library, maybe only accept strings as keys, not as file paths, and only validate if that string looks like an rsa key.

**What you can do to handle this without changing the library**
Depending on your container orchestration solution you should be able to mount files in the container somewhere during startup. You can then just reference/configure the file paths in your application. You should not need root privileges or write rights inside the container to mount these files.

This can be configured to work nicely with docker-compose and other local tooling too of course.

Sephster:
Thanks for the suggestion @marc-mabe. We've just released a new major version and as this would be a BC change, it would be a while before we'd get this into the system. 

As time goes by, I'm wondering more and more if we should have a .env file for the server for all these small options to make them easier to configure such as PKCE enablement, keys etc.

I will mark this as a future feature for the timebeing but don't think we can implement this just now. Thanks for the suggestion though. It's very much appreciated.

sg3s:
[...]

PR #1044, opened 2019-08-17
Abstract CryptKey public methods to the CryptKeyInterface

erickskrauch:
The library provides flexible storage for all possible types of entities but strictly adheres to the use of the RSA key written to the file for signing JWT tokens. Using RSA, the size of the signature is almost 3 times larger than the key itself, so I would prefer to use a different key and signature algorithm. But the current implementation of CryptKey sets a strict requirement for RSA.

Of course, I can extend from CryptKey and override the constructor, but this is a way of avoiding the problem rather than solving it.

This solution is BC, so I expect it to be available only in the 9.0.0 release. Anyway, I think we need a lot more abstraction on the keys (see https://github.com/thephpleague/oauth2-server/issues/1007).

### DIFF
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -101,11 +101,17 @@ class CryptKey implements CryptKeyInterface
         return 'file://' . $keyPath;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function getKeyPath()
     {
         return $this->keyPath;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function getPassPhrase()
     {
         return $this->passPhrase;
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -14,7 +14,7 @@ namespace League\OAuth2\Server;
 use LogicException;
 use RuntimeException;
 
-class CryptKey
+class CryptKey implements CryptKeyInterface
 {
     const RSA_KEY_PATTERN =
         '/^(-----BEGIN (RSA )?(PUBLIC|PRIVATE) KEY-----)\R.*(-----END (RSA )?(PUBLIC|PRIVATE) KEY-----)\R?$/s';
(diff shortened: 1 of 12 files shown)


Sephster:
Thanks @erickskrauch. Out of interest, which signing mechanism were you looking to use instead? Can you provide some background as to the issues you are having with the size of the file at the moment? We've chosen this signing mechanism to keep things secure by default but happy to look at changing the library if this is causing real problems for people

erickskrauch:
@Sephster, `ECDSA SHA-256`. This algorithm provides a much shorter signature length (compare them on https://jwt.io). You can say that this is a bit saving, but if it can be optimized, I prefer this algorithm over RSA.

Sephster:
Thanks @erickskrauch. I can see why someone would want this. ECDSA is slower than RSA but does offer some space saving so I guess the implementer would have to weigh up the benefits/drawbacks. Seems like a good idea to include something like this. I will give it a full review soon. Cheers

src/CryptKey.php
```diff
@@ -101,22 +101,12 @@ private function saveKeyToFile($key)
         return 'file://' . $keyPath;
     }
 
-    /**
-     * Retrieve key path.
-     *
-     * @return string
-     */
-    public function getKeyPath()
+    public function getKeyPath(): string
```
Sephster: Remove the type hint as we don't implement this in the package yet. I'd prefer to be consistent across the codebase.

src/CryptKey.php
```diff
@@ -101,22 +101,12 @@ private function saveKeyToFile($key)
...
         return $this->keyPath;
     }
 
-    /**
-     * Retrieve key pass phrase.
-     *
-     * @return null|string
-     */
-    public function getPassPhrase()
+    public function getPassPhrase(): ?string
```
Sephster: Same comment here regarding type hinting.

tests/AuthorizationServerTest.php
```diff
@@ -153,7 +152,7 @@ public function testMultipleRequestsGetDifferentResponseTypeInstances()
         $encryptionKey = 'file://' . __DIR__ . '/Stubs/public.key';
 
         $responseTypePrototype = new class extends BearerTokenResponse {
-            /* @return null|CryptKey */
+            /* @return null|\League\OAuth2\Server\CryptKeyInterface */
```
Sephster: I would prefer we add a use statement instead of specifying the FQDN.

PR #1048, opened 2019-08-22
Add getter for OAuthServerException::redirectUri property

erickskrauch:
In the current implementation, the only way to get the final `redirectUri` is to call the `generateHttpResponse` method and then read the `Location` header from the formed response. It is enough for a classic website, but it is inconvenient for SPA, which converts errors into an intermediate format and transfers it to the frontend.

This change abstracts logic of generation of the `redirectUri` into its own public getter.

### DIFF
--- a/src/Exception/OAuthServerException.php
+++ b/src/Exception/OAuthServerException.php
@@ -294,14 +294,9 @@ class OAuthServerException extends Exception
 
         $payload = $this->getPayload();
 
-        if ($this->redirectUri !== null) {
-            if ($useFragment === true) {
-                $this->redirectUri .= (strstr($this->redirectUri, '#') === false) ? '#' : '&';
-            } else {
-                $this->redirectUri .= (strstr($this->redirectUri, '?') === false) ? '?' : '&';
-            }
-
-            return $response->withStatus(302)->withHeader('Location', $this->redirectUri . http_build_query($payload));
+        $redirectUri = $this->getRedirectUri($useFragment);
+        if ($redirectUri !== null) {
+            return $response->withStatus(302)->withHeader('Location', $redirectUri);
         }
 
         foreach ($headers as $header => $content) {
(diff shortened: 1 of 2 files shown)


Sephster:
Hi @erickskrauch. Thanks for your submission. Can you provide some more details as to why it is difficult for you to retrieve the redirect URI from the location? This is a standard http header so most clients should be able to read this. 

I'm perhaps misunderstanding the issue here but this feels more like a quirk of a specific SPA implementation rather than a wider issue. Any additional information you can give (perhaps with code examples) would be much appreciated. 

Thank you

Sephster:
@erickskrauch I was wondering if you have had a chance to look at my comment? Any additional information you can give regarding this would be much appreciated. Thanks!

erickskrauch:
@Sephster, I'm sorry for the delay. It was such a coincidence that I went on vacation, both from work and from my own projects :)

--------------

First of all, I would like to point out that my project is written using the Yii2 framework, which doesn't use the PSR-7 on which the library relies. This generates a lot of converting code when I create intermediate objects for requests and responses, and then _deconvert_ them into a response object for the Yii2 framework.

I'm all for using the PSR standards, but encapsulating the results of the authorization flow in the response object doesn't seem right to me. It will be better to use some intermediate entity objects that contain grant results, as well as individual helpers that can wrap these results in PSR-compatible response objects.

This will ensure compatibility with any frameworks, whether or not based on PSR standards, as well as allow them to create their own request/response formats for the authorization processes on the OAuth-provider site (my case).

-------------

This change is an attempt of the first shift in this direction, when part of the private data, which otherwise would have to be extracted through the full response object, can be obtained directly by passing it to the frontend by any convenient end-developer method.

Sephster:
Sorry @erickskrauch but I feel this is a niche case so won't be progressing this PR. If other people come forward saying this is an issue for them and can provide a more concrete example, I will be happy to take another look.

Thank you for your time and sorry I'm not progressing this today.

PR #1051, opened 2019-09-12
Declare all abstract functions

Nyholm:
We are using this function but we've never declared it abstract.


Sephster:
Thanks for the kind words @Nyholm and thanks again for the contribution. It is very much appreciated. Apologies I didn't immediately understand the issue and thanks for your patience with this. Cheers!

PR #1057, opened 2019-10-02
Check if there is the auth_code_id key in auth code

### LINKED ISSUE #1050: AuthCodeGrant issues a 500 response if user sends a valid refresh token by mistake
This is because a refresh token doesn't have an `auth_code_id` in its payload and we check for this in the AuthCodeGrant. See issue https://github.com/laravel/passport/issues/1077 for further details.

While it is unlikely this scenario would occur, we should add a basic check for the presence of the `auth_code_id` so we can respond with an appropriate error message to aid developers that make this error.

kpn13:
Fix #1050

Just add a simple test to check if the `auth_code_id` property is present in the token payload.

Happy Hacktoberfest :)


Sephster:
Thanks @kpn13 - I will look at this this evening but on first pass, looks good!

PR #1066, opened 2019-10-31
Add oauth server

rannmann:
Fixes:

```
127.0.0.1:49826 [500]: /client_credentials.php/access_token - Interface 'League\OAuth2\Server\Repositories\ClientRepositoryInterface' not found in /var/www/html/thirdparty-integration-example/src/Repositories/ClientRepository.php on line 15
```

when running the first example curl request.

### DIFF
--- a/examples/composer.json
+++ b/examples/composer.json
@@ -1,6 +1,7 @@
 {
     "require": {
-        "slim/slim": "^3.0.0"
+        "slim/slim": "^3.0.0",
+        "league/oauth2-server": "^8.0"
     },
     "require-dev": {
         "league/event": "^2.2",


rannmann:
It's a little awkward requiring its own repo, but for the example to work as the docs say it needs to be included as far as I can tell.

Sephster:
Hi @rannmann. The examples are supposed to be run without changing the folder structure. It is expected that you just clone the whole repo and then `cd` into the _examples_ directory and follow the readme instructions.

If you do this, the examples should work. I just tested this out on my machine and all was fine. I will close this PR as I don't think there is an issue here but if you'd like to discuss further or I've made a mistake please feel free to respond and I can take a second look. Thank you for your submission anyways. Cheers.

rannmann:
Strange.  That's exactly what I did and it was missing.  Maybe you have to run it with optimized auto-loaders or something?  I mean, I see that it *should* work based on the composer file but it wasn't on my machine.

Sephster:
In the example you have given, the root of the main library would need to be in your `/var/www/html` folder. I thought you might have changed the directory structure because you had modified the _examples_ folder to be _thirdparty-integration-example_ instead.

When running the examples, I had not changed any composer settings so am a bit stumped. If you can replicate the issue, I'm happy to give this another look. Thanks for your efforts with this.

PR #1072, opened 2019-11-21
Update .gitattributes

reedy:
Remove duplicates and add a couple of extras


Sephster:
Thanks for the PR. Looks good to me. Cheers.

Issue #125, opened 2013-12-12
generating authCode


origal:
Hi,

I'm trying to implement something similar to whats described in here:
https://github.com/alexbilbie/oauth2-example-auth-server/blob/master/oauth.php

I understand this tutorial is somewhat deprecated, but I was wondering in the meanwhile if anybody can tell me how to generate the authCode nowdays?

I'm talking about this line:
`$authCode = $server->newAuthoriseRequest('user', $params['user_id'], $params);`
which is line 179. There isn't seems to be a `newAuthoriseRequest` method anymore.
I see this method on League\OAuth2\Server\Grant\AuthCode.php on line 187. How would I call it?
Everything else pretty much works with minor adjustments.

Many thanks

PR #1074, opened 2019-11-28
Device Authorization Grant

lucadegasperi:
Hello,
Since I needed the device authorization grant for a project I'm working on, I've decided to give a shot at implementing it in a presentable way.

For the most part I copied other grants code where needed and tried to keep the methods and interfaces as similar to what's already implemented as possible.

The interval checking between requests for an access token is currently not implemented, as I think this could be more easily delegated to a framework middleware than implement the logic right inside the package.

Pairing a user identifier with the device code is left outside the scope of this implementation as well.

I've created a laravel passport implementation based off this code that you can find here: https://github.com/lucadegasperi/passport/tree/device-flow

Let's make this grant happen!


lucadegasperi:
Thanks for the fixes @Sephster, do you think it's good enough to merge now?

Sephster:
Hi @lucadegasperi - sorry I still have a wee bit more to look at. Specifically around handling of pass checking and response rates. I think this probably does need including so have started to add these in. Thanks for your patience with this.

yovchev:
Hi, @Sephster I think the rate limiter must be Middleware implementation.

Let me know what you think I will try to get something mocked up soon.

yovchev:
@Sephster I have mocked up something that uses cache let me know if you will be ok with something like this in the examples composer I started setting up the Middleware and the Rate limiter but I need a way of caching between request. 

If you don't want to include other lib's I can use raw `file_put_contents` to cache between requests

https://github.com/lucadegasperi/oauth2-server/pull/1

yovchev:
@Sephster I was able to do the example without middleware please see a working example here 

https://github.com/lucadegasperi/oauth2-server/pull/1/commits/3865bace7768e85454398ff6d26cf2568a833867

yovchev:
I did some extra work on this here if you interested feel free to look and let me know if I can help to get this merged soon.

https://github.com/yovchev/oauth2-server/commits/device-flow

yovchev:
@Sephster Ignore my previous comments the middleware implementation don't make sense please check this as a possible candidate for handling polling rate https://github.com/yovchev/oauth2-server/commits/device-flow-grant

Sephster:
@lucadegasperi and @yovchev sorry for the radio silence on this ticket. I have some time coming up this week so will aim to push this through asap. Cheers for your patience.

yovchev:
No worries @Sephster I will be at hand if I can help with something.

Sephster:
@yovchev thanks for your time on this. I had a look yesterday and couldn't get the examples to work. I think this is because my system wasn't creating a cache folder as it didn't have permission to do so.

I'm thinking that it might be simpler to just do away with the cache system and instead rewrite the example with a ready made SQLite DB which would probably be easier to maintain. If we went down this route, I'd be ok with just pushing through the device code grant changes and sorting out the examples in a separate PR. 

I will have a look at the changes without the examples to see if we can progress them. Thanks for your patience

AlastairDewar:
@Sephster I appreciate it's a chaotic time for everyone at the moment. Is there anything that I can do to help get this merged?

Sephster:
Hi Ally. Thanks for offering to help. There isn't much that can be done at this time to be honest and I just need to find a significant block of time to finish this off as conscious it has been sitting here a good while now. The main issue is I don't want to merge prior to making sure it covers the spec properly. 

To do this I also want to rewrite the examples for Slim 4 so I can easily do some manual testing rather than just read code and tick off all of the points which is time consuming.

I'm hesitant to commit to a date as I have a baby in the house just now so my time isn't as available as it was a few months ago but hoping to get some time to work on this soon. I just can't commit to a firm date at present as don't want to disappoint. Thanks for your offer of help though and I will try my best to get this out the door sooner rather than later.

yovchev:
Hi, @Sephster congrats on the baby.

Let me know if this is the speck you want to make sure is covered [rfc8628](https://tools.ietf.org/html/rfc8628)
and I will do one run of tests on the latest pull request and reply here any findings.

Bloemendaal:
Hi!

I've got a question about the given example. The example uses `client_secret` for obtaining a `device_code` and obtaining an `access_token`. However, the standard states that a device should always be treated as a public client, see [RFC8628 Section 5.6](https://tools.ietf.org/html/rfc8628#section-5.6) and [RFC6749 Section 2.1](https://tools.ietf.org/html/rfc6749#section-2.1).

Is the `client_secret` optional?

Let's hope this feature gets implemented soon! 🙌

GreyXor:
Thank you, very useful! Can I use it in production or are there important things missing?

GreyXor:
@lucadegasperi Thank you, very useful! Can I use it in production or are there important things missing?

lucadegasperi:
@GreyXor I do use it in production, but I'm not sure if the spec has changed in the meanwhile. Probably @Sephster has a better understanding of the state of things.

GreyXor:
@lucadegasperi @Sephster thanks :) do you know if it can be merged soon ?

GreyXor:
@lucadegasperi @Sephster thanks :) do you know if it can be merged soon ?

ivanleskovar:
Hi all,

Nice work!

@Sephster when can we expect this can be merged?

Sephster:
Aiming for next major.

ivanleskovar:
@Sephster thx.
Any estimate when we can expect it?

modricl:
@Sephster do you have estimate date for next major?

Sephster:
Hopefully in the next few weeks. Testing underway now. If you're interested in this grant, I would welcome you giving it a spin

src/AuthorizationServer.php
```diff
@@ -132,6 +133,34 @@ public function completeAuthorizationRequest(
...
+            }
+        }
+
+        throw OAuthServerException::unsupportedGrantType();
+    }
+
+    /**
+     * Complete a device authorization request
+     */
+    public function completeDeviceAuthorizationRequest(string $deviceCode, string|int $userId): ResponseInterface
```
sniper7kills, line 157:
```php
public function completeDeviceAuthorizationRequest(string $deviceCode, string|int $userId, bool $approved): ResponseInterface
```

src/AuthorizationServer.php
```diff
@@ -132,6 +133,34 @@ public function completeAuthorizationRequest(
...
+    {
+        return $this->enabledGrantTypes['urn:ietf:params:oauth:grant-type:device_code']
+            ->completeDeviceAuthorizationRequest($deviceCode, $userId)
```
sniper7kills, line 160:
```php
->completeDeviceAuthorizationRequest($deviceCode, $userId, $approved)
```

src/AuthorizationServer.php
```diff
@@ -132,6 +133,34 @@ public function completeAuthorizationRequest(
...
+            ->generateHttpResponse($response);
+    }
```
sniper7kills, line 162: Testing this out w/ Laravel Passport; Getting the following error: Still a WIP; so unsure if actually a bug/issue.
```
ArgumentCountError: Too few arguments to function League\OAuth2\Server\Grant\DeviceCodeGrant::completeDeviceAuthorizationRequest(), 2 passed in /var/www/html/vendor/league/oauth2-server/src/AuthorizationServer.php on line 160 and exactly 3 expected in file /var/www/html/vendor/league/oauth2-server/src/Grant/DeviceCodeGrant.php on line 125
```

sniper7kills (changes requested):

src/AuthorizationServer.php
sniper7kills: Additionally; should it be returning a ResponseInterface; when in src/Grant/DeviceCodeGrant.php Line #125 returns void; or should that be returning a ResponseInterface instead of void?

sniper7kills:
@Sephster @lucadegasperi ,

Just finished taking a look and implementing this locally. Beyond the above few comments I was able to get @lucadegasperi passport implementation updated (https://github.com/RunbookSolutions/passport/tree/device-code) and got everything working using a python as the "device".

Great Work Guys!

src/AuthorizationServer.php
Sephster: Thanks for this suggestion @sniper7kills - I just went to implement it then and looks like I already came to the same conclusion. I'd added an additional argument, `$userApproved` which is pretty much what you suggested. Thank you for taking the time to look at this and thanks for the valuable input

src/AuthorizationServer.php
Sephster: Great suggestion and incorporated into the branch. Thank you

src/AuthorizationServer.php
Sephster: You are not the first person to have suggested this so I might change this. My thinking was that unlike the auth code grant, we aren't passing calls around and redirecting etc so we could just call the method directly. I think I might need to revisit this though. I'm going to get the first release candidate out today, then will look at how this works with Passport to get a level of confidence on the implementation, and perhaps change it if I run into problems. Cheers

Sephster (approved):

PR #1080, opened 2020-01-03
Wrong passwords are throwing "invalid_grant" error

cinco-de-mauro:
I noticed that my wrong passwords are now throwing an "invalid_grant" error. I noticed that at some point that at some point `throw OAuthServerException::invalidCredentials();` was changed to `throw OAuthServerException::invalidGrant()`.

I'm attempting to change it back, because I don't think an "invalid_grant" error is correct when the credentials are incorrect.

### DIFF
--- a/src/Grant/PasswordGrant.php
+++ b/src/Grant/PasswordGrant.php
@@ -104,7 +104,7 @@ class PasswordGrant extends AbstractGrant
         if ($user instanceof UserEntityInterface === false) {
             $this->getEmitter()->emit(new RequestEvent(RequestEvent::USER_AUTHENTICATION_FAILED, $request));
 
-            throw OAuthServerException::invalidGrant();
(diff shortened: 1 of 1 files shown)


cinco-de-mauro:
Could use some help getting this process correct. Would love to learn how to do this right. 

(Unless of course my update isn't wanted or incorrect)

Spomky:
The [RFC6749 section 5.2](https://tools.ietf.org/html/rfc6749#section-5.2) shows that in case of invalid credential the `invalid_grant` should be returned as error message.

For me this modification is not in line with the specification.

cinco-de-mauro:
Ahh... I see. That makes sense. Should have looked at the spec. 

Unfortunate name. I was confusing it with `unsupported_grant_type`.

Thanks!

PR #1081, opened 2020-01-21
Fix invalid annotation in AbstractGrant

minextu:
This was causing JMSTranslationBundle to not extract translations:

```[Semantical Error] The annotation "@string" in property League\OAuth2\Server\Grant\AbstractGrant::$defaultScope was never imported. Did you maybe forget to add a "use" statement for this annotation?```


Sephster:
Thanks for spotting. Merging in.

PR #1082, opened 2020-01-27
Return invalid_grant error when the authorization code is revoked

arietimmerman:
According to rfc6749, the correct error for trying to exchange a revoke (used) authorization code is `invalid_grant`, not `invalid_request`.

~~~
invalid_grant
    The provided authorization grant (e.g., authorization
    code, resource owner credentials) or refresh token is
    invalid, expired, revoked, does not match the redirection
    URI used in the authorization request, or was issued to
    another client.
~~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added support for PHP 7.4 (PR #1075)
 
 ### Fixed (v9)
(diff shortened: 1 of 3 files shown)


Sephster:
Thanks for this change @arietimmerman - would you be able to add a test to support this change? Thank you

arietimmerman:
Turned out I could enrich an existing test. So that's what I just did.

Sephster:
Looks great thanks @arietimmerman - I've merged this into branch 9.0.0 as this would be a breaking change but looks good to me. Thanks for your contribution

PR #1084, opened 2020-02-04
Expressive -> Mezzio

michalbundyra:
Zend Framework has been migrated to Laminas. Expressive changed name to Mezzio.
All zf components are abandoned, better refer to mezzio component now.


Sephster:
Thanks. Looks good to me.

PR #1096, opened 2020-02-25
Access tokens cannot be requested without a redirect_uri which according to spec they should

christiaan:
According to  [rfc6749 section-4.1.3](https://tools.ietf.org/html/rfc6749#section-4.1.3) the redirect_uri of a access_token request can be optional.
```text
redirect_uri
         REQUIRED, if the "redirect_uri" parameter was included in the
         authorization request as described in Section 4.1.1, and their
         values MUST be identical.
```

Currently this behavior is not possible with this library.
As soon as you create a `AuthorizationRequest` by calling `validateAuthorizationRequest` on the `AuthCodeGrant` it always contains a request_uri, even when the original request did not.

This is a problem because on a consecutive request to get an access_code the encrypted payload contains a redirect_uri making it required for that request.

All other code like `AuthCodeGrant#completeAuthorizationRequest` are able to handle AuthorizationRequests without a request_uri.

This is broken since version 7.0.0 before it did work.

### DIFF
--- a/src/Grant/AuthCodeGrant.php
+++ b/src/Grant/AuthCodeGrant.php
@@ -281,7 +281,7 @@ class AuthCodeGrant extends AbstractAuthorizeGrant
         $authorizationRequest = new AuthorizationRequest();
         $authorizationRequest->setGrantTypeId($this->getIdentifier());
         $authorizationRequest->setClient($client);
-        $authorizationRequest->setRedirectUri($redirectUri);
+        $authorizationRequest->setRedirectUri($this->getQueryStringParameter('redirect_uri', $request));
 
         if ($stateParameter !== null) {
             $authorizationRequest->setState($stateParameter);
--- a/tests/Grant/AuthCodeGrantTest.php
+++ b/tests/Grant/AuthCodeGrantTest.php
@@ -160,6 +160,48 @@ class AuthCodeGrantTest extends TestCase
         $this->assertInstanceOf(AuthorizationRequest::class, $grant->validateAuthorizationRequest($request));
     }
 
+    public function testValidateAuthorizationRequestWithoutRedirectUri()
+    {
+        $client = new ClientEntity();
+        $client->setRedirectUri('http://foo/bar');
+        $client->setConfidential();
+
+        $clientRepositoryMock = $this->getMockBuilder(ClientRepositoryInterface::class)->getMock();
+        $clientRepositoryMock->method('getClientEntity')->willReturn($client);
+
+        $scope = new ScopeEntity();
+        $scopeRepositoryMock = $this->getMockBuilder(ScopeRepositoryInterface::class)->getMock();
+        $scopeRepositoryMock->method('getScopeEntityByIdentifier')->willReturn($scope);
+
+        $grant = new AuthCodeGrant(
+            $this->getMockBuilder(AuthCodeRepositoryInterface::class)->getMock(),
+            $this->getMockBuilder(RefreshTokenRepositoryInterface::class)->getMock(),
+            new DateInterval('PT10M')
+        );
+        $grant->setClientRepository($clientRepositoryMock);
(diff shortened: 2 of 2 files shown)


christiaan:
Maybe I can illustrate the point by showing our Symfony2 implementation

```php
    public function authorize(Request $request): Response
    {
        $psrRequest = $this->httpMessageFactory->createRequest($request);
        $token = $this->tokenStorage->getToken();
        $user = $token->getUser();

        try {
            $authRequest = $this->authorizationServer->validateAuthorizationRequest($psrRequest);
            $authRequest->setUser(new User($user->getId()));
            $authRequest->setAuthorizationApproved(true);

            // This is needed as long as https://github.com/thephpleague/oauth2-server/pull/1096 is not merged
            if (!$request->query->has('redirect_uri')) {
                $authRequest->setRedirectUri(null);
            }

            $psrResponse = $this->authorizationServer->completeAuthorizationRequest($authRequest, new Psr7Response());
        } catch (OAuthServerException $exception) {
            $psrResponse = $exception->generateHttpResponse(new Psr7Response());
        }

        return $this->httpFoundationFactory->createResponse($psrResponse);
    }
```

Adding
```
            // This is needed as long as https://github.com/thephpleague/oauth2-server/pull/1096 is not merged
            if (!$request->query->has('redirect_uri')) {
                $authRequest->setRedirectUri(null);
            }
```
Was necessary because otherwise the response contains a redirect_uri in the encrypted payload.
And having the redirect_uri in the payload makes it impossible to request a access_token without a redirect_uri.

Sephster:
I think I'm following now. The redirect URI returned by the function `validateAuthorizationRequest()` is the pre-registered client redirect URI if you did not specify one on the initial request.

I don't think this is a problem. Could you outline how it is for you? If you don't specify a redirect URI, the one returned by `validateAuthorizationRequest()` is the one that would ultimately be used.

Thank you for your quick response.

christiaan:
> I don't think this is a problem. Could you outline how it is for you? If you don't specify a redirect URI, the one returned by `validateAuthorizationRequest()` is the one that would ultimately be used.

And that is completely fine, however it is also part of the authorizationRequest and it ends up in the encrypted payload that is in the 'code' query parameter that is added to the redirect_uri.

When the application that is redirected to receives the code parameter and wants to use that code parameter to do a access_token request with it MUST pass a redirect_uri in its request that matches the redirect_uri in the encrypted payload exactly.
Hence violating `if the "redirect_uri" parameter was included in the
         authorization request as described in Section 4.1.1, and their
         values MUST be identical`

I'm not passing a redirect_uri parameter in the authorization request but I DO have to pass it to the access_token request because of [AuthCodeGrant.php#L211](https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php#L211). It is currently impossible to get `empty($authCodePayload->redirect_uri) === false` to evaluate to true with a real authCodePayload generated by the authorization request.


To write a test for this would require an integration test as it spans two functionalities but there were none in the repository yet so I did not want to impose a integration test implementation.

christiaan:
Also this behavior change appears to be introduced in b3cd73cac7c788aa50d8a924ccfa0adbc9f1d5eb and appears to have been done unintentionally as I do not see any reference to this change in behavior in either the commit message or PR description.

denniscoorn:
I've come across this situation myself too and I think @christiaan has a valid point regarding this not being in accordance with the RFC.

@Sephster This PR contains a fix and also includes a test to prevent future regressions. Is there a reason why this can't be merged or is there something that needs to be added?

Sephster:
Yes I see where the issue is now and this is indeed a bug. I think this requires more than the fix here though. The logic needs refactoring as has become too complicated causing this issue to arise in the first place. I need to look at this more closely. Thank you @christiaan for taking the time to explain this thoroughly as I didn't notice the issue on first pass.

PR #1100, opened 2020-03-06
fix PasswordGrant::validateUser was throwing invalidGrant instead of invalidCredentials

marc-mabe:


Sephster:
Hi @marc-mabe. This was changed deliberately as the previous implementation didn't conform to the OAuth spec. Many thanks.

marc-mabe:
Hi @Sephster,

You are right, the server needs to return `error: invalid_grant` instead of `error: invalid_credentials` 👍 

But I don't think just returning the OAuth2 spec for `invalid_grant` as error description is helpful here and I don't think that the description has to be the same message described in the spec.

It would be much more helpful if the authentication server returns a description for each individual case as long as it doesn't leak internal information to attackers.

Spomky:
Hi @marc-mabe,

My 2cts on this discussion.
The main error from my point of view is that the OAuth2 is often viewed as an authentication protocol. It is not; it is an authorization one.

The user is never authenticated by the OAuth2 server. Only the client is (if confidential) and this is why the message `invalid_credentials` is returned in case of client authentication error.
The OAuth2 server just checks the grant data and returns `invalid_grant` if something is invalid.

So the behaviour of the library is compliant with the spec and follows the logic behind.
If the value of the `error` parameter should be `ìnvalid_grant`, there are other parameters that could be added to help the users/developers to understand what is going on: [`error_description` and `error_uri`](https://tools.ietf.org/html/rfc6749#section-4.1.2.1).
With these parameters, you can you can guide the users to resolve the error without sensitive data leakage.

I am not sure how to add these parameters for your project, but you should be able to set the message/uri you want. This is certainly the way to follow.

marc-mabe:
Hi @Spomky,

I totally agree with you. The server has to return `invalid_grant` and this is what it does (now).
And the ´error_description` / `error_uri` should be added to help the developer/user what happend.

Unfortunately I couldn't find a "simple" way to define a customized ´error_description` / `error_uri` without overwriting grant classes or hooking into the response.
-> Is there any other way to define custom ´error_description` / `error_uri`?

Furthermore, there is already an `error_description` returned with the message `The provided authorization grant (e.g., authorization code, resource owner credentials) or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.` which is the same test as the in RFC.

What I mean is that this message is very cryptic, even to developers, and so leaks to confusion and could be improved by providing more helpful error descriptions by default.

PR #1102, opened 2020-04-01
PasswordGrant.php, Fixing invalid credentials Exception

kakashispfc:


kakashispfc:
Can anyone help me with this CI error

Sephster:
I believe this was intentionally changed to conform to the OAuth specs. This is not a bug.  I will double check this later but I suspect this was changed for version 8. If this is the case, we would not merge in this PR. 

Thank you for your efforts and time looking at this though. I will update you later this evening but please feel free to check the spec in the meantime if you are able. Cheers.

kakashispfc:
Thanks for the feedback.
I use it in laravel/passport and in all my projects I have to extend the class to fix the exception of invalid credentials.

Sephster:
Hi @kakashispfc - sorry for the delay in checking this. I've had a look and this was changed intentionally to conform to the OAuth specs. The previous implementation did not. As such, we won't be going ahead with merging in this PR. Thank you for your time though. It is very much appreciated and sorry again for the delay in looking into this. Cheers

PR #1104, opened 2020-04-08
Bugfix: Adding validation to check if the request client

amauri-cloudbeds:
Adds a validation to check that the client obtained from the request params is the same that owns the session. 

The way it is, any valid API client that obtains access to a refresh token can grant itself access to the API and impersonate the API client for which that API session was created. Adding this validation prevents this behavior, as the refresh token grant will check that the client params sent on the request are the ones for the client that owns the session.

### DIFF
--- a/src/Grant/RefreshTokenGrant.php
+++ b/src/Grant/RefreshTokenGrant.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * OAuth 2.0 Refresh token grant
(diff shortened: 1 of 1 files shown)


Sephster:
I was confused about the session talk here until I realised this is for a very old version of this library. We no longer support this version and recommend upgrading to a later one instead of bug fixing the older versions.

Sorry that I'm not going to take this forwards.

Issue #1103, opened 2020-04-06
[Help needed] Adding authorization level to access-token


Richard87:
Hi!

I can add the required authentication security level to my AuthCode, since I know it at the time, but how can I transfer that knowledge to the AccessToken (either in the database, or as a claim in the JWT)

I'm already using OpenID Connect, but I'm not sure how to transfer that "knowledge" properly...
I need to know if the user is authenticated with Email/Password, 2-FA with a SMS-code, or 2FA from an external auth provider (external OIDC service)

Relevant other issues I found:
https://github.com/thephpleague/oauth2-server/issues/885
https://github.com/thephpleague/oauth2-server/issues/442
https://github.com/thephpleague/oauth2-server/issues/672

Richard87:
Ps, I don't need to add it to the JWT, but I do

Richard87:
My hacky solution...

1. Add a function "getAuthCode" in `AuthCodeRepositoryInterface`, so we can fetch it later
2. Modify the access token
 - Fetch the request token first, so we can access the encyrpted AuthCode
3. Modify AccessToken, to include our new claims in `__toString()`

First get a hold of the request object
```php
class AuthCodeGrantWithProvider extends AuthCodeGrant
{

    private ?ServerRequestInterface $request = null;

    public function respondToAccessTokenRequest(ServerRequestInterface $request, ResponseTypeInterface $responseType, DateInterval $accessTokenTTL)
    {
        $this->request = $request;
        return parent::respondToAccessTokenRequest($request, $responseType, $accessTokenTTL); // TODO: Change the autogenerated stub
    }

    protected function issueAccessToken(DateInterval $accessTokenTTL, ClientEntityInterface $client, $userIdentifier, array $scopes = [])
    {
        /** @var TokenRepository $authCodeRepo */
        $authCodeRepo = $this->accessTokenRepository;

        $encryptedAuthCode = $this->getRequestParameter('code', $this->request, null);
        $authCodePayload = json_decode($this->decrypt($encryptedAuthCode), false, 512, JSON_THROW_ON_ERROR);

        /** @var AuthCode $authCode */
        $authCode = $authCodeRepo->getAuthCode($authCodePayload->auth_code_id);

        /** @var AccessToken $accessToken */
        $accessToken = parent::issueAccessToken($accessTokenTTL, $client, $userIdentifier, $scopes);
        $accessToken->setAuthProvider($authCode->getAuthProvider());
[...]

Issue #672, opened 2016-10-17
Giving the ScopeRepository a reference to the AuthCode during AuthCode grant.


Shkeats:
It would be useful if the ScopeRepositoryInterface finalizeScopes() method took a nullable param of $authCodeIdentifier so that we can use contextual information stored on the authCode object to make decisions about which scopes a user can get. The AuthCode Grant could pass in it $authCodePayload->auth_code_id to this.

In my case this would be helpful as I need to persist some information about the state of the users session at the auth server when the auth code is issued and use this to determine if they get a certain scope when their token is issued. Obviously as the auth_code->access_token exchange is server to server the user's session is no longer present at the point the scopes are determined. The AuthCode object seems a natural place to persist this.

Let me know your thoughts or if I'm missing an obvious solution to my problem.

alexbilbie:
I believe the [`finalizeScopes`](http://oauth2.thephpleague.com/scope-repository-interface/) does what you require?

Shkeats:
@alexbilbie as far as I can tell there's currently no way to get a reference to the authcode inside finalize scopes?

alexbilbie:
`finalizeScopes` tells you the user ID, the scopes that were requested, the client and the grant type - what additional information do you need to make a decision?

Shkeats:
@alexbilbie It would be useful to have the AuthCode's id to access information persisted alongside the AuthCode when the user was present. In my case, details about the state of the user's session at the point in time the AuthCode was made.

Currently what I'm doing is using the user_id, client_id and scopes to grab the newest auth code that matches those criteria from the table, but this isn't always guaranteed to be the same one.

alexbilbie:
Okay; this will have to wait until V6 as it will require an interface change 👍

Shkeats:
On a related note I think it would also be helpful to give the access token repository a reference to the auth code so that we can persist the relationship between the two without needing to create a custom Grant type.

Richard87:
@alexbilbie Sorry for bringing this upp, 4! years later...

I have a similar problem, I need to know the security level the user authenticated with (username/password, username/password+otp or high-security external oidc-provider).

This information is available when authorizing the user, and stored in the AuthCode, but it's not available when I produce the AccessToken or the IdToken. I don't care if the security level is stored in the jwt, or in the entity/database, but I need the information when authorizing a request using the provided access-token (or IdToken)

Richard87:
Is this something you would accept a pull-request to fix?

[...]

PR #1105, opened 2020-04-10
Add previous authorized token information to getNewToken()

Richard87:
Solves #1103 and #672 and possibly #885


Sephster:
@Richard87 this doesn't look like it was branched off the 9.0.0-WIP branch. Because this is an interface change it will need to be in the next version. Would you mind resubmitting once branched off the v9 branch please? Thank you for your efforts

Richard87:
Yup, will do!

(You might want to update the Contributing guidelines that sais all PRs should be based off of master :)

Issue #962, opened 2018-11-08
Support nonce for Authorization Code and Implicit Grant


m0ark:
Currently I have to integrate full [OpenId Connect Core](https://openid.net/specs/openid-connect-core-1_0.html) spec, so I came across https://github.com/steverhoades/oauth2-openid-connect-server to this library. 

OIDC Implicit Flow requires a `nonce` to be included in the id_token issued. (see [3.2.2.1.](https://openid.net/specs/openid-connect-core-1_0.html))

In OIDC Authorization Code Grant the request can contain a nonce and in this case it has to be included in the id_token aswell (see [3.1.2.1.](https://openid.net/specs/openid-connect-core-1_0.html)).

For OAuth2 handling `nonce`s is only ever mentioned in a security advisory (mentioned in #634), but for OpenId Connect handling is mandatory. Since OpenId Connect is just an extension of OAuth2 i think `nonce`s should be handled by the latter.

Sephster:
We aren't supporting OpenId Connect at present but do plan to do so in the future. I don't believe including a nonce is required in any of the RFC's we currently support so don't think this will be added in until full support for OpenId Connect is.

MichaelKubovic:
Hi @m0ark, how did you deal with the missing nonce?

lordrhodos:
@m0ark I am interested in this as well. Hit the same issue.

ThaDaVos:
Any ETA when the official OpenID connect is coming? I am running into the same issue that the `nonce` is missing - figured out one can easily add it inside https://github.com/steverhoades/oauth2-openid-connect-server `IdTokenResponse` using a `->withClaim('nonce', '<VALUE>')` inside the overriden `getBuilder` method - but struggling to get the `nonce` all the way through so it's accessible there

Sephster:
No plans to add into the next major so likely after that. Steve's repo is the best option for now 👍

ThaDaVos:
Sadly it doesn't handle the `nonce` part (I am already using it) - but trying to make it spec compliant - any hint on how one uses the new `AuthorizationRequestInterface` introduced with https://github.com/thephpleague/oauth2-server/pull/1110 ?

I can't find any docs of it but seems to be able to solve the issue

ThaDaVos:
Nvm... V9 isn't even out yet....

ThaDaVos:
I added a comment in https://github.com/steverhoades/oauth2-openid-connect-server/issues/47 - how I did it today succesfully

PR #1110, opened 2020-04-16
Introduce AuthorizationRequestInterface

lordrhodos:
This PR introduces a new `AuthorizationRequestInterface` which is implemented by the current `AuthorizationRequest`. This way implementors are provided with a more flexible way to override certain parts of the authorization grant types than with the concrete class alone.

In my case I was looking into a way to use a decorated authorization request to be able to add the `nonce` attribute used of an OpenID Connect (OICD) implementation. I understand that this library is only supporting the OAuth2 specs and nothing related to OIDC (see https://github.com/thephpleague/oauth2-server/issues/962) which is totally fine with me. 

This PR raises the flexibility of the library without doing any harm 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added (v9)
 - A CryptKeyInterface to allow developers to change the CryptKey implementation with greater ease (PR #1044)
 - The authorization server can now finalize scopes when a client uses a refresh token (PR #1094)
+- An AuthorizationRequestInterface to make it easier to extend the AuthorizationRequest (PR #1110)
 
 ### Added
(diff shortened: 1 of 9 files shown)


lordrhodos:
related to https://github.com/thephpleague/oauth2-server/pull/1003

Sephster:
Looks good to me @lordrhodos although this is a breaking change so will need to go in version 9.

lordrhodos:
Thank you @Sephster 👍

iNviNho:
Looks good :)

shineability:
@lordrhodos Would you be willing to share how you fixed the `nonce` issue using the decorator? I solved it in a very hacky way by saving the `nonce` on the `User` when authenticating and then retrieve the value from the `User` again when generating the ID token in the response. I could use a more elegant solution :)

PR #1113, opened 2020-04-20
Decouple error responses from exception

lordrhodos:
This PR decouples the generation of the error responses from the OAuthServerException by moving the function to a new ExceptionResponseHandlerInterface with a default ExceptionResponseHandler implementation. 

The handler gets injected into the authorization and resource server and is used via a `::generateHttpRepsones` proxy method. 

This way required changes are minized while open up the library to support translated error response payloads in a second step.

### DIFF
--- a/examples/public/auth_code.php
+++ b/examples/public/auth_code.php
@@ -59,7 +59,6 @@ $app = new App([
         return $server;
     },
 ]);
-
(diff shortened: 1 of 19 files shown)


strongholdmedia:
I find this idea an awesome one.

PR #1121, opened 2020-06-04
abstract out the refresh token payload so that it can be customizable

Misosooup:
Would it be possible to make the refresh token payload a little more extensible?


Sephster:
Sorry @Misosooup I somehow missed this PR. I think that the addition of the nonce would be a breaking change and something that isn't usually required for OAuth 2 setups so we would have been unlikely to progress with this PR.

Thank you very much for your PR though and sorry we wouldn't be moving forwards with it this time.

Misosooup:
No worries. I was hoping to get something up to support OpenId (hybrid flow). Is it something that is being develop actively at the moment?

Issue #1120, opened 2020-06-03
Usage of Custom Claims


skroczek:
First of all thank you for this great library and for all the work that went into it. And of course congratulations to @Sephster on the new human fork. :partying_face: 

So what I'm currently trying to implement are custom claims.  What I actually want is relatively simple. I want an equivalent to this:

```
<php
// File: src/Entities/Traits/AccessTokenTrait.php
// [...]
trait AccessTokenTrait
{
// [...]
    /**
     * Generate a JWT from the access token
     *
     * @param CryptKey $privateKey
     *
     * @return Token
     */
    private function convertToJWT(CryptKey $privateKey)
    {
        return (new Builder())
            ->permittedFor($this->getClient()->getIdentifier())
            ->identifiedBy($this->getIdentifier())
            ->issuedAt(\time())
            ->canOnlyBeUsedAfter(\time())
            ->expiresAt($this->getExpiryDateTime()->getTimestamp())
            ->relatedTo((string) $this->getUserIdentifier())
            ->withClaim('scopes', $this->getScopes())

             // I want to add addiditonal information here:
             ->withClaim('custom', 'Custome DATA')

            ->getToken(new Sha256(), new Key($privateKey->getKeyPath(), $privateKey->getPassPhrase()));
    }
// [...]
}
```

-> I would like to add additional information to the JWT.
However, I don't see how I can do this without implementing a custom Grant, with TokenRepository etc. I'm aware that there are a few PR's out there on this topic, but none of them really seem to solve this. So my question is, is there an easier way to handle this?

Sephster:
Hi @skroczek - thank you so much for the well wishes. I think I'm over the worst of the sleepless nights now. Thanks for your kind words regarding the package.

I've seen in Passport that there are some calls to add custom claims to the JWT so this is something we definitely want to look at. You are right though, that unfortunately at this stage, there is no way to do this out of the box. It would require a custom BearerTokenResponse and AccessTokenTrait in order to achieve this.

You can add custom fields to the response but not the JWT which is what a lot of people are looking for. It is something I would like to get into the library in due course though. Unfortunately, the only way to do it at this time is to write your own implementations. Sorry this probably isn't the answer you are looking for.

skroczek:
Thanks for your quick response. Of course, it would be nice if it would already work, but I was already expecting this answer. It rather makes me feel better that I didn't miss anything obvious and it is also a chance to contribute to this library
[...]

PR #1122, opened 2020-06-04
Private claims implementation

skroczek:
This PullRequest should implement private claims as described in RFC7519 in section 4.2. According to #1120 , this is still missing, but is wanted.
This implementation is based on that of scopes, which is basically also only a (registered) claim.
I don't think the implementation is finished yet. There are a few questions open ~~and the testing is still missing completely~~. But the code works and can already be used for evaluation.
One open question is whether the claims should be checked for registered claims before the token is generated. If so, we would have to think about how to react in case of a violation. Will the claim simply be ignored, or should an exception be thrown?
~~About the testing: I have already looked at the test suite, but I'm not quite sure where and how to put it there.~~ A suggestion would be very welcome.

In general I would be very happy about feedback.


src/Entities/ClaimEntityInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * @author      Sebastian Kroczek <me@xbug.de>
+ * @copyright   Copyright (c) Sebastian Kroczek
+ * @license     http://mit-license.org/
+ *
+ * @link        https://github.com/thephpleague/oauth2-server
+ */
+
+namespace League\OAuth2\Server\Entities;
+
+use JsonSerializable;
```
Pchelolo, line 12: I couldn't find a reason for this to extend JsonSerializable? I couldn't find any.

src/Entities/ClaimEntityInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+interface ClaimEntityInterface extends JsonSerializable
```
Pchelolo, line 14: It would be neat to provide a ClaimEntityTrait, similar to TokenEntityTrait to simplify implementation.

tests/Stubs/ClaimEntity.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        return $this->name;
+    }
+
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    public function jsonSerialize()
```
Pchelolo, line 29: If it's needed (see comment above), $this->value being mixed is not guaranteed to be jsonSerializable itself

src/Repositories/ClaimRepositoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Returns claims
+     *
+     *
+     * @return ClaimEntityInterface[]
+     */
+    public function getClaims(
+        array $claims,
+        $grantType,
```
Pchelolo, line 28: grantType will be a string, can typehint

src/Repositories/ClaimRepositoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Claim repository interface.
+ */
+interface ClaimRepositoryInterface extends RepositoryInterface
+{
+    /**
+     * Returns claims
+     *
+     *
+     * @return ClaimEntityInterface[]
```
Pchelolo, line 24: params docs are missing

src/Repositories/ClaimRepositoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public function getClaims(
+        array $claims,
```
Pchelolo, line 27: What is the idea behind passing an array $claims in? AFAIK it's not needed

tests/Grant/AbstractGrantTest.php
```diff
@@ -360,7 +361,8 @@ public function testIssueAccessToken()
             new DateInterval('PT1H'),
             new ClientEntity(),
             123,
-            [new ScopeEntity()]
+            [new ScopeEntity()],
+            [new ClaimEntity('private', 'claim')]
```
Pchelolo, line 365: Would be good to have tests for non-string claim values as well

tests/Grant/ClientCredentialsGrantTest.php
```diff
@@ -43,10 +44,14 @@ public function testRespondToRequest()
         $scopeRepositoryMock->method('getScopeEntityByIdentifier')->willReturn($scope);
         $scopeRepositoryMock->method('finalizeScopes')->willReturnArgument(0);
 
+        $claimRepositoryMock = $this->getMockBuilder(ClaimRepositoryInterface::class)->getMock();
+        $claimRepositoryMock->method('getClaims')->willReturn([]);
```
Pchelolo, line 48: Would be good to have a test where $claimRepo is returning something

tests/ResponseTypes/BearerResponseTypeTest.php
```diff
@@ -32,11 +33,14 @@ public function testGenerateHttpResponse()
...
         $scope->setIdentifier('basic');
 
+        $claim = new ClaimEntity('_private', 'claim');
+
         $accessToken = new AccessTokenEntity();
         $accessToken->setIdentifier('abcdef');
         $accessToken->setExpiryDateTime((new DateTimeImmutable())->add(new DateInterval('PT1H')));
         $accessToken->setClient($client);
         $accessToken->addScope($scope);
+        $accessToken->addClaim($claim);
```
Pchelolo, line 43: Would be nice to have an assertion downstream that the claim has been set

tests/Stubs/ClaimEntity.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+
+namespace LeagueTests\Stubs;
+
+use League\OAuth2\Server\Entities\ClaimEntityInterface;
+
+class ClaimEntity implements ClaimEntityInterface
```
Pchelolo, line 8: In case you had a ClaimEntityTrait this would be much simpler, like other stub objects

tests/Grant/AbstractGrantTest.php
Pchelolo: Would be good to have an assertion verifying the claim was set.

Pchelolo:
I have left a few comments inline for the things that popped out. We have not yet fully tested this implementation, will report the results as we go, but this looks great overall. 

I have been looking into standards around this, and haven't found much: OAuth spec doesn't specify what would be used as a token, thus it doesn't contain any requirements or restrictions regarding the claims we can add into JWT. It seems logical that the library itself wouldn't impose any additional restrictions, so in my opinion the ClaimEntityInterface is rightfully very generic. 

I've surveyed a few other popular oauth2 server frameworks that support to adding custom claims within OAuth:
- Spring Security: allows to register a `TokenEnhancer` interface with the authentication manager. The enhancer receives the token object and a context-like Authentication object and can set additional information to the token object. [Example](https://sultanov.dev/blog/custom-jwt-claims-in-spring-security-oauth/). That is much more generic then what's implemented in the pull request, however this library encapsulates more logic than spring does, so I think following more generic approach taken by spring would be an overkill here. 
- nodejs [oauth2-server](https://oauth2-server.readthedocs.io/en/latest/index.html) - leaves generating the access token entirely to the library client, does not care about the access token semantics, treating it as an opaque string. Not viable approach here either
- python [OAuthLib](https://oauthlib.readthedocs.io/en/latest/index.html) - allows to provide a custom token generator, treating the token as an opaque string. alternatively has JWT-based implementation, that allows setting custom claims and is fairly similar to what's implemented here.

To sum up the approach taken in this PR seems reasonable to me, is on par with the capabilities of other libraries and does not violate any existing standards or RFCs as far as I could see.

src/Entities/ClaimEntityInterface.php
skroczek: You are right, this is really not needed since the builder takes care of it. I guess I just took it from the scope implementation. Nice catch.

src/Repositories/ClaimRepositoryInterface.php
skroczek: Right, I'll add that.

src/Repositories/ClaimRepositoryInterface.php
skroczek: Good point, again the Scope implementation was used as a prototype. However, it seems that I did not use the right name, it should be finalizeClaims. Also the comment should be adjusted accordingly. Anyway, the point of the method is that the claims can be validated and optionally more can be added or removed. This is especially useful when working with listeners. If there are no further concerns, I would simply rename the method and add an usefull comment accordingly.

skroczek:
Thank you very much for the close look and for the feedback. I have already answered to a few points. The suggestions all make sense to me, so I will implement them over the next days.

Pchelolo:
@skroczek do you need any assistance implementing/testing your changes?

tests/Grant/AbstractGrantTest.php
skroczek: This took some time, but this is not the right place to test values of the AccessToken. In this test the AccessToken is returned by a mock as a fixed return value. Since the claim parameter for the method AbstractGrant::issueAccessToken is optional, I will replace it with an empty array to match the value of the returned AccessToken.

tests/ResponseTypes/BearerResponseTypeTest.php
skroczek: As in the ClientCredentialsGrantTest, the AccessToken is returned by a mock. Therefore it is not possible to test whether claims are set here, at least not without fundamentally rewriting the test.

skroczek:
@Pchelolo It took me a while to find time to take care of it. But I think I've implemented all your suggestions except for the two tests. Thanks again for your suggestions and your patience. If you notice anything else, feel free to bring it up.

Pchelolo:
Thanks a lot and sorry that some of the suggestions didn't make sense. The CI seem to be failing. We will test the new version out and I'll share the results

src/Repositories/AccessTokenRepositoryInterface.php
```diff
@@ -24,11 +25,17 @@ interface AccessTokenRepositoryInterface extends RepositoryInterface
...
+     * @param ClaimEntityInterface[] $claims
      *
      * @return AccessTokenEntityInterface
      */
-    public function getNewToken(ClientEntityInterface $clientEntity, array $scopes, $userIdentifier = null);
+    public function getNewToken(
+        ClientEntityInterface $clientEntity,
+        array $scopes,
+        $userIdentifier = null,
+        array $claims = []
```
Pchelolo, line 37: Adding this new parameter to the interface method makes it incompatible with older library versions, php throws a fatal error when using old code with the new library due to incompatible method declaration, and this is the only backwards-incompatible change I could find. 

I'm not sure what's the best way to mitigate the problem. Some options:
1. Not changing the interface and making the client code responsible for adding the claims via addClaim(). In the next major version of the library the interface can be updated to have this extra parameter
2. Making this change and releasing a new major version of the library with this change.

I'm not a maintainer of the library, so asking @Sephster for judgement.

Pchelolo:
We've tested this implementation more and it seems to work well. I've left a comment about backwards compatibility inline, otherwise this LGTM. I'm not a maintainer of the library, so I would ask @Sephster to kindly look at it.

src/Repositories/AccessTokenRepositoryInterface.php
skroczek: Good point, my intention was actually to avoid backward incompatibilities. However, I agree with you that it boils down to the possibilities you mentioned. Basically it is up to @Sephster to decide if it is OK for him to create a new major version for the pull request or if I should adapt this change to be able to use it in a minor version. For me both solutions are fine, but I think that a minor version could be released sooner.

Pchelolo: ping @Sephster

Sephster: Thank you both for your work on this so far. I have a bit of time off work so will be aiming to review this myself in due course but the work you have both done has been invaluable.

I'd prefer not to release a major version for this change so I think it would be preferable to avoid any BC changes. If we want to keep the implementation as is, I'd merge it into next major branch but I don't think it would be released for some time to come yet.

I would lean towards going with the `addClaim()` method proposal instead if possible. Thank you

skroczek: I have adapted the arguments of the method, but I noticed something.Because the pull-request also changes interfaces, I wonder if it is even possible not to have back incompatibilities. This is not directly noticeable, because the interfaces are implemented by traits, which I have adapted accordingly and therefore no errors occur. However, these do not have to be used, which means that you might have to change your own implementation.
Depending on how it is handled in this project, we might have to wait for a major release. Although I would love to see the pull request accepted soon.

src/Entities/Traits/AccessTokenTrait.php
```diff
@@ -42,13 +43,20 @@ public function setPrivateKey(CryptKey $privateKey)
...
             ->expiresAt($this->getExpiryDateTime()->getTimestamp())
-            ->relatedTo((string) $this->getUserIdentifier())
+            ->relatedTo((string) $this->getUserIdentifier());
+
+        foreach ($this->getClaims() as $claim) {
+            $builder->withClaim($claim->getName(), $claim->getValue());
+        }
+
+        return $builder
+            // Set scope claim late to prevent it from being overridden.
```
Sephster, line 59: Is this correct? I think an exception is thrown if we try to register the same claim twice.

Sephster (commented):
I've added in some minor comments. If these could be addressed I'm happy to merge this in. 

Unfortunately this will be a break change release due to the changes to the interfaces to add the `getClaims`, `addClaim`, and `setClaimRepository` methods so will need to be in a major release.

If you are happy for me to do so, I will merge this into the 9.0.0-WIP branch instead of the master branch. Thanks for all of your work on this and do let me know if you have any comments

src/Entities/Traits/AccessTokenTrait.php
skroczek: Yeah, I think that's correct. I took another look at the code of the library, and the builder does not check if a claim has already been added. At least I couldn't find anything.

skroczek:
I already expected, after the last review, that the pull request will not make it into the current master branch. I noticed the breaking changes quite late. I have implemented your suggested changes so far. Anyway, I agree of course, if it is merged into the 9.0.0-WIP branch.  

Currently there seems to be a problem with styleci, but unfortunately I couldn't find out why the check fails. As far as I have seen, no changes have been made that could cause the check to fail. However, this is also the first time I am working with this service.

I will be poorly reachable in the next two weeks, so it may take some time before I can respond.

Sephster:
Thanks @skroczek for the changes. If you are happy to, I can pick up any remaining issues to push this through to the 9.x branch? Will aim to do that tomorrow

skroczek:
Hi @Sephster, if you could take care of the rest of these issues, I would be very grateful. Thank you very much for your support.

Pchelolo:
Hello, anything we can do to help getting this in quicker?

src/Entities/Traits/AccessTokenTrait.php
Sephster: Cool - thanks for checking.

src/Entities/ClaimEntityInterface.php
Sephster: Added so resolving comment.

tests/Stubs/ClaimEntity.php
Sephster: Function has been removed. Marking as resolved.

src/Repositories/ClaimRepositoryInterface.php
Sephster: Added. Marking as resolved.

src/Repositories/ClaimRepositoryInterface.php
Sephster: Removed so marking as resolved.

src/Entities/TokenInterface.php
```diff
@@ -82,4 +82,18 @@ public function addScope(ScopeEntityInterface $scope);
      * @return ScopeEntityInterface[]
      */
     public function getScopes();
+
+    /**
+     * Associate a claim with the token.
```
Sephster, line 87: I'm going to move these functions out of this interface. Claims aren't part of the OAuth spec. To keep this library as flexible as possible, we've kept the JWT specific implementation out of the interfaces so that if you don't want to use it, you don't have to. 

These functions would be better placed in the trait where we keep a similar JWT specific function, `convertToJWT`.  I will move this there along with the `getClaims()` function. Thanks

skroczek: I am totally fine with that. Thank you very much.

garethellis36:
Hello folks. Is this PR likely to be merged soon? I think it would help me with a problem I am facing.

Sephster:
No timescale on this I'm afraid. It will be done as soon as I get time but there is a lot to check.

Starfox64:
Sorry for bumping this but is this PR still alive?

I was hoping this could be merged for or before 9.00. I'm happy to help if required.

I'm asking to know if I should wait on an official implementation or just roll out my own like many have done by overriding the library.

skroczek:
We are now using zitadel as our IDP for various reasons. We have also switched from PHP to Go as our main language. Because of this, I unfortunately have neither much time nor interest to take care of it anymore. Nevertheless, this remains an excellent library. I will of course leave the PR open so that it can be merged with the repository if desired.

fredsal:
Hi, is this PR likely to be merged soon?

parallels999:
> No timescale on this I'm afraid.

@Sephster still no timescale?

MJTheOne:
Would also like to know why this is "ignored"? Custom claims are perfectly normal

parallels999:
@Sephster is possible merging or rebasing this with master?

systemsolutionweb:
@Sephster hi, we would love to have this feature
any news?

This is the PR with the most comments

Thanks

src/Grant/AuthCodeGrant.php
```diff
@@ -161,8 +161,18 @@ public function respondToAccessTokenRequest(
             }
         }
 
+        $privateClaims = [];
+
+        if ($this->claimRepository !== null) {
```
MJTheOne, line 166: First of all, sorry for the (very) late comment. But just got an email that people are interested again.
Maybe just personal preference, but checking if something is not null is not actually checking anything (because it can be ANYTHING at this point).
I'd rather check if it is what I want it to be, e.g. an instanceof check.

ssw1cblarrion:
Any news?

PR #1123, opened 2020-06-15
Return Redirect URI in OAuthServerException

StephanMeijer:


StephanMeijer:
Fix needed, know doing workaround:

```
       $reflectionProperty = new ReflectionProperty(OAuthServerException::class, 'redirect_uri');
        $reflectionProperty->setAccessible(true);
        
        $uri = $reflectionProperty->getValue($e);
```

Sephster:
Can you give a bit of detail about your scenario? Which grant do you need this for and in what context? Cheers!

Sephster:
@StephanMeijer do you have some time to provide a bit of context for this change? Thanks for your help.

StephanMeijer:
Yes @Sephster I will respond ASAP

StephanMeijer:
@Sephster I used this to translate `OAuthServerException` to my custom class: `MedMijServerException`, which allows me to report and render the error using Laravel error handling.

Sephster:
Thanks for the reply @StephanMeijer - could you detail why you needed to use a different exception class? Apologies for all the questions, I just want to understand what the current implementation is restricting you from doing. Thanks for your help

StephanMeijer:
The use case I approached was a bit more complex. I had to create an OAuth service which does not strictly implement OAuth but was more stricter on specs from the government.

So I had to create custom error requests as well, where redirects were required.

Sephster:
Thanks @StephanMeijer - can't think of a reason not to expose this so have merged. Cheers

StephanMeijer:
Thanks for your help

PR #1124, opened 2020-06-24
Use InvalidCredentials instead invalidGrant when user not found

maguilar92:
On login using password gran client with invalid credentials it was returning invalid grant instead invalid credentials.

I am using Laravel Passport with invalid email and/or password.

```php
$http = new \GuzzleHttp\Client([
    'verify' => !in_array(env('APP_ENV'), ['local', 'testing']),
]);
$response = $http->post(route('passport.token'), [
    'form_params' => [
        'grant_type'    => 'password',
        'client_id'     => ?,
        'client_secret' => ?,
        'username'      => $request->get('email'),
        'password'      => $request->get('password'),
        'scope'         => null,
    ],
]);

return response($response->getBody()->getContents(), $response->getStatusCode());
```

### DIFF
--- a/src/Grant/PasswordGrant.php
+++ b/src/Grant/PasswordGrant.php
@@ -104,7 +104,7 @@ class PasswordGrant extends AbstractGrant
         if ($user instanceof UserEntityInterface === false) {
             $this->getEmitter()->emit(new RequestEvent(RequestEvent::USER_AUTHENTICATION_FAILED, $request));
 
-            throw OAuthServerException::invalidGrant();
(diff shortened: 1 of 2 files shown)


Sephster:
Thank you for your pull request. The behaviour is expected according to the OAuth spec. We used to respond with invalid credentials but this was not correct so we modified this a few versions back. Because of this, we won't be going ahead with your pull request at this time.

Thank you very much for your efforts though. I will mark this PR as closed but if you have any follow up comments or questions please don't hesitate to get in touch.

PR #1132, opened 2020-08-12
Replace DateTime with DateTimeImmutable in repository docs.

Aliance:


Sephster:
Thanks for finding that. Much appreciated.

PR #1138, opened 2020-08-31
Add support for specifying iss claim on access token JWT

### LINKED ISSUE #1137: Support adding `issuer` claim to the access token JWT
The JWT [iss](https://tools.ietf.org/html/rfc7519#section-4.1.1) claim is a registered claim name for the JWT standard. Some validator implementations, most notably [envoyproxy](https://www.envoyproxy.io) require the issuer claim to be set in various circumstances, however there's no way right now to specify the claim in the access token emitted by the library.

Pchelolo:
Fixes #1137 

This is a breaking change since the TokenInterface has been changes, so this should not be merged into mater. The PR is created agains master though per contributing guidelines.

### DIFF
--- a/examples/src/Repositories/AccessTokenRepository.php
+++ b/examples/src/Repositories/AccessTokenRepository.php
@@ -16,6 +16,20 @@ use OAuth2ServerExamples\Entities\AccessTokenEntity;
 
 class AccessTokenRepository implements AccessTokenRepositoryInterface
 {
+
+    /**
+     * @var string
+     */
+    private $issuer;
+
+    /**
+     * @param string $domain token issuer identifier
+     */
+    public function __construct($issuer)
+    {
+        $this->$issuer = $issuer;
+    }
+
     /**
      * {@inheritdoc}
      */
@@ -51,6 +65,7 @@ class AccessTokenRepository implements AccessTokenRepositoryInterface
             $accessToken->addScope($scope);
         }
         $accessToken->setUserIdentifier($userIdentifier);
+        $accessToken->setIssuer($this->domain);
 
         return $accessToken;
     }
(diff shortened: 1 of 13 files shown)


Sephster:
We will not progress with this at this time as per reasoning in issue #1137. Thank you for your time and apologies we won't be progressing this PR on this occasion.

src/Entities/Traits/AccessTokenTrait.php
```diff
@@ -48,7 +48,14 @@ private function convertToJWT(CryptKey $privateKey)
             ->issuedAt(\time())
             ->canOnlyBeUsedAfter(\time())
             ->expiresAt($this->getExpiryDateTime()->getTimestamp())
-            ->relatedTo((string) $this->getUserIdentifier())
+            ->relatedTo((string) $this->getUserIdentifier());
+
+        if ($this->getIssuer()) {
```
reedy: And here... Code would've been unreachable...

examples/src/Repositories/AccessTokenRepository.php
```diff
@@ -16,6 +16,20 @@
...
+     * @var string
+     */
+    private $issuer;
+
+    /**
+     * @param string $domain token issuer identifier
+     */
+    public function __construct($issuer)
+    {
+        $this->$issuer = $issuer;
```
reedy: `$this->$issuer` is wrong too...

examples/src/Repositories/AccessTokenRepository.php
```diff
@@ -51,6 +65,7 @@ public function getNewToken(ClientEntityInterface $clientEntity, array $scopes,
             $accessToken->addScope($scope);
         }
         $accessToken->setUserIdentifier($userIdentifier);
+        $accessToken->setIssuer($this->domain);
```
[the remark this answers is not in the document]
Sephster: this PR was never merged and has been closed for some time

src/Entities/Traits/AccessTokenTrait.php
Sephster: this PR was never merged and has been closed for some time. Just flagging in case you've worked on something to do with this

PR #1149, opened 2020-11-12
Use proper if in workflow

driesvints:
Just noticed the code coverage step wasn't running on master. This should fix it.


Sephster:
Thanks @dries - looks great.

PR #1164, opened 2020-11-30
fix for passphrase with private key

bradymiller:
I had to make the following change for passphrase with private key to work. Otherwise the passphrase does not make it into any of the sign etc. functions in lcobucci/jwt library. (note this bug cropped up in the most recent versions of oauth2-server/jwt that tackled PHP8)

It is striking me as odd that we are also sending the passphrase as the verification key, which I am still looking into.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [8.2.2] - released 2020-11-30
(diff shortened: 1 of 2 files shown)


Sephster:
Thanks @bradymiller - I'd confused the verificationKey with the passphrase. We don't need a verificationKey so I've just set this to be blank. Thanks for the spotting this and the subsequent fix.

bradymiller:
hi @Sephster , Sure thing. Thanks for fixing it up and getting a release out so quick!

Issue #1117, opened 2020-05-28
Question on refresh token scopes


Misosooup:
At the moment, refresh token scopes are returned base on the access token scopes set. 

When a user request for a new access token via the refresh_token grant type and he/she requested lesser scope than what the original access token has. Should the refresh token have the original scopes or should the refresh token has the new scopes requested.

1. If the refresh token has the new scopes requested, does this mean that eventually, they will run out of scopes if they keep requesting lesser scopes?
2. Should the refresh token keep the original scopes? This would mean that an access token returned would have different scopes as to what is stored in the refresh token, and the next request to get a new access token may result in more scopes than the current access token.

Can someone please enlighten me on this issue?

I have read the RFC docs and there is a point that states
`If a new refresh token is issued, the refresh token scope MUST be identical to that of the refresh token included by the client in the request.`

Sephster:
It looks like we should be issuing the refresh token with the same scopes as the original, regardless of what scopes were requested.

I think this should probably be changes to better match the spec. Thanks for flagging.

PR #1166, opened 2020-12-02
The refresh_token grant should just reuse the scopes of the previously issued access token

morvans:
First attempt to tackle https://github.com/thephpleague/oauth2-server/issues/1117.


Sephster:
Thanks for your PR. Unfortunately this issue can't be fixed without a breaking change. The refresh token currently derives its scopes from the attached access token. The access token can have fewer scopes than the original refresh token so we need to adjust the way scopes are assigned to the refresh token which will cause a BC break. Thanks for your PR but unfortunately we won't be able to go further with this just now.

PR #1168, opened 2020-12-02
Upgrade to League/Event 3.0

beryllium:
I started this PR because a project I was working on was butting heads with OAuth2Server + Event 3.0. I saw the dependabot PR that was failing tests and thought I'd contribute a working PR.

Some changes:

- Events are now using Classes instead of Names
- Subscribing to events now involves calling `$dispatcher->subscribeTo(MyEventClass::class, $invokableOrCallable)`

The new event class names are:

- `League\OAuth2\Server\Events\ClientAuthenticationFailed`
- `League\OAuth2\Server\Events\ClientRefreshTokenFailed`
- `League\OAuth2\Server\Events\IssuedAccessToken`
- `League\OAuth2\Server\Events\IssuedRefreshToken`
- `League\OAuth2\Server\Events\UserAuthenticationFailed`

Documentation updates will be required to guide users through the changes they might need to make in their own systems.


Sephster:
Thanks for putting in the effort with this. It is on our radar to add support for this but it will require a BC change so we aren't going to add it in until v9 which is some way away just now. Apologies we won't move forwards with this at this time.

beryllium:
That's quite understandable. No worries. (It turns out that the package wasn't actually being utilized in the product I was working on, so I was able to remove the dependency altogether.)

eugene-borovov:
Half a step away from completely abandoning league/event in favor of psr/event-dispatcher. :)

Sephster:
Latest version of league/event uses psr/event-dispatcher so will be moving to that at our next major 👍

eugene-borovov:
The league/oauth2-server should depend on psr/event-dispatcher rather than league/event implementation. That's what I wanted to say.

Sephster:
Unsure of the implementation details yet but we will use the interface so you will be free to swap out if you wish if we do use event in some capacity. I'm hoping we will.

Issue #1162, opened 2020-11-29
OAuthServerException::invalidClient() produces browser login prompt


symbioquine:
In experimenting with the Drupal's simple_oauth module, I have come across some surprising behavior.

Suppose site A uses thephpleague/oauth2-server and site B wants to connect to site A on behalf of site A's users. If the developers of site B use an oauth client id which is missing from site A, their users will get a HTTP error code 401 access denied - browsers respond to that code by showing an authentication prompt like this;

![image](https://user-images.githubusercontent.com/30754460/100551349-b289de00-3234-11eb-8326-35a241165420.png)

This doesn't make sense because nothing site A's users can enter here will fix the problem in site B's configuration.

The code which returns an HTTP code 401 is [src/Exception/OAuthServerException.php#L154](https://github.com/thephpleague/oauth2-server/blob/284c2b5125ac71c6564e96d6e3c2d6eef3a1ae24/src/Exception/OAuthServerException.php#L154);

```php
/**
 * Invalid client error.
 *
 * @param ServerRequestInterface $serverRequest
 *
 * @return static
 */
public static function invalidClient(ServerRequestInterface $serverRequest)
{
    $exception = new static('Client authentication failed', 4, 'invalid_client', 401);

    $exception->setServerRequest($serverRequest);

    return $exception;
}
```

According to [RFC 6749 section-4.1.2.1](https://tools.ietf.org/html/rfc6749#section-4.1.2.1);

> 4.2.2.1.  Error Response
> 
>    If the request fails due to a missing, invalid, or mismatching
>    redirection URI, **or if the client identifier is missing or invalid,
>    the authorization server SHOULD inform the resource owner of the
>    error** and MUST NOT automatically redirect the user-agent to the
>    invalid redirection URI.

To effectively inform the resource owners (site A's users), it would make more sense to return an HTTP 400 error and a message like;"Request to access resources by invalid client 'example_client_id'."

Sephster:
I see your concern here but the server is acting as expected in the spec. Because the client identifier sent to the server is invalid this triggers an `invalid_client` error response as the client is unknown and should issue a 401 response.

Surfacing this information to the resource owner should, in most situations, not be useful. The resource owner would likely not have access to site B's client code to make any meaningful adjustments but I suppose they could inform the site B admin. However, I agree with you that an endless login prompt could well be confusing. 

This is the section in the RFC we are following:

> invalid_client
               Client authentication failed (e.g., unknown client, no
[...]

PR #1170, opened 2020-12-08
Fix confusing auth behavior when used with a common .htaccess pattern

symbioquine:
**Why:** It is common for .htaccess files - including the default one
provided by Drupal Core - to coerce the `Authorization` header to the
empty string if not specified by the client. Without this fix, that
results in a `WWW-Authenticate: Basic` header response which causes
the browser to show a useless basic authentication dialog.

For more info see https://github.com/thephpleague/oauth2-server/issues/1162

### DIFF
--- a/src/Exception/OAuthServerException.php
+++ b/src/Exception/OAuthServerException.php
@@ -334,13 +334,12 @@ class OAuthServerException extends Exception
         // respond with an HTTP 401 (Unauthorized) status code and
         // include the "WWW-Authenticate" response header field
         // matching the authentication scheme used by the client.
-        // @codeCoverageIgnoreStart
-        if ($this->errorType === 'invalid_client' && $this->serverRequest->hasHeader('Authorization') === true) {
+        if ($this->errorType === 'invalid_client' && $this->requestHasAuthorizationHeader()) {
             $authScheme = \strpos($this->serverRequest->getHeader('Authorization')[0], 'Bearer') === 0 ? 'Bearer' : 'Basic';
 
             $headers['WWW-Authenticate'] = $authScheme . ' realm="OAuth"';
         }
-        // @codeCoverageIgnoreEnd
+
         return $headers;
     }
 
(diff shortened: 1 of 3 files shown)


src/Exception/OAuthServerException.php
```diff
@@ -334,13 +334,12 @@ public function getHttpHeaders()
         // respond with an HTTP 401 (Unauthorized) status code and
         // include the "WWW-Authenticate" response header field
         // matching the authentication scheme used by the client.
-        // @codeCoverageIgnoreStart
-        if ($this->errorType === 'invalid_client' && $this->serverRequest->hasHeader('Authorization') === true) {
+        if ($this->errorType === 'invalid_client' && $this->requestHasAuthorizationHeader()) {
```
Sephster, line 337:
```suggestion
        if ($this->errorType === 'invalid_client' && !empty($this->serverRequest->getHeader('Authorization'))) {
```

We could remove the function and replace with the line change above I think

Sephster (changes requested):

src/Exception/OAuthServerException.php
symbioquine: I don't think so. The case we're trying to detect has a *non-empty* array containing a single empty string.

It would need to be something more like;

```patch
-        if ($this->errorType === 'invalid_client' && $this->requestHasAuthorizationHeader()) {
+        if ($this->errorType === 'invalid_client' && !empty($this->serverRequest->getHeader('Authorization')) && !empty($this->serverRequest->getHeader('Authorization')[0])) {
```

I think you suggested putting it out in a separate function on https://github.com/thephpleague/oauth2-server/issues/1162 because the above gets a bit unwieldy and didn't leave a good place to include an explanatory comment for this case.

Likely irrelevant, but if the code didn't need to be written to support earlier versions of PHP, it could be written more tersely (cryptically?) along these lines;

```php
$headers = $this->serverRequest->getHeaders();

if ($this->errorType === 'invalid_client' && !empty($headers['Authorization'][0] ?? '')) {
```

Sephster: Ahh sorry. I thoght it returned an empty array. Cheers for the explanation. Makes sense

Sephster:
Looks great thanks @symbioquine - I've updated the changelog and happy to merge. Thank you

PR #1173, opened 2020-12-11
Throw OAuthServerException::invalidCredentials for password grant

xterr:
**Problem:**
When then user/password combination is incorrect, the exception thrown by the Password Grant is `OAuthServerException::invalidGrant()`

There is no way of knowing that the user/pass combination is incorrect. 
A temporary solution would be to directly throw `OAuthServerException::invalidCredentials()` from the UserRepository implementation in case the user is not found but the `RequestEvent::USER_AUTHENTICATION_FAILED` event is not emitted anymore.

**Solution:**
Change it to `OAuthServerException::invalidCredentials()`


Sephster:
Thanks for your PR. The invalid_credentials response is not part of the OAuth 2 spec so is being removed in version 9. We also plan on providing a better error description in version 9 which will be more explicit so will not merge this in at this time. Thank you very much for your contribution though and sorry we won't be progressing it this time.

PR #1178, opened 2020-12-21
Fix wrong cmd to gen private key with passphrase

PATROMO:
To generate a private key with passphrase an encryption must be specified `-aes128`. Otherwise a key without passphrase will be created and no error will occur. The stupid thing is that the PHP code does not throw an error if you specify a passphrase for a private key without passphrase. Therefore no one notices the error.

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -31,7 +31,7 @@ openssl genrsa -out private.key 2048
 If you want to provide a passphrase for your private key run this command instead:
 
 ~~~ shell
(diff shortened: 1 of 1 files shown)


Sephster:
Thanks for spotting this. I am surprised openssl doesn't produce any error. Just silently creates the priv key without a password. Scary. 

It looks like we should also update this to use `genpkey` instead of `genrsa` going forwards. Thanks very much for this.

PR #1180, opened 2020-12-24
Use key as string without a temporary file

ssigwart:
- Addresses thephpleague/oauth2-server#1007
- This allows you to use a private key from a string without writing to a temporary file.  For security reasons, I'd rather not have the key written to disk on the server.

### DIFF
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -54,7 +54,7 @@ class CryptKey
             $this->keyPath = '';
             // There's no file, so no need for permission check.
             $keyPermissionsCheck = false;
-        } else if (\is_file($keyPath)) {
+        } elseif (\is_file($keyPath)) {
             if (\strpos($keyPath, self::FILE_PREFIX) !== 0) {
                 $keyPath = self::FILE_PREFIX . $keyPath;
             }
@@ -66,11 +66,11 @@ class CryptKey
             $this->keyPath = $keyPath;
             $this->key = LocalFileReference::file($this->keyPath, $this->passPhrase ?? '');
             if (!$this->isValidKey($contents, $this->passPhrase ?? '')) {
-               throw new LogicException('Unable to read key from file ' . $keyPath);
+                throw new LogicException('Unable to read key from file ' . $keyPath);
             }
         }
         else {
-           throw new LogicException('Unable to read key from file ' . $keyPath);
+            throw new LogicException('Unable to read key from file ' . $keyPath);
         }
 
         if ($keyPermissionsCheck === true) {
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -49,11 +49,11 @@ class CryptKey
         $this->passPhrase = $passPhrase;
 
         if (\strpos($keyPath, self::FILE_PREFIX) !== 0 && $this->isValidKey($keyPath, $this->passPhrase ?? '')) {
-          $contents = $keyPath;
-          $this->key = InMemory::plainText($keyPath, $this->passPhrase ?? '');
-          $this->keyPath = '';
-          // There's no file, so no need for permission check.
-          $keyPermissionsCheck = false;
+            $contents = $keyPath;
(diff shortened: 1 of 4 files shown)


Sephster (changes requested):
I think we are removing the ability to use a password against the private key here. Is it possible to change this to accommodate this? Thanks for your submission.

src/Entities/Traits/AccessTokenTrait.php
```diff
@@ -46,7 +45,7 @@ public function initJwtConfiguration()
     {
         $this->jwtConfiguration = Configuration::forAsymmetricSigner(
             new Sha256(),
-            LocalFileReference::file($this->privateKey->getKeyPath(), $this->privateKey->getPassPhrase() ?? ''),
+            $this->privateKey->getKey(),
```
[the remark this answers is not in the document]
ssigwart: The constructor for `CryptKey` handles it for a file.  However, it looks like I was missing the ability for a string key, so I added that.

ssigwart (commented):
Thanks for looking @Sephster.  I think my change addresses your comment.

ssigwart:
@Sephster, is there anything I should do to get this merged?

ssigwart:
@Sephster, there are merge conflicts now.  Should I rebase and resolve them?

Sephster:
That would be appreciated cheers. I've been tidying up a personal project but hoping to get back to this within the week. Thanks very much

ssigwart:
Thanks, @Sephster.  I rebased it and fixed the conflicts.

src/CryptKey.php
```diff
@@ -41,21 +48,28 @@ public function __construct($keyPath, $passPhrase = null, $keyPermissionsCheck =
     {
         $this->passPhrase = $passPhrase;
 
-        if (\is_file($keyPath)) {
+        if (\strpos($keyPath, self::FILE_PREFIX) !== 0 && $this->isValidKey($keyPath, $this->passPhrase ?? '')) {
+            $contents = $keyPath;
+            $this->key = InMemory::plainText($keyPath, $this->passPhrase ?? '');
```
eugene-borovov, line 53: The `CryptKey` class is a DTO with fast fail validation. You should store key content instead of create `Lcobucci\JWT\Signer\Key` instance. `CryptKey` and  `Lcobucci\JWT\Signer\Key` are the same. Why not use the `Lcobucci\JWT\Signer\Key` directly?

ssigwart: Hi, @eugene-borovov.  Thanks for reviewing, but I don't really understand your comment.  `CryptKey` is not a subclass of `Lcobucci\JWT\Signer\Key`, so I don't know why you said they are the same.

I could store `$contents` as the class variable instead of `$key`.  It seems like it would be more efficient to once have to create the key object once though.  Sorry if I'm missing something obvious here.

eugene-borovov: Hi, @ssigwart.
I apologize for my confused question. I wanted to say that it is better to create `Lcobucci\JWT\Signer\Key` immediately before using it in `lcobucci\jwt`. Otherwise, an extra dependency is created on `lcobucci\jwt`, which increases the coherence with the external library. Do you agree?

ssigwart: Thanks, @eugene-borovov.  It sounds like you want as few places as possible referencing `lcobucci\jwt`, so I updated it to save just the key contents.  Then I updated the places that used the keys to set up the `Lcobucci\JWT\Signer\Key` object since those already were using `lcobucci\jwt`.

eugene-borovov (approved):

Sephster:
@ssigwart would it be possible for you to resolve the conflict? I tried to do this myself but don't have write access to your branch. There is also a trailing comma on line 80 of the BearerToken file which is causing issues in PHP 7.2. Thank you

ssigwart:
@Sephster, I fixed the trailing comma and rebased to fix the conflict.  Thanks!

Sephster (approved):

Sephster:
Thanks @ssigwart - looks great.

Issue #1092, opened 2020-02-12
Default Scope does not work as expected


SOHELAHMED7:
Server - http://localhost:7876/web/index.php

Client - http://localhost:7878/web/index.php

When we have URL (present in client app to Oauth with server)

scope - email is present - this case works fine 

http://localhost:7876/web/index.php?r=oauth/default/authorize&client_id=1&response_type=code&scope=email&redirect_uri=http%3A%2F%2Flocalhost%3A7878%2Fweb%2Findex.php%3Fr%3Dsite%2Fredirectedhere

Both below cases  does not set scopes but also does not give error, (please see scope in URL), Why?

http://localhost:7876/web/index.php?r=oauth/default/authorize&client_id=1&response_type=code&scope=&redirect_uri=http%3A%2F%2Flocalhost%3A7878%2Fweb%2Findex.php%3Fr%3Dsite%2Fredirectedhere

http://localhost:7876/web/index.php?r=oauth/default/authorize&client_id=1&response_type=code&scope=0&redirect_uri=http%3A%2F%2Flocalhost%3A7878%2Fweb%2Findex.php%3Fr%3Dsite%2Fredirectedhere

`AuthCodeGrant.php`
```php
$scopes = $this->validateScopes(
            $this->getQueryStringParameter('scope', $request, $this->defaultScope),
            $redirectUri
        );
```

`validateScopes` will return empty array when scope is empty string or 0 but present in request query string as shown above

`AbstractGrant.php`
```php
/**
     * Validate scopes in the request.
     *
     * @param string|array $scopes
     * @param string       $redirectUri
     *
     * @throws OAuthServerException
     *
     * @return ScopeEntityInterface[]
     */
    public function validateScopes($scopes, $redirectUri = null)
    {
        if (!\is_array($scopes)) {
            $scopes = $this->convertScopesQueryStringToArray($scopes);
        }

        $validScopes = [];

        foreach ($scopes as $scopeItem) {
            $scope = $this->scopeRepository->getScopeEntityByIdentifier($scopeItem);

            if ($scope instanceof ScopeEntityInterface === false) {
                throw OAuthServerException::invalidScope($scopeItem, $redirectUri);
            }

            $validScopes[] = $scope;
        }

        return $validScopes;
    }
```

```php
/**
     * Converts a scopes query string to an array to easily iterate for validation.
     *
     * @param string $scopes
     *
     * @return array
     */
    private function convertScopesQueryStringToArray($scopes)
    {
[...]

PR #1181, opened 2021-01-03
Default Scope does not work as expected

datapp:
resolves issue #1092


Sephster:
Thanks @datapp - merged as confirmed fix.

Issue #1145, opened 2020-10-29
Domain events do not provide information about tokens


halfpastfouram:
When a domain event is emitted the receiver of that event has no way to access information about for instance the access/refresh-token that was issued.

### Situation

Our access tokens have a supplementary property called `mfa_validated`. That property's value is either `true` or `false`. When issuing a new access token that was requested via a refresh token the new access token should inherit the value of that property from the previous access token.

We're unable to find out what the previous access token or refresh token was. We are also unable to find out what the new access token is. This results in our users having to redo the MFA steps when the access token is expired but the refresh token is still valid.

### Proposal

- `League\OAuth2\Server::ACCESS_TOKEN_ISSUED`: Provide the new access token to the `RequestEvent` when a new access token was issued. 
- `League\OAuth2\Server::REFRESH_TOKEN_ISSUED`: Provide the new refresh token to the `RequestEvent` when a new refresh token was issued.

Or provide any other way to retrieve the access token or refresh token related to the event that was emitted.

halfpastfouram:
Any comments on this are very much welcome. I can create a PR if you'd like.

GDmac:
Isn't your custom property `mfa_validated` something you 
can store in your own implementation of `OurCustomToke implements TokenInterface` ?

The moment your implementation of `AccessTokenRepositoryInterface` 
or `RefreshTokenRepositoryInterface` returns a new token, it can set the property.

halfpastfouram:
Late response, but no that won't help. The problem is that we need our logic to run AFTER the token is issued and BEFORE the http response is generated.

I'm thinking about suggesting a small change that would introduce two new event classes to make our situation possible until #1184 is fully implemented. In short, it would introduce `AccessTokenIssuedEvent` and `RefreshTokenIssuedEvent`, both extending the `RequestEvent` class. The first would accept an issued access token as a third parameter that can be accessed via an event listener. The second would do the same but with a refresh token.

halfpastfouram:
Just checked. It seems my issue is solved in #1211 that was released in 8.3.0. I will close this issue if I can use this solution.

GDmac:
please tell me you are not modifying the token object thru an dispatched/submitted event

halfpastfouram:
@GDmac In our case a user has to go through MFA. When MFA was succesful, the token's property `mfa_validated` will be set to true. That will not happen through an event.

[...]

PR #1184, opened 2021-01-17
PSR-14 Event Dispatcher

eugene-borovov:
This PR inspired of #1168 and #1145.

* Low coupling with league/event.
* Events have individual object type.
* Event objects enriched by contextual data.
* Code of examples was updated. Now it contain Event Dispatcher usage.


src/Events/EventDispatchableTrait.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    /** @var EventDispatcherInterface|null */
+    protected $eventDispatcher;
+
+    public function useEventDispatcher(?EventDispatcherInterface $eventDispatcher): void
+    {
+        $this->eventDispatcher = $eventDispatcher;
+    }
+
+    public function dispatchEvent(object $event): object
```
derrabus, line 17:
```suggestion
    protected function dispatchEvent(object $event): object
```

This method is only called as `$this->dispatchEvent()`, so there is no need for this method to be public.

src/Events/EventDispatchableInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace League\OAuth2\Server\Events;
+
+use Psr\EventDispatcher\EventDispatcherInterface;
+
+interface EventDispatchableInterface
```
derrabus, line 7:
```suggestion
interface EventDispatcherAwareInterface
```

This `SomethingAwareInterface` + `Trait` is a common naming scheme that has been used by the previous implementation as well. I don't think that we should invent something new here.

src/Events/EventDispatchableTrait.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace League\OAuth2\Server\Events;
+
+use Psr\EventDispatcher\EventDispatcherInterface;
+
+trait EventDispatchableTrait
```
derrabus, line 7:
```suggestion
trait EventDispatcherAwareTrait
```

derrabus (changes requested):
Thank you for working on this. Switching to PSR-14 would certainly help with integrating this library into frameworks that already provide a PSR-14 event dispatcher.

src/Events/EventDispatchableInterface.php
```diff
@@ -0,0 +1,12 @@
...
+{
+    public function useEventDispatcher(?EventDispatcherInterface $dispatcher): void;
+
+    public function dispatchEvent(object $event): object;
```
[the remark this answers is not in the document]
eugene-borovov: > All methods declared in an interface must be public; this is the nature of an interface. [php.net](https://www.php.net/manual/en/language.oop5.interfaces.php)

You're right, but I won't be able to change the access level of the method.

src/Events/EventDispatchableTrait.php
eugene-borovov: I guess I need to make `dispatchEvent` method private and remove from interface.

src/Events/EventDispatchableInterface.php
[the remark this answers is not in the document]
derrabus: So why is this method on the interface anyway?

src/Events/EventDispatchableInterface.php
eugene-borovov: Ok. I changed interface name.

src/Events/EventDispatchableInterface.php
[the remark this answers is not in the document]
eugene-borovov: There is no reason. I removed method from interface. Thanks.

TomHAnderson:
What's the hold-up on this PR and PSR-14?  I'm using the `league/event` 3.0 in my GraphQL library https://github.com/API-Skeletons/doctrine-graphql but can't use this League library in tandem.

stof:
@Sephster is there any chance to get this merged ? It would make things a lot easier for frameworks (for instance, symfony/event-dispatcher implements the PSR-14 interface so the bundle could drop its current adapter layer forwarding events to the Symfony EventDispatcher)

PR #1189, opened 2021-02-11
Add switch to prevent revoking of refresh tokens.

janhopman-nhb:
Hi,
Recently we (my colleagues and i) came across a specific use case in our software which uses oauth2-server where we needed to prevent refresh tokens to be revoked after requesting a new access token. We looked at the ouath rfc and read that this behaviour is allowed according to the spec. I've gone ahead and implemented the change in this pull request. See below for references to the specific information on refresh tokens in the rfc.

According to [section 1.5](https://tools.ietf.org/html/rfc6749#section-1.5), issuing a new refresh token is optional (see step H).
Also according to [section 4.1](https://tools.ietf.org/html/rfc6749#section-4.1) which describes the authorization code grant type, the refresh token is optional (see step E).
Finally according to [section 6](https://tools.ietf.org/html/rfc6749#section-6) (Refreshing an Access token), in the last paragraph it states that the server MAY issue a new refresh token.
And also states that the server MAY revoke the old refresh token.

Hoping someone can take a look at this, any suggestions and/or feedback, would be much appreciated!

### DIFF
--- a/src/AuthorizationServer.php
+++ b/src/AuthorizationServer.php
@@ -82,7 +82,7 @@ class AuthorizationServer implements EmitterAwareInterface
     /**
      * @var bool
      */
-    private $revokeRefreshTokens;
+    private $revokeRefreshTokens = true;
 
     /**
      * New server instance.
@@ -93,7 +93,6 @@ class AuthorizationServer implements EmitterAwareInterface
      * @param CryptKey|string                $privateKey
      * @param string|Key                     $encryptionKey
      * @param null|ResponseTypeInterface     $responseType
-     * @param bool                           $revokeRefreshTokens
      */
     public function __construct(
         ClientRepositoryInterface $clientRepository,
@@ -101,8 +100,7 @@ class AuthorizationServer implements EmitterAwareInterface
         ScopeRepositoryInterface $scopeRepository,
         $privateKey,
         $encryptionKey,
-        ResponseTypeInterface $responseType = null,
-        bool $revokeRefreshTokens = true
+        ResponseTypeInterface $responseType = null
     ) {
         $this->clientRepository = $clientRepository;
         $this->accessTokenRepository = $accessTokenRepository;
@@ -122,7 +120,6 @@ class AuthorizationServer implements EmitterAwareInterface
         }
 
         $this->responseType = $responseType;
-        $this->revokeRefreshTokens = $revokeRefreshTokens;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Sephster:
Thanks @janhopman-nhb  - is there a way to implement this that won't break BC? If not, I think we will leave this PR for now. I also want to change refresh tokens so they don't have an expiration as this isn't part of the spec so would like to roll that into one as part of a major release. Thanks

janhopman-nhb:
Hi @Sephster, Thank you for your reply! We will have a look if we can implement without breaking BC.

janhopman-nhb:
Hi @Sephster, is it possible to approve running the workflows by any chance?

Sephster:
Done, thanks!

janhopman-nhb:
@Sephster Really sorry for bothering you... 😅  Just fixed the bc issue i think, but the workflows need to be triggerd again to confirm...

src/AuthorizationServer.php
```diff
@@ -88,14 +93,16 @@ class AuthorizationServer implements EmitterAwareInterface
...
      */
     public function __construct(
         ClientRepositoryInterface $clientRepository,
         AccessTokenRepositoryInterface $accessTokenRepository,
         ScopeRepositoryInterface $scopeRepository,
         $privateKey,
         $encryptionKey,
-        ResponseTypeInterface $responseType = null
+        ResponseTypeInterface $responseType = null,
+        bool $revokeRefreshTokens = true
```
Sephster, line 105: Can we remove this from the constructor? I'm already conscious that the constructors are quite heavy in the lib. Lets set the class variable to true on init instead and remove this from the constructor. Anyone that wants to change it can just call the setter function. Thank you

tests/Grant/RefreshTokenGrantTest.php
```diff
@@ -68,6 +68,7 @@ public function testRespondToRequest()
         $grant->setAccessTokenRepository($accessTokenRepositoryMock);
         $grant->setEncryptionKey($this->cryptStub->getKey());
         $grant->setPrivateKey(new CryptKey('file://' . __DIR__ . '/../Stubs/private.key'));
+        $grant->setRevokeRefreshTokens(true);
```
Sephster, line 71: Would you be able to add in a test to check that when we revoke a token, it is revoked and when we don't, we are still able to use it?

Sephster (changes requested):
Thank you - a few comments but this is looking good thanks!

tests/Grant/RefreshTokenGrantTest.php
janhopman-nhb: Thank you for your comments, going to have a look if i can get this done!

Sephster (approved):

Sephster:
LGTM - thank you for your work on this @janhopman-nhb

PR #1193, opened 2021-02-26
README en Español

Juangemelo01:
Traducción al Español del archivo README, para que otros lectores que tengan dificultad con el ingles, puedan interpretar mejor la información y se enriquezca mas la comunidad.


Juangemelo01:
Actualizando contenido al Español
Updating content to Spanish

Sephster:
I think such translations would probably be better on our website. This is common with other repos such as Laravel and React which have English readmes but internationalised websites. 

Thank you for your effort but to keep the maintenance efforts for this package as easy as possible, I won't be moving forwards with this PR.

PR #1198, opened 2021-03-07
Separation of token parsing and validation

### LINKED ISSUE #1197: Test invalid JWT fails with lcobucci/jwt:4.1.2
[Test invalid JWT](https://github.com/thephpleague/oauth2-server/blob/master/tests/ResponseTypes/BearerResponseTypeTest.php#L155) fails with `lcobucci/jwt:4.1.2` and OK with `lcobucci/jwt:4.0.0`

The reason is `lcobucci/jwt:4.0.0` try to check invalid JWT and `lcobucci/jwt:4.1.2` fail to parse base64.
Expected test result is `Access token could not be verified`. With `lcobucci/jwt:4.1.2` we\`ve got `Error while decoding from Base64Url, invalid base64 characters detected`.

eugene-borovov:
Fixes #1197

Test `testDetermineAccessTokenInHeaderInvalidJWT` asserts only on validation error. Parse errors will be catched by test `testDetermineAccessTokenInHeaderInvalidToken`.

### DIFF
--- a/src/AuthorizationValidators/BearerTokenValidator.php
+++ b/src/AuthorizationValidators/BearerTokenValidator.php
@@ -12,12 +12,9 @@ namespace League\OAuth2\Server\AuthorizationValidators;
 use DateTimeZone;
 use Lcobucci\Clock\SystemClock;
 use Lcobucci\JWT\Configuration;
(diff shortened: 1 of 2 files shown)


Sephster:
Thank you.

Issue #1199, opened 2021-03-08
Make AuthCodeGrant::validateAuthorizationCode() visibility protected


GameplayJDK:
Hello everyone,

I'd like to propose making the `AuthCodeGrant::validateAuthorizationCode() method`'s visibility `protected`. In addition the method should also carry a `@throws` tag.

Currently the method is `private`, which means there is no way to alter (extend or change) the validation logic for the auth code, for example in a sub-class of `AuthCode`.

The new signature would look like this:

```php
    /**
     * Validate the authorization code.
     *
     * @param stdClass               $authCodePayload
     * @param ClientEntityInterface  $client
     * @param ServerRequestInterface $request
     *
     * @throws OAuthServerException
     */
    protected function validateAuthorizationCode(
        $authCodePayload,
        ClientEntityInterface $client,
        ServerRequestInterface $request
    ) {
        // ...
    }
```

I could also provide a pull request, if this change is desired by the maintainers.

Best regards

Sephster:
What kind of change would you want to make to the authorisation mechanism? Can you give an example? Thank you

GameplayJDK:
The use-case at hand is a sub-class of `AuthCode`, which must have access to the submitted auth code before the access and refresh tokens are issued. The intent is to link the provided auth code to the issued tokens for later use.

Sephster:
What is the reason for wanting to link the auth code to the access token? I'm hesitant to make this function non-private as it is a part of core security of the package and a mistake could make implementations less secure.

GameplayJDK:
It would be pretty complicated to explain the whole concept I decides to use for single sign on here.

For the `auth_code` grant, when a authentication request is completed and an auth code is issued, a session is created and already associates with it. I'm currently overriding the `issueAuthCode()` method from the `AbstractGrant` class for that. In the next step, when the access token request is completed, I'll need to check the session of the supplied auth code against the session id sent with the request to verify the integrity of the request. A good point to do this at would be the `validateAuthorizationCode()` method in question here, as it is supplied with the decrypted auth code payload and it can safely be assumed, the code itself is valid when calling the parent method returns without throwing.

For the time being I'm limited to loading a hotfix class via the composer autoload, which is far beyond ideal.

Sephster:
[...]

PR #1200, opened 2021-03-10
Change visibility of `AuthCodeGrant::validateAuthorizationCode()` (#1199)

GameplayJDK:
Currently the method is `private`, which means there is no way to alter (extend or change) the validation logic for the auth code, for example in a sub-class of `AuthCode`.

In addition, add a `@throws` tag to the method for `OAuthServerException`.

See #1199 for reference.


Sephster:
Not going to implement this at this time as per discussion in #1199

PR #1204, opened 2021-03-26
Support for EC keys

NikoGrano:


src/CryptKey.php
```diff
@@ -17,7 +17,7 @@
 class CryptKey
 {
     const RSA_KEY_PATTERN =
-        '/^(-----BEGIN (RSA )?(PUBLIC|PRIVATE) KEY-----)\R.*(-----END (RSA )?(PUBLIC|PRIVATE) KEY-----)\R?$/s';
+        '/^(-----BEGIN (RSA |EC )?(PUBLIC|PRIVATE) KEY-----)\R.*(-----END (RSA |EC )?(PUBLIC|PRIVATE) KEY-----)\R?$/s';
```
eugene-borovov, line 20: Now this is not an RSA pattern. Maybe you should create a separate constant for the EC pattern?

NikoGrano: Renaming the constant would cause BC change. Separating it to two would cause extra check into code for same purpose. 

Please see reference how this constant is being used to understand fully the point of adding "EC" support there.

eugene-borovov: I suppose this constant is private so renaming is safe. If RSA_KEY_PATTERN is public, then changing the expression breaks backward compatibility in any case.

Sephster:
I think we need to change the implementation of this really. The phpseclib seems like it would be a better option to validate and save keys. For this reason, and the fact that EC keys aren't RSA keys so the name might cause confusion, I'm going to leave this PR for now. 

If you would like to attempt a phpseclib implementation I'd welcome a PR.

PR #1211, opened 2021-04-06
Added token events

lookyman:
This allows access to issued tokens in request events.


Sephster:
Looks good thanks @lookyman - I think the intention in the next major will be to have a PSR15 compatible implementation but this is needed to happy to merge in now. Cheers!

PR #1215, opened 2021-04-19
Validate key with openssl

### LINKED ISSUE #1214: Change CryptKey to use PHPSecLib for key validation
As the title says :)

eugene-borovov:
Added key validation using phpseclib. RSA and EC keys are supported. In the future, I would like to make separate types for Public and Private keys to check this aspect as well.

Fixes #1214

### DIFF
--- a/src/CryptKey.php
+++ b/src/CryptKey.php
@@ -41,7 +41,7 @@ class CryptKey
         $this->keyPath = $keyPath;
         $this->passPhrase = $passPhrase;
 
-        if (is_file($this->keyPath) && !$this->isFilePath()) {
+        if (\is_file($this->keyPath) && !$this->isFilePath()) {
             $this->keyPath = self::FILE_PREFIX . $this->keyPath;
         }
 
@@ -50,7 +50,7 @@ class CryptKey
                 throw new LogicException(\sprintf('Key path "%s" does not exist or is not readable', $this->keyPath));
             }
 
-            $contents = file_get_contents($this->keyPath);
+            $contents = \file_get_contents($this->keyPath);
         } else {
             $contents = $keyPath;
         }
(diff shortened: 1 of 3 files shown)


Sephster:
Is there a way to make this a non-BC change? If not, I think we'd need to target this for v9. Also would be good to do something when we encounter an exception. Thank you for this

Spomky:
You don't need PHPSECLIB at all (which is a big fat dep from my POV). As openssl is already required, you just need to get the key details to ensure it is valid: https://www.php.net/manual/fr/function.openssl-pkey-get-details.php

eugene-borovov:
I returned the RSA_KEY_PATTERN constant that the backward compatibility check passed.

eugene-borovov:
Version of the key verification using OpenSSL.  @Sephster , maybe we should try OpenSSL directly?

```php
function isValidKey($contents, $passPhrase)
{
    $pkey = openssl_pkey_get_private($contents, $passPhrase) ?: openssl_pkey_get_public($contents);
    if ($pkey === false) {
        return false;
    }
    $details = openssl_pkey_get_details($pkey);

    return $details !== false && in_array(
        $details['type'] ?? -1,
        [OPENSSL_KEYTYPE_RSA, OPENSSL_KEYTYPE_EC],
        true
    );
}
```

src/CryptKey.php
```diff
@@ -96,7 +108,40 @@ private function saveKeyToFile($key)
...
+            true
+        );
+    }
+
+    /**
+     * Checks whether the key is a file.
+     *
+     * @return bool
+     */
+    private function isFilePath()
```
Sephster, line 142: There it the native `is_file()` function that we could use. I think this would be preferable over rolling our own method unless you can see any issues?

Sephster (changes requested):
Looks good to me. Thanks very much for making this change. If we could use the native `is_file()` and add in some basic tests, this should be good to go. Thanks so much for your efforts with this. Much appreciated!

Spomky:
Excellent!

Looks way more clear and efficient compared to pure PHP functions.
Well done @eugene-borovov!

src/CryptKey.php
eugene-borovov: The purpose of this method is to show the key storage type, but I haven't found any use for it outside of the class. So I removed it completely.

Sephster (approved):

Sephster:
Thank you @eugene-borovov  - looks great. Happy to merge!

PR #1217, opened 2021-04-22
Fix the issue when client id is a number but in the refresh token it'…

dmytro-skelia:
In case if `$refreshTokenData['client_id'] = 22` and `$clientId = "22"` it breaks. This is not an edge case. I faced this issue in Laravel 5.6, that uses Zend Framework to handle the request.

### DIFF
--- a/src/Grant/RefreshTokenGrant.php
+++ b/src/Grant/RefreshTokenGrant.php
@@ -106,7 +106,7 @@ class RefreshTokenGrant extends AbstractGrant
         }
 
         $refreshTokenData = \json_decode($refreshToken, true);
-        if ($refreshTokenData['client_id'] !== $clientId) {
+        if ((string) $refreshTokenData['client_id'] != (string) $clientId) {
             $this->getEmitter()->emit(new RequestEvent(RequestEvent::REFRESH_TOKEN_CLIENT_FAILED, $request));
             throw OAuthServerException::invalidRefreshToken('Token is not linked to client');
         }


src/Grant/RefreshTokenGrant.php
```diff
@@ -106,7 +106,7 @@ protected function validateOldRefreshToken(ServerRequestInterface $request, $cli
         }
 
         $refreshTokenData = \json_decode($refreshToken, true);
-        if ($refreshTokenData['client_id'] !== $clientId) {
+        if ((string) $refreshTokenData['client_id'] != (string) $clientId) {
```
eugene-borovov, line 109: Non-strictly comparison will be enough (https://3v4l.org/EPkhS)

Sephster:
@dmytro-skelia - have you looked into why your refresh token client ID is being decrypted as a number? If it is encrypted as a string, it should be returned as a string. I would anticipate that if this issue was widespread, we would have a lot more issues raised regarding this.

src/Grant/RefreshTokenGrant.php
marc-mabe: It's not! https://3v4l.org/dsmQH

eugene-borovov: > It's not! https://3v4l.org/dsmQH

@marc-mabe you`re right. Comparison of numeric strings is very [tricky](https://www.php.net/manual/en/language.operators.comparison.php). Thank you for reminding me of that.
Obviously, the code should look like this:
```php
if ((string) $refreshTokenData['client_id'] !== (string) $clientId) 
```

Sephster:
Closing this because of a lack of response from @dmytro-skelia - will reopen if I get a reply.

PR #1219, opened 2021-04-25
Change query delimiter sign to question mark

alirezadp10:
Hey there!
I've testing implicit grant type and find out maybe query delimiter default value is wrong.
so when I tried to get token in query string I came across this this URL:
```
http://localhost:5000/third-party#access_token=eyJ0...&state=u4J2...
```
which is not accessible from Illuminate\Support\Facades\Http\Request object.
so I think if query delimiter default value is changed, every thing will be fine
```
http://localhost:5000/third-party?access_token=eyJ0...&state=u4J2...
```


Sephster:
The default in the spec is to have this as a `#`. If you need to change it for your own purposes, please use the constructor. Many thanks.

Issue #1183, opened 2021-01-15
Custumize request attributes in BearerTokenValidator


2618094:
Hello! I'm looking for a way to add custom request attributes for token.
The problem is here:
https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationValidators/BearerTokenValidator.php#L120
There is no option to add attributes except for write my own implementation of validator.

Sephster:
At the moment this is the only way sorry. Hoping to get custom claims added in soon though.

PR #1222, opened 2021-05-03
BC support of Custom Claims

ElisDN:
Easy backward compatible way for parsing custom claims for #1120, #1122 and #1183 in the next 8.x release.

## How to use it

For example, if you want to add a role field into JWT, just add `$role` property into your token entity and override `convertToJWT` method from trait for adding `role` claim:

```php
class AccessToken implements AccessTokenEntityInterface
{
    // ...

    private $userRole;

    public function setUserRole($role) { $this->userRole = $role }
    public function getUserRole() { return $this->userRole }

    private function convertToJWT()
    {
        return $this->jwtConfiguration->builder()
            // ...
            ->withClaim('role', $this->getUserRole())
            // ...
    }
}
```

After all fetch user role and fill the property in token repository:

```php
class AccessTokenRepository implements AccessTokenRepositoryInterface
{
    // ...

    public function getNewToken(ClientEntityInterface $clientEntity, array $scopes, $userIdentifier = null)
    {
        $accessToken = new AccessTokenEntity();
        // ...

        if ($userIdentifier !== null) {

            $row = $this->db->query('SELECT role FROM users WHERE id = :id', ['id' => $userIdentifier]);
            $accessToken->setUserRole($row['role']);
        }

        return $accessToken;
    }

    // ...
}
```

And now you can retreive `$request->getAttribute('oauth_custom_claims')` with value like `['role' => 'admin']`.

### DIFF
--- a/src/AuthorizationValidators/BearerTokenValidator.php
+++ b/src/AuthorizationValidators/BearerTokenValidator.php
@@ -118,7 +118,8 @@ class BearerTokenValidator implements AuthorizationValidatorInterface
             ->withAttribute('oauth_access_token_id', $claims->get('jti'))
             ->withAttribute('oauth_client_id', $this->convertSingleRecordAudToString($claims->get('aud')))
             ->withAttribute('oauth_user_id', $claims->get('sub'))
-            ->withAttribute('oauth_scopes', $claims->get('scopes'));
+            ->withAttribute('oauth_scopes', $claims->get('scopes'))
+            ->withAttribute('oauth_custom_claims', $this->extractCustomClaims($claims->all()));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Sephster:
Thanks @ElisDN - this is a simple solution but does require overriding of existing functions. I think ideally, long term, I'd like something more deliberately baked into the package. There are a few other potential PRs which go further than this one so for that reason, I won't be proceeding with this at present.

Thank you for your efforts though and apologies I won't be merging this in at this time.

systemsolutionweb:
@Sephster neither this nor the other

vrusua:
@Sephster is there any PR finally considered to add custom claims support? It's so useful for SPA and actually discussing here for a couple of years. Thanks.

Sephster:
Definitely want to add this so not off the radar

systemsolutionweb:
Great, it's been a long time, I'm still waiting

PR #1223, opened 2021-05-04
Fix request variable name

ElisDN:
Fixes variable name in test.


Sephster:
Makes sense. Thank you

PR #1224, opened 2021-05-04
Test for parsed attributes

ElisDN:


Sephster:
I don't think this adds anything to the test. The test is checking if the token provided is valid and the present of an auth header verifies this. Thanks for your efforts but I won't be proceeding with this PR

PR #1228, opened 2021-05-07
Decouple from concrete validator

ElisDN:
It decouples `ResourceServer` from checking of concrete `BearerTokenValidator` class.


Sephster:
The server is not tied to the BearerTokenValidator at present, You can use any validator that implements the authorization validator interface. We check if you are using BearerTokenValidator because this required a public key to be set but there are no restrictions or enforcements of using this class.

Thank you for your efforts here but I don't think the original issue was identified correctly.

Issue #1093, opened 2020-02-16
invalid_grant Error When Credentials are Incorrect


Imboga:
I'm using the native oatuh2-serve included in Laravel 6.2.
In the previous version of oauth2-server, I used to face a 400 error when credentials were missing and 401 error when credentials were wrong.
This is not happening since this [PR #967](https://github.com/thephpleague/oauth2-server/pull/967/commits/2b4974b697ca937d37f407c429d792c467af7ec5).

Is this the way you handle returning errors now? How can I know about missing or wrong credentials?

Sorry if this question does not belong here, but not sure where to post it.

Sephster:
Hi @Imboga - yep, we changed this because the old response wasn't conforming to the OAuth 2 specs properly. You could possibly leverage the event to track an invalid credentials attempt

Imboga:
Perfect! Thanks for your fast reply. I will try to find a way to handle this.

lk77:
Hello,

is there a way to restore the old behaviour ?

Sephster:
The only way would be to downgrade to an older version of the library. I would recommend that if you need to track this, you should leverage the events systems instead. 

We won't be reverting this change as the previous implementation didn't conform to the OAuth 2 spec unfortunately.

siarheipashkevich:
@Imboga did you find the solution for your problem?

meyer59:
@Sephster can you provide an example how to handle events to return different status code ?

Sephster:
@meyer59 If you are using the server middleware, you could maybe intercept the exception and change the response to whatever you want it to be. I wouldn't recommend this though as the responses given by the server are valid OAuth responses. Hope this helps

PR #1230, opened 2021-05-21
Respond with helpful and spec complient error on invalid user credentials

marc-mabe:
See #967 #1175 #1093

This PR tries to solve the non helpful error message if a user types in wrong credentials in a a much smaller scope then done in #1093 and hopefully can be merged and released much faster.

In #967 the error type should be `invalid_grant` (instead of `invalid_credentials`) and the HTTP status code should be `400` instead of `401` but this also had changed the error message from `The user credentials were incorrect.` into `The provided authorization grant (e.g., authorization code, resource owner credentials) or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.` which is not user friendly at all.

Instead of calling `invalidGrant()` (as done in #967) this PR calls `invalidCredentials()` but modifies the logic of `invalidCredentials()` to return `invalid_grant` with `400`.

Interestingly `invalidCredentials()` wasn't used anymore since #967.

TODOs
- [ ] Update tests

### DIFF
--- a/src/Exception/OAuthServerException.php
+++ b/src/Exception/OAuthServerException.php
@@ -189,7 +189,7 @@ class OAuthServerException extends Exception
      */
     public static function invalidCredentials()
     {
-        return new static('The user credentials were incorrect.', 6, 'invalid_credentials', 401);
+        return new static('The user credentials were incorrect.', 6, 'invalid_grant', 400);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Sephster:
Hi @marc-mabe - seems like a good suggestion. Happy to approve this once the tests are fixed. Thank you

marc-mabe:
@Sephster  done

Sephster:
LGTM. Thanks @marc-mabe

PR #1237, opened 2021-07-12
Checking the URI before matching

eugene-borovov:
A redirect URI without scheme causes an error.

### DIFF
--- a/src/RedirectUriValidators/RedirectUriValidator.php
+++ b/src/RedirectUriValidators/RedirectUriValidator.php
@@ -41,6 +41,11 @@ class RedirectUriValidator implements RedirectUriValidatorInterface
      */
     public function validateRedirectUri($redirectUri)
     {
+        $redirectUri = \filter_var($redirectUri, FILTER_VALIDATE_URL);
+        if ($redirectUri === false) {
+            return false;
+        }
+
         if ($this->isLoopbackUri($redirectUri)) {
             return $this->matchUriExcludingPort($redirectUri);
         }
--- a/src/RedirectUriValidators/RedirectUriValidator.php
+++ b/src/RedirectUriValidators/RedirectUriValidator.php
@@ -66,8 +66,8 @@ class RedirectUriValidator implements RedirectUriValidatorInterface
     {
         $parsedUrl = \parse_url($redirectUri);
 
-        return $parsedUrl['scheme'] === 'http'
-            && (\in_array($parsedUrl['host'], ['127.0.0.1', '[::1]'], true));
+        return ($parsedUrl['scheme'] ?? '') === 'http'
+            && (\in_array(($parsedUrl['host'] ?? ''), ['127.0.0.1', '[::1]'], true));
     }
 
     /**
--- a/src/RedirectUriValidators/RedirectUriValidator.php
+++ b/src/RedirectUriValidators/RedirectUriValidator.php
@@ -132,7 +132,7 @@ class RedirectUriValidator implements RedirectUriValidatorInterface
             return false; // obviously invalid url
         }
 
-        /**
+        /*
          * @see https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2
          *
          * The redirection endpoint URI
--- a/src/RedirectUriValidators/RedirectUriValidator.php
+++ b/src/RedirectUriValidators/RedirectUriValidator.php
@@ -127,6 +127,7 @@ class RedirectUriValidator implements RedirectUriValidatorInterface
     private static function isValidRedirectUrl($url)
     {
         $parsedUrl = \parse_url($url);
+
         return $parsedUrl !== false && (isset($parsedUrl['host']) || isset($parsedUrl['path']));
     }
 }
--- a/src/RedirectUriValidators/RedirectUriValidator.php
+++ b/src/RedirectUriValidators/RedirectUriValidator.php
@@ -128,6 +128,32 @@ class RedirectUriValidator implements RedirectUriValidatorInterface
     {
         $parsedUrl = \parse_url($url);
 
-        return $parsedUrl !== false && (isset($parsedUrl['host']) || isset($parsedUrl['path']));
+        if ($parsedUrl === false) {
+            return false; // obviously invalid url
+        }
+
+        /**
+         * @see https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2
+         *
+         * The redirection endpoint URI
+         * - MUST be an absolute URI as defined by [RFC3986] Section 4.3.
+         * - MAY include an "application/x-www-form-urlencoded" formatted (per Appendix B) query
+         *   component ([RFC3986] Section 3.4), which MUST be retained when adding
+         *   additional query parameters.
+         * - MUST NOT include a fragment component.
+         *
+         * @see https://datatracker.ietf.org/doc/html/rfc3986#appendix-A
+         * absolute-URI  = scheme ":" hier-part [ "?" query ]
+         */
+
+        if (empty($parsedUrl['scheme']) || !empty($parsedUrl['fragment'])) {
+            return false;
+        }
+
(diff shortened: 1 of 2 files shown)


src/RedirectUriValidators/RedirectUriValidator.php
```diff
@@ -41,6 +41,11 @@ public function __construct($allowedRedirectUri)
      */
     public function validateRedirectUri($redirectUri)
     {
+        $redirectUri = \filter_var($redirectUri, FILTER_VALIDATE_URL);
```
Sephster, line 44: This doesn't only check for `http` schemas. It also allows `ftp`, `mailto` etc. I think we  need to be more specific with this check if possible.

Sephster (changes requested):

src/RedirectUriValidators/RedirectUriValidator.php
eugene-borovov: This is only the first stage of checking the URL to make sure that it is correct. Then the mapping is performed with the redirect URL. Without this check, an incorrect URL will cause an error in the schema checking code.

Sephster: Could you paste an example of the error you are getting? Thank you

eugene-borovov: I have got an error like this:
```
Undefined index: scheme
 /opt/project/src/RedirectUriValidators/RedirectUriValidator.php:69
 /opt/project/src/RedirectUriValidators/RedirectUriValidator.php:49
 /opt/project/src/Grant/AbstractGrant.php:288
 /opt/project/src/Grant/AuthCodeGrant.php:269
 /opt/project/tests/Grant/AuthCodeGrantTest.php:394
```

You can reproduce by changing `redirect_uri` in query parameter to `localhost` or something without scheme.
I used `\LeagueTests\Grant\AuthCodeGrantTest::testValidateAuthorizationRequestBadRedirectUriString`

eugene-borovov: > This doesn't only check for `http` schemas. It also allows `ftp`, `mailto` etc. I think we need to be more specific with this check if possible.

I added http schema validation. So only valid urls could be check or add as allowed redirect urls.

boesing:
Oh, haven't seen this before raising my issue.
This relates to #1239

Spomky:
URNs shall be accepted too.  See example at https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-desktop-app-registration

The URN `urn:ietf:wg:oauth:2.0:oob` won't pass this validation.

eugene-borovov:
I fixed the `isLoopbackUri` check and remove scheme restrictions. So URNs and private URI schemes and redirect path should pass. The host and/or path are required for the correct URL.

eugene-borovov:
https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2

> The redirection endpoint URI MUST be **an absolute URI** as defined by
   [[RFC3986] Section 4.3](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3).  The endpoint URI MAY include an
   "application/x-www-form-urlencoded" formatted (per Appendix B) query
   component ([RFC3986] Section 3.4), which MUST be retained when adding
   additional query parameters.  The endpoint URI MUST NOT include a
   fragment component.

So scheme is required.

Sephster:
I see Microsoft mentioning urns but don't see anything in the official specs about urn support. Just custom app schemes. We always follow the official RFCs so I'm not concerned with urns at this time but if someone can find info in an oauth RFC that specifies we should support urns, we will of course include this.

Spomky:
Actually, URNs are more likely used for native apps. 
See also the Google Documentation at https://developers.google.com/youtube/v3/live/guides/auth/installed-apps

eugene-borovov:
https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2

> The redirection endpoint URI MUST be an absolute URI as defined by [RFC3986] Section 4.3. 

https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.3

> A URI can be further classified as a locator, a name, or both.  The term "Uniform Resource Locator" (URL) refers to the subset of URIs that, in addition to identifying a resource, provide a means of locating the resource by describing its primary access mechanism (e.g., its network "location").  The term "Uniform Resource Name" (URN) has been used historically to refer to both URIs under the "urn" scheme [RFC2141], which are required to remain globally unique and persistent even when the resource ceases to exist or becomes unavailable, and to any other URI with the properties of a name.

https://datatracker.ietf.org/doc/html/rfc3986#section-3

>    The following are two example URIs and their component parts:

         foo://example.com:8042/over/there?name=ferret#nose
         \_/   \______________/\_________/ \_________/ \__/
          |           |            |            |        |
       scheme     authority       path        query   fragment
          |   _____________________|__
         / \ /                        \
         urn:example:animal:ferret:nose


So URNs can also be redirect URIs.

Sephster:
I'm not sure this is more common but would happily be proved wrong. As I said before though, I would really want to see something about URNs in the Oauth specs before committing to this. If it is an official implementation, recommended by the IETF then we should support this. Is there any spec out there specifying this?

Sephster:
Thanks @eugene-borovov and @Spomky for clarifying this

eugene-borovov:
By the way, the main purpose of this PR is to fix the `isLoopbackUri` method and add some verification for redirect URI. The URN discussion is a side effect of the too strict validation rules that I gave initially.

Spomky:
> [...] , I would really want to see something about URNs in the Oauth specs before committing to this.

The spec refers to URIs, which includes URLs and URNs right?
https://en.m.wikipedia.org/wiki/Uniform_Resource_Identifier

tests/RedirectUriValidators/RedirectUriValidatorTest.php
```diff
@@ -72,4 +72,52 @@ public function testValidIpv6LoopbackUri()
             'Loopback redirect URI can change the port number'
         );
     }
+
```
Spomky, line 75: These new tests could be simplified using a dataProvider

eugene-borovov: Good point. I added dataProvider. Have I listed all the cases? Maybe you know more?

Spomky: Excellent. what a quick update!

Spomky: Is there somewhere a test with a valid url e.g. `http://example.com:8080/endpoint?foo=BAR`

eugene-borovov: I have added more valid URLs for clarity.

Sephster:
Thank you for this PR @eugene-borovov and I'm sorry I didn't get a chance to push this through. Since this was submitted a newer PR was proposed (#1274) which I think provides a more elegant solution as it relies on the league/uri package to handle parsing. For this reason, I'm going to close this PR but wanted to thank you for trying to resolve the issue.

eugene-borovov:
@Sephster #1274 resolves only loopback URI. This PR resolves all redirect URI validation. These are different branches of the validate algorithm. This PR allows to use URN as redirect URI in addition.

Sephster:
thanks for the quick reply @eugene-borovov. I added in some further tests to main, based on yours, to ensure we are able to accept custom schemes and urns. I think that the solution that has been merged does ensure all valid uris are accepted because it is using the league/uri package. If an invalid uri is passed (poorly formed rather than not pre-registered), the library will throw an exception. Thank you again for your efforts here and sorry I didn't merge it this time.

eugene-borovov:
@Sephster I`d suggested to use URI validation in constructor. So on loopback analyze step you have valid URI and test only the host. Also this prevents using no-URI strings as redirect_uri as in the issue https://github.com/thephpleague/oauth2-server/issues/1239.

PR #1238, opened 2021-07-16
Simplifies `Bearer` regex

donatj:
`(?:\s+)?` is just `\s*` but harder


eugene-borovov:
Regular expressions are not needed here.

```php
$jwt = \trim($header[0] ?? '');
$jwt = \strpos($jwt, 'Bearer') === 0 ? \trim(\substr($jwt, 6)) : '';
```

Sephster:
Looks good to me. Thanks @donatj

PR #1242, opened 2021-08-05
Fix parse_url() may return FALSE case

dongm2ez:
On seriously malformed URLs, parse_url() may return false.

https://www.php.net/manual/en/function.parse-url.php


eugene-borovov:
Part of #1237

PR #1249, opened 2021-09-30
Avoid using broken `LocalFileReference` from lcobucci/jwt

chalasr:
It's deprecated since https://github.com/lcobucci/jwt/security/advisories/GHSA-7322-jrq4-x5hf


Sephster:
Looks good. Thanks @chalasr

PR #1250, opened 2021-10-13
Use thephpleague/oauth2-server-bundle for the Symfony integration

chalasr:
Hey,
The bundle has been tagged now and it already integrates the same features as the Trikoder bundle which is not actively maintained anymore.

I'm not sure if the content of https://oauth2.thephpleague.com/framework-integrations comes from this repo so it may need an update as well.

Fixes https://github.com/thephpleague/oauth2-server-bundle/issues/54


chalasr:
Great! Thanks a lot @Sephster.

PR #1251, opened 2021-10-17
Test against php 8.1

sergiy-petrov:


iansltx:
I'm seeing a deprecation error due to a missing return type on `jsonSerialize()` in `ScopeTrait`. Based on that trait's contents, we should be able to safely add the return typehint of `string` but it would be enforcing something that's only docblock'd at the moment so I figured I'd ask before just pushing a PR. For now the workaround in our codebase is to implement jsonSerialize() on our own in our scope entity rather than using the trait, which works for now in both 8.0 and 8.1.

tobias-trozowski:
@iansltx shouldn't it be ok if we just add `#[\ReturnTypeWillChange]` to the `ScopeTrait::jsonSerialize()`?

iansltx:
That would remove the warning, but we'll still need to change the return type eventually. Guessing we want to wait for another major version to do that since it's not enforced now?

Sephster:
I aim to take a look at this before the Christmas hols are over. Thanks for all your input guys and Merry Christmas

Mauricevb:
> Just to update here, I'm working on a release for version 9 which will fully support PHP 8.1. I would have loved to push this out the door asap but I think it is right I take time to release a proper major version.

First of all, thank you very much for this great library. 
I’m wondering whether version 9 (or another version with PHP 8.1 support) is still expected to be released. If not, it might be a idea to merge this PR?

Sephster:
Just realised this PR is still open but PHP 8.1 and 8.2 support has now been properly added with corresponding tests. Will close this accordingly. Many thanks

PR #1254, opened 2021-12-03
Docs. Fix namespace for UserEntityInterface

vommuan:


Sephster:
Thanks for spotting this!

PR #1255, opened 2021-12-11
Add introspection implementation - updated

kromacie:
Hello, 

There is a [PR](https://github.com/thephpleague/oauth2-server/pull/925) started by **Steve Porter** related to [RFC 7662](https://datatracker.ietf.org/doc/html/rfc7662.html) that wasn't updated for almost three years already. Based on the activity under the old thread and my personal willingness to use it, I think that there is still a demand for this feature. Therefore I decided to refurbish it a bit and refactor it slightly.

There is what I changed compared to the previous PR.
- I added authorization to use introspection, which should prevent tokens fishing. It's reused from _ResourceServerMiddleware_, which means, without a valid access token, you can't access the introspection endpoint without a valid access token.
- I separated introspection from **Authorization Server** to a dedicated **Introspection Server** because previously, it introduced many unrelated changes to existing logic. Also, I thought it would be better to keep it apart from **Resource Server** because it can be and often have to be separated in practice. For the compromise, I decided to put it somewhere in between.

Also, It's my very first contribution to an open-source project, so please feel free to mention any mistake or gap that I didn't find. I hope you will find it helpful.


hafezdivandari:
#1473

Sephster:
My apologies this was never progressed and thank you for your efforts. I think this has inspired the implementation in #1473 which is close to being merged. I'm going to close this in favour of that implementation but thank you very much for your efforts here

kromacie:
@Sephster , all good, inspiring anyone is more that I could expect in that case. Good luck guys :)

PR #1256, opened 2021-12-20
Update Laravel example to work when configuration is cached

Krisell:
The Laravel example of reading the application key doesn't work in a production setting where the configuration cache is used. All calls to `env()` return null then and `config()` should be used instead.

Using `config('app.key')` always works, cached or not. Of course a Laravel developer already knows this so the risk of it causing an actual problem is very low, but I think any Laravel example should follow Laravel conventions.

Laravel documentation about this
https://laravel.com/docs/8.x/deployment#optimizing-configuration-loading


Sephster:
Thanks for this @Krisell - happy to merge

PR #1257, opened 2021-12-23
Bug: GrantTypeInterface not respected

tobias-trozowski:
when enabling a grant type via `enableGrantType` the provided type `GrantTypeInterface` is not respected (`$grantType->revokeRefreshTokens($this->revokeRefreshTokens);`). 
A non-existent method will be called.

I just added unit test to test against a custom grant type


Sephster:
I think this is related to the changes in PR #1189. If I am understanding correctly, an update to the interface to add the `revokeRefreshTokens` method in the next major release should fix this. Can you confirm this is what you are expecting @tobias-trozowski?

It threw me a little that your PR isn't addressing this specific issue (but thank you all the same for submitting this).

tobias-trozowski:
@Sephster i wouldn't say "expect" but adding the method would do just fine. Thank you so much!

Sephster:
Thanks @tobias-trozowski  - the interface change will be added to the next major release. Thank you for the PR too

Issue #1266, opened 2022-03-01
revokeRefreshTokens isn't part of the GrantTypeInterface


reedy:
Renamed in d7634f92fcc2a4dc90a611987dc06579957b68d0, it seems in https://github.com/thephpleague/oauth2-server/commit/d7634f92fcc2a4dc90a611987dc06579957b68d0#diff-e3aa81140e7f04a25ca4314bf068da734e8d6b03c9ee7ef90825e65c292913c5R131-R144 `revokeRefreshTokens` (and `setRevokeRefreshTokens` before it) are used as they're part of the `GrantTypeInterface`, but the interface doesn't contain this function...

Sephster:
Issue has been noted and is being fixed as part of v9 release. Thank you for attempting a fix for this.

netniV:
I swore I posted a comment on this already but I can't seem to find it.  I was asking about the release plan for v9 or at least the possibility of submitting a fix to the existing v8 code by commenting out the line in `enableGrantType` so people don't have to modify vendor code to make things work?

PR #1267, opened 2022-03-01
GrantTypeInterface: Add revokeRefreshTokens

reedy:
Follows up d7634f92fcc2a4dc90a611987dc06579957b68d0

Ping https://github.com/thephpleague/oauth2-server/issues/1266


Sephster:
Thanks for this. We've already got a fix in place which will be released with the next major so I'm going to decline merging this for now. Thank you very much for trying to solve this though. It is much appreciated.

PR #1268, opened 2022-03-10
Update requirements to allow PHP 8

WouterSioen:
This is according to the allowed PHP versions in the composer.json file


Sephster:
Thank you for this but at this time, I still haven't finished my checks for 8.1 so don't want to proceed with this merge. I believe 8.1 works without issues but need to verify this before making it official.

PR #1269, opened 2022-03-10
PHP version requirement (docs)

reindert-vetter:
It is unclear which PHP versions are supported.

https://oauth2.thephpleague.com/requirements/:
> The following versions of PHP are supported:
> PHP 7.2
> PHP 7.3
> PHP 7.4

https://github.com/thephpleague/oauth2-server/blob/master/README.md:
> The latest version of this package supports the following versions of PHP:
> PHP 7.2
> PHP 7.3
> PHP 7.4
> PHP 8.0

https://github.com/thephpleague/oauth2-server/blob/master/composer.json:
> "require": {
>     "php": "^7.2 || ^8.0",

I assume we support both PHP 8.0 and 8.1. Maybe it would be a good idea to update that.

### DIFF
--- a/requirements.md
+++ b/requirements.md
@@ -15,6 +15,8 @@ The following versions of PHP are supported:
 * PHP 7.2
 * PHP 7.3
 * PHP 7.4
+* PHP 8.0
+* PHP 8.1
 
 The `openssl` and `json` extensions are also required.
 


Sephster:
PHP 8.0 should be added to the docs. Thanks for spotting this. 8.1 is not fully supported yet as I'm aware of some issues but for the most part, it should work

reindert-vetter:
Okay. What issues are known with php 8.1? I use that now.

Sephster:
The scope trait has a deprecation error thrown because we don't have a return type for jsonSerialize. That is the only thing I'm aware of at present so it is minor

PR #1270, opened 2022-03-10
PHP version requirement (readme)

reindert-vetter:
It is unclear which PHP versions are supported.

https://oauth2.thephpleague.com/requirements/:
> The following versions of PHP are supported:
> PHP 7.2
> PHP 7.3
> PHP 7.4

https://github.com/thephpleague/oauth2-server/blob/master/README.md:
> The latest version of this package supports the following versions of PHP:
> PHP 7.2
> PHP 7.3
> PHP 7.4
> PHP 8.0

https://github.com/thephpleague/oauth2-server/blob/master/composer.json:
> "require": {
>     "php": "^7.2 || ^8.0",

I assume we support both PHP 8.0 and 8.1. Maybe it would be a good idea to update that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ The latest version of this package supports the following versions of PHP:
 * PHP 7.3
 * PHP 7.4
 * PHP 8.0
+* PHP 8.1
 
 The `openssl` and `json` extensions are also required.
 


Sephster:
Thanks for this @reindert-vetter. I think we need to update the docs to list PHP 8.0. At this time, I believe PHP 8.1 works but haven't completed my testing of this yet. I plan on getting a new major out in the not too distant future which will have official support for 8.1. Thank you for the PR.

PR #1272, opened 2022-03-15
Prevent warning when malformed redirect URI is provided

axlon:
The `RedirectUriValidator` does not account for the fact that `parse_url()` may return `false`, when a consumer provides a malformed URL. The validator will attempt to retrieve the scheme from the parsed URL which at that point will be set to false causing a warning to be emitted (in my case immediately triggering a HTTP 500 response because of error handling).

This PR aims to fix this by checking if the parsed URL is false, in which case the validator exits early.


Sephster:
Thank you for this PR @axlon but we've just reviewed and merged a PR that removes our usage of parse_url so you shouldn't have this issue going forwards. I expect to release a patch update tomorrow. Thank you for raising this though

Issue #1273, opened 2022-03-16
Non-URL redirection URIs fail validation (e.g., for native apps with custom schemes)


bradjones1:
I have a native app which performs OAuth2 with, among other clients, a native app.

The redirection URI is something like `native.app.reverse.dns://`, which is a valid URI but is not a URL.

Spec references:
https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2
https://datatracker.ietf.org/doc/html/rfc3986#section-4.3

The culprit is `parse_url()`, which as the name implies is specific to URLs. Further FWIW the PHP docs [say](https://www.php.net/manual/en/function.parse-url.php):

> This function is not meant to validate the given URL, it only breaks it up into the parts listed below...

Which is more or less what we're doing here.

To be truly spec-compliant, we must allow and match against URIs, not just the subset which are URLs.

This being a League package, good thing there's https://github.com/thephpleague/uri, which advertises URI validation and manipulation. Let's use that.

Also this [affects the Drupal implementation](https://www.drupal.org/project/drupal/issues/3186651), since Drupal core assumes URIs are URLs, despite the various interfaces being called `UriWhatever`.

bradjones1:
Potentially related https://github.com/thephpleague/oauth2-server/issues/1039

bradjones1:
Refs https://github.com/thephpleague/oauth2-server/issues/1188#issuecomment-906264171, though that and its related PR still use `parse_url()` which I think is incorrect.

Sephster:
Thanks @bradjones1 and thanks for the PR.  I will take a look at this tomorrow

victorbalssa:
Hello, 
same problem here. 
I have a native application with `scheme://` as  `redirect_uri`,
Thank you @bradjones1 💪🏻

PR #1274, opened 2022-03-16
Use league/uri for true URI validation

bradjones1:
Fixes https://github.com/thephpleague/oauth2-server/issues/1273

### DIFF
--- a/src/RedirectUriValidators/RedirectUriValidator.php
+++ b/src/RedirectUriValidators/RedirectUriValidator.php
@@ -110,6 +110,6 @@ class RedirectUriValidator implements RedirectUriValidatorInterface
     {
         $uri = Uri::createFromString($url);
 
-        return (string) $uri->withPort(NULL);
+        return (string) $uri->withPort(null);
     }
 }
--- a/src/RedirectUriValidators/RedirectUriValidator.php
+++ b/src/RedirectUriValidators/RedirectUriValidator.php
@@ -9,6 +9,8 @@
 
 namespace League\OAuth2\Server\RedirectUriValidators;
 
+use League\Uri\Uri;
+
 class RedirectUriValidator implements RedirectUriValidatorInterface
 {
     /**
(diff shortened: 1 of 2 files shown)


bradjones1:
I think this is a bit more elegant than https://github.com/thephpleague/oauth2-server/pull/1237 - though that does include some test coverage for URIs, so we should include that here too.

Speaking of tests, it doesn't appear the failures on this PR seem to be related to my changes... but I'll have to dig deeper.

src/RedirectUriValidators/RedirectUriValidator.php
```diff
@@ -59,10 +61,10 @@ public function validateRedirectUri($redirectUri)
      */
     private function isLoopbackUri($redirectUri)
     {
-        $parsedUrl = \parse_url($redirectUri);
+        $uri = Uri::createFromString($redirectUri);
 
-        return $parsedUrl['scheme'] === 'http'
-            && (\in_array($parsedUrl['host'], ['127.0.0.1', '[::1]'], true));
+        return $uri->getScheme() === 'http'
```
Spomky, line 66: Why only `http`?
`https` could also be used on these loopback URIs?
I think only the host should be checked.

bradjones1: I agree - though I was trying to keep my changes atomic to the URI parsing so as to not change too much existing behavior. That said, I do think we could drop this comparison.

bradjones1:
Tests are failing wth
```
Error: Call to undefined method LeagueTests\Stubs\GrantType::revokeRefreshTokens()
```

Doesn't feel related to this change but the test setup.

Sephster:
No this is an ongoing issue I've not bothered to resolve as will be fixed in the next major. Happy to ignore it. Sorry for any confusion there

src/RedirectUriValidators/RedirectUriValidator.php
Sephster: In RFC8252 it says:

> Loopback redirect URIs use the "http" scheme and are constructed with the loopback IP literal and whatever port the client is listening on.

I'm assuming we've just put this check in to be compliant with the RFC. As to whether the RFC is correct on this point, I'm not sure so am happy to leave as is for now unless someone can say otherwise

Sephster:
Thanks @bradjones1 - this looks great and I'm embarrassed to say it didn't occur to me to use this library. Great find and thanks for the PR

PR #1282, opened 2022-04-19
Fix compatibility with lcobucci/jwt ^4.2

chalasr:
Hey, 
Here is a forward compatibility layer for lcobucci ^4.2 and specifically this change https://github.com/lcobucci/jwt/pull/833 which forbids creating `InMemory` keys with empty strings.
Spotted in https://github.com/thephpleague/oauth2-server-bundle/runs/6077089570?check_suite_focus=true

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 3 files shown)


Sephster:
Sorry for the delay on checking this. I had to work out why I'd used the assymetric signer for signing and the symmetric signer for decoding. It turns out it was to avoid a breaking change. Long term I want to modify the lib so users can pass their own config but for now, will need to stick with this. Thanks for spotting this

micronax:
Hi @Sephster, is there a planned release with this fix?

Sephster:
Likely in the next 48 hours 👍

PR #1298, opened 2022-09-08
Fix "state" parameter of error responses

hafezdivandari:
According to [RFC 6749](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2.1) on error response of the implicit grant, the authorization server [should add parameters to the fragment component of the redirection URI](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2:~:text=by%20adding%20the%20following%0A%20%20%20parameters%20to%20the%20fragment%20component%20of%20the%20redirection%20URI) and [the `state` parameter is required](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2:~:text=5B%20/%20%25x5D%2D7E.-,state%0A%20%20%20%20%20%20%20%20%20REQUIRED%20if%20a%20%22state%22%20parameter%20was%20present%20in%20the%20client%0A%20%20%20%20%20%20%20%20%20authorization%20request.%20%20The%20exact%20value%20received%20from%20the%0A%20%20%20%20%20%20%20%20%20client.,-For%20example%2C%20the) if it was present on the request.

There already 2 error responses that have redirect URI:
* `Invalid_scope`:
   * The error parameters were already on fragment component when using implicit grant as expected, but this error response doesn't include the `state` parameter when redirecting. This PR fixes this on both 'auth code' and 'implicit' grants.
    * Auth code grant:
       * before: `http://example.com/callback?error=access_denied&error_description=...`
       * after:   `http://example.com/callback?state=123&error=access_denied&error_description=...`
    * Implicit grant:
       * before: `http://example.com/callback#error=access_denied&error_description=...`
       * after: `http://example.com/callback#state=123&error=access_denied&error_description=...`
* `access_denied`:  This error response does include the `state` parameter when redirecting as expected, but when using implicit grant, the error's `state` parameter is on query string and other parameters were on fragment component. This PR fixes this.
  * Before: `http://example.com/callback?state=123#error=access_denied&error_description=...`
  * After: `http://example.com/callback#state=123&error=access_denied&error_description=...`


Sephster:
Please can you provide some information about why you want to add this change? Thank you

hafezdivandari:
Sorry I didn't add any description because I thought it is clear.

According to [RFC 6749](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2.1) on error response of the implicit grant, the authorization server should add parameters to the fragment component of the redirection URI.

It's already done for access token response [on this line](https://github.com/thephpleague/oauth2-server/blob/20252ac5251c199e860f15f78730393a00dcc579/src/Grant/ImplicitGrant.php#L214), But not for error response. this PR fixes this by passing `$this->queryDelimiter` to `$this->makeRedirectUri` function.

hafezdivandari:
Just merged `master` into this and resolved conflicts.

Sephster:
Thanks for this @hafezdivandari - the code looks good. I was wondering if you would be up for adding some tests to ensure we don't add any regressions for this in future. Because I've taken so long to get around to this, if you don't have time I will try to add them in myself and we can then get this merged. If it is also possible to add a changelog entry that would be appreciated but I can do this myself if you don't have time. Thanks so much for this!

hafezdivandari:
@Sephster just added some tests and a changelog entry.

Sephster:
Thanks for this @hafezdivandari  - looks great

PR #1299, opened 2022-09-15
Use 2048 bit test keys for compatibility with lcobucci/jwt >= 4.2.1.

EreMaijala:
It seems that https://github.com/lcobucci/jwt/commit/5a28a406a4d943a9237b2080ad16c1d37f12c05d changed the minimum RSA key length to 2048, which broke tests with version 4.2.1.


Sephster:
Thanks for spotting this!

PR #1300, opened 2022-09-15
Fix PHP 8.1 deprecations.

EreMaijala:
- Fixes a couple of cases where null was passed instead of string to `AbstractAuthorizeGrant::makeRedirectUri()` by ensuring that the redirect uri is always a string.
- Adds `#[\ReturnTypeWillChange]` attributes to `jsonSerialize` methods. This was mentioned in #1251 and is indeed the only solution when PHP 7.x compatibility needs to be maintained, because the pseudo-type `mixed` is only available from PHP 8.0 onwards.


Sephster:
Thank you for your PR. Can I just check, in which situation do you have a null value being sent? It should be the case that we either use the default redirect uri registered with your client or another redirect uri passed in at call time which has been pre-registered.

If all has been set up properly, I don't think we should ever have an instance where no redirect uri can be retrieved. If this is the case, I think it is probably an implementation error which we might still need to guard against. Thank you for raising this

EreMaijala:
@Sephster I'm not sure if it can happen in real life when things are properly set up, but it happens during unit tests. Perhaps the tests are incomplete, but if they trigger the case without throwing somewhere else, I suppose it's possible to trigger the case also in real life. I also checked that e.g. in `AuthorizationServerTest::testCompleteAuthorizationRequest` it's a real instance and not because it was insufficiently mocked or something.

Sephster:
Thank you @EreMaijala. So did you notice the issue when running tests against PHP 8.1? Thanks for the quick reply

EreMaijala:
@Sephster Yes. Passing null to `strstr` (AbstractAuthorizeGrant::makeRedirectUri() calls it) was fine before, but now logs a deprecation notice in PHP 8.1.

EreMaijala:
More details about the deprecations are available in the [PHP 8.1 migration guide](https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.null-not-nullable-internal).

EreMaijala:
@Sephster Would it be possible to get this merged, or is there something more you expect from me? Since PHP 7.4 is already EOL'd, we're seeing more and more of deprecation warnings, and while they don't affect functionality or production use, they come up during tests and development, and cause some confusion.

demiankatz:
@Sephster, just checking in on the status of this issue -- I'd love to see a resolution since this is currently adding noise to our test output. If there's anything I can do to help, please let me know!

Sephster:
Have created PR #1329 which I think will address the issues. In a real world application, `validateAuthorizationRequest` and `completeAuthorizationRequest` should be called in tandem and the former prevents a null uri from passing. Because of this, I've just updated the test cases to pass a dummy uri which fixes the issue as the test cases are by passing the `validateAuthorizaitionRequest` method for brevity.

Thank you very much for highlighting this issue and the effort spent on subbing a PR.

EreMaijala:
I'll leave this open until `#[\ReturnTypeWillChange]` is added to ScopeTrait.php as well in #1329 or elsewhere.

Sephster:
Resolved in #1329 so marking as closed. Thank you for alerting me to this and your suggested PR

PR #1308, opened 2022-10-28
[UPGRADE] Upgrade league/event to version 3.

frankdejonge:
This PR upgrades League\Event to version 3 while making the upgrade process smooth. I've added compatibility layers in to make the upgrade process smooth for users. Still, because return types are different, these changes are technically BC breaks.


Sephster:
Thanks Frank. Will get to this tomorrow evening 🙏

irazasyed:
Can this be merged please?

ganey:
Hey, any updates on this? I'm already using league/event@^3 but now need to add oauth, currently i'll have to downgrade event to make it work

Issue #1287, opened 2022-07-18
BearerTokenValidator::validateAuthorization requires the JWT to have the 'Not Before' claim


goldi80:
With [commit 07bdaeb](https://github.com/thephpleague/oauth2-server/commit/07bdaebb5d63ded6e227da9ba4ad72473fc4cd62) the deprecated Constraint `Lcobucci\JWT\Validation\Constraint\ValidAt` is replaced with the newer `Lcobucci\JWT\Validation\Constraint\StrictValidAt`. This replacement broke our authentication as our tokens (provided by a third party system) do not contain the optional claim 'Not Before' ([RFC 7519 - JSON Web Token (JWT) -- nbf](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5)).

In fact, the correct replacement for the deprecated constraint is `Lcobucci\JWT\Validation\Constraint\LooseValidAt` which also does not require the nbf claim.

Unfortunately, there seems to be no obvious way how to configure which constraints should be checked by `BearerTokenValidator`.

Sephster:
Hello. Thanks for reporting this and sorry for the delayed response. When I replaced this I checked and our default access token issues an nbf claim so I didn't expect this to be an issue sorry.

Can I ask why you aren't issuing the tokens with this library or why you can't add the nbf claim in to your third party provider so I can better understand the issue.

The easiest way for you to fix this would be to provide your own BearerTokenValidator implementation. You can pass this in to the constructor of your ResourceServer - as long as it implements the AuthorizationValidator interface, it will be accepted. Hope this helps and sorry for the delay in getting back to you

goldi80:
Thanks for your reply. We're not creating the tokens with this library because the system using it to validate the tokens is only a small building block in the ecosystem of our company. There are many different systems validating tokens issued by a central system. The colleagues managing the token creation service are of the opinion, that an optional thing shouldn't be mandatory.

In fact, passing in our own token validator is the 'workaround' we've implemented right now. But to honest, we want to get rid of this customization.

Sephster:
I've checked the docs and looks like looseValidAt will check nbf if it is present so that seems to be the best of both worlds. Users can still be assured nbf is checked if the token is issued via the library so happy to make this adjustment. Will try and get this implemented this week. Thanks for flagging it!

goldi80:
Thanks for adjusting this. I'm really looking forward to receive this update.

Sephster:
[...]

PR #1312, opened 2022-11-14
Use LooseValidAt instead of StrictValidAt

Sephster:
This change fixes issue #1287, which reported the clients were being forced to set the nbf claim.

By using LooseValidAt, we will check this claim if it is present but not reject the token if it is not.


parallels999:
how about private claims?

PR #1313, opened 2022-11-16
Update the Param Type on the Validate Scopes Method

chrisguitarguy:
`validateScopes` does accept and handle `null` values.

Hit this while writing a new grant type and PHPStan complained. Since `validateScopes` does handle `null` appropriately, just updating the type in PHPDoc


Sephster:
Thanks for this. I've fixed it in the PHP 8 compat branch but happy to merge in early here. Thanks for spotting this and submitting a PR

PR #1316, opened 2022-11-25
Added OpenID Responce Type

marcriemer:


doobry-systemli:
I'm glad to see some progress here! @marcriemer do you intend to continue working on it? Did you get stuck somewhere particularly?

marcriemer:
@doobry-systemli So far my implementation works very well. I will continue my work on in this feature. Please let me know If you have any successions for improvements.

Sephster:
Thanks for this PR. I am a bit stumped as to why the tests aren't running in this PR. Everything looks ok to me. Do you have any ideas? If not, I will put in a GH support request as I have a vague recollection something similar has happened in the past

Sephster:
> Thanks for this PR. I am a bit stumped as to why the tests aren't running in this PR. Everything looks ok to me. Do you have any ideas? If not, I will put in a GH support request as I have a vague recollection something similar has happened in the past

And now they are running... odd. Must have been some weird delay. Apologies for taking so long to pick this up. Are you able to take a look at the failing tests? Thank you

fruitwasp (commented):
keep up the great work! Thanks for this PR!

src/ClaimExtractorIntercace.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace League\OAuth2\Server;
+
+interface ClaimExtractorIntercace
```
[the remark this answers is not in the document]
marcriemer: Thanks, will fix it soon.

pat0s:
Hi @marcriemer,
I would like to ask one question. I got a little bit lost. I think it's not a mistake, but it is confusing.

```php
interface ClaimSetRepositoryInterface
{
    /**
     * Get ClaimSetEntries
     *
     * @param AccessTokenEntityInterface $authCode
     *
     * @return ClaimSetInterface
     */
    public function getClaimSetEntry(AccessTokenEntityInterface $authCode): ClaimSetInterface;
}
```

```php
$claimSet = $this->claimRepository->getClaimSetEntry($accessToken);

if ($claimSet instanceof ClaimSetInterface) {
            foreach ($this->extractor->extract($accessToken->getScopes(), $claimSet->getClaims()) as $claimName => $claimValue) {
                $builder->withClaim($claimName, $claimValue);
            }
        }
```

`getClaimSetEntry()` sounds like I should return `ClaimSetEntry`. `ClaimSetEntry` also implements `ClaimSetEntryInterface`, which means new method `getScope()`. I think it is irrelevant in this case. It seems like I should return only `ClaimSet` from this  repository.

`ClaimSet` is new Entity which I created. It implements only `ClaimSetInterface` and stores all claims with values (from all scopes) in one array. Entity is created according to method's return type `ClaimSetInterface`.

So in my case method `getClaimSetEntry()` returns `new ClaimSet(['email' => 'email@example', 'email_verified' => 1, ...])`

abin2hats:
Hi @marcriemer 

Thank you for the initiative.

Why this PR is not merged even if the PR is reviewed?

Is there any plan to merge this PR?

Thank you.

Sephster:
Version 9 bringing in the device code is being worked on just now. Then I will start to look at other features. V9 RC1 should be tagged today all being well

marcriemer:
@pat0s Thanks for bringing this up. You are right, the method should return a ClaimSetEntryInterface. 

The ClaimSetEntryInterface extends ClaimSetInterface.

src/ResponseTypes/IdTokenResponse.php
```diff
@@ -0,0 +1,124 @@
...
+        if (!self::isOpenIDRequest($accessToken->getScopes())) {
+            return [];
+        }
+
+        $claimSet = $this->claimRepository->getClaimSetEntry($accessToken);
+
+        $builder = $this->idTokenRepository->getBuilder($accessToken);
+
+        if ($claimSet instanceof ClaimSetEntryInterface) {
+            foreach ($this->extractor->extract($accessToken->getScopes(), $claimSet->getClaims()) as $claimName => $claimValue) {
```
stof, line 84: `getClaims` is documented as returning `string[]` while the extractor expects `array<string, string>`. This looks inconsistent to me.

stof:
@marcriemer @Sephster what is the status of this work ?

examples/src/Repositories/IdTokenRepository.php
```diff
@@ -0,0 +1,43 @@
...
+    public function getBuilder(AccessTokenEntityInterface $accessToken): Builder
+    {
+        $builder = (new Builder(new JoseEncoder(), ChainedFormatter::withUnixTimestampDates()))
+            ->permittedFor($accessToken->getClient()->getIdentifier())
+            ->issuedBy($this->issuedBy)
+            ->issuedAt(new DateTimeImmutable())
+            ->expiresAt($accessToken->getExpiryDateTime())
+            ->relatedTo($accessToken->getUserIdentifier());
+
+        if ($this->nonce) {
```
stof, line 37: shouldn't the nonce come from the authorization request ? Configuring as a constructor argument of the repository is probably not usable. It would be great to have an actual example of the openid setup.

src/ResponseTypes/IdTokenResponse.php
stof: To me, this looks like we need 2 *separate* interface: one returning the map of claim names to claim values (used by the ClaimRepository when getting the claims for an access token) and one listing a set of claim names associated with a scope.

Even if both interfaces have a method named `getClaims`, the would still be separate interface due to different signatures.

stof: btw, rebasing this PR would probably make the CI red because phpstan should be complaining about that.

examples/src/Repositories/IdTokenRepository.php
marcriemer: Agree, the example is far from being perfect. The nonce must be part of the AuthCodeEntityInterface and used by ClaimSetRepositoryInterface to create ClaimSetInterface with the nonce claim.

marcriemer:
@stof I would love to contribute more, but some commits are sitting on the sideline right now. My complete OIDC implementation is nearly finished, including ImplicitOpenidGrant, AuthCodeOpenidGrant, and AbstractAuthorizeGrant along with all the related response types. However, it will require significant changes to this project.

@Sephster It may make sense to create a LTS branch and move faster forward for those who require new features. Authentication is changing fast right now. We need to keep up.

src/ResponseTypes/IdTokenResponse.php
marcriemer: Well, the ClaimExtractor needs to be fixed. It extracts ClaimSetEntryInterface, therefore most vars and properties are completely off.

marcriemer: Implementation of a `ClaimSetInterface` used by the `ClaimExtractor` could look like this:

``` php
    /**
     * Get users claims
     * 
     * @return array<string, string>
     */
    public function getClaims(): array
    {
        return [
            'preferred_username' => $this->getUsername(),
            'email' => $this->getEmail(),
        ];
    }
```

stof: but ClaimSetEntryInterface currently **extends** ClaimSet. this is what looks wrong to me as it seems like a separate concern (on one side, we associate a scope with a *set of claim names*, and on the other side, we have a map of claim names to claim values, which is not a set at all but a map)

marcriemer: The ClaimExtractor was confusing: https://github.com/marcriemer/oauth2-server/commit/b1400a1fe0b09d8978270c4b94ef63a0331d0f24

examples/src/Repositories/IdTokenRepository.php
stof: the nonce indeed needs to be part of the AuthCodeEntityInterface. But setting it in the AuthCodeEntityInterface should also come from the authorization request, not from a configured nonce there either when issuing the auth code (which requires adding dedicated support for it).

floxblah:
Just checking if there's still a chance this pull request might be merged ?

src/ResponseTypes/IdTokenResponse.php
stof: The linked commit still does not fix the fact that ClaimSetEntryInterface extends ClaimSet while they have different needs for the precise type of the array returned by `getClaims`.

Btw, updating the interface to actually document the precise type would force you to solve this issue

jmalinens:
@marcriemer  it looks like https://github.com/ronvanderheijden/openid-connect does similar thing as your MR

PR #1331, opened 2023-03-09
Allow lcobucci/jwt v5

maxhelias:


Sephster:
Thanks but I will decline this just now as plan on making more definitive moves to the latest version of the jwt lib to support the features fully. Sorry I'm not going to merge this just now

PR #1243, opened 2021-08-15
Added exception chaining in BearerTokenValidator

rhertogh:
Added exception chaining for RequiredConstraintsViolated in `\League\OAuth2\Server\AuthorizationValidators\BearerTokenValidator::validateAuthorization()`.

This helps with logging the exception (similar to line 103).

PR #1335, opened 2023-04-03
include caught exception when throwing OAuthServerException

dungsaga:
When throwing OAuthServerException, the caught exception should be included as a parameter (named $previous). The would help with debug and allow showing customized message for each kind of validation error


Sephster:
Closing in favour of #1243

PR #1336, opened 2023-04-03
Tweaked version constraints

GrahamCampbell:
The important change is the PHP version. It is too late to apply an upper-bound in a major release series - composer will just go and get v8.4.1 instead of v8.5.0 which is definitely worse. ;)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,15 +4,15 @@
     "homepage": "https://oauth2.thephpleague.com/",
     "license": "MIT",
     "require": {
-        "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
+        "php": "^8.0",
         "ext-openssl": "*",
         "league/event": "^2.2",
         "league/uri": "^6.7",
-        "lcobucci/jwt": "^4.3.0 || ^5.0.0",
+        "lcobucci/jwt": "^4.3 || ^5.0",
         "psr/http-message": "^1.0.1",
-        "defuse/php-encryption": "^2.3.0",
+        "defuse/php-encryption": "^2.3",
         "ext-json": "*",
-        "lcobucci/clock": "^2.2.0 || ^3.1.0"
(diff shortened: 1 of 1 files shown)


parallels999:
is this for fixing this downgrading?
```
Package operations: 1 install, 2 updates, 0 removals
  - Installing stella-maris/clock (0.1.7): Extracting archive
  - Downgrading lcobucci/clock (3.0.0 => 2.3.0): Extracting archive
  - Upgrading league/oauth2-server (8.4.1 => 8.5.0): Extracting archive
```

GrahamCampbell:
@parallels999 No that is not related. Can you run `composer why lcobucci/clock` please, so we can see why it couldn't install `lcobucci/clock` 3.0.0?

parallels999:
```php
composer why lcobucci/clock
lcobucci/jwt         4.3.0 requires lcobucci/clock (^2.0 || ^3.0)     
league/oauth2-server 8.5.0 requires lcobucci/clock (^2.2.0 || ^3.1.0) 

```

GrahamCampbell:
Ohh, it's because `lcobucci/clock 3.1.0` requires PHP 8.2. We should probably change the version constraint in this package to allow 3.0 as well as 3.1.

parallels999:
You're right, I'm using PHP 8.1, I don't plan on upgrading to 8.2 yet

Sephster:
Thanks @GrahamCampbell

Mauricevb:
I just ran `composer update` on my machine with PHP 8.1.13. Composer doesn't show any errors, but when I open the project I see the following error: 

`Fatal error: Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 8.2.0". You are running 8.1.13. in <path>`

It's still using Clock version `3.1.0`. I thought this patch would fix that issue? Am I doing something wrong?

PR #1337, opened 2023-04-11
Pass along previous exception

ash-m:
The current message is ambiguous on it's own. Supplementing the exception with the previously caught exception helps debug _which_ constraint was violated.


hafezdivandari:
Duplicate of #1243

ash-m:
@hafezdivandari ah snap. Thank you for replying. I'll close it.

PR #1338, opened 2023-04-16
Added auth_code event

tjveldhuizen:
Like events are emitted (#1211) when the access and refresh tokens are created, I've added an event when the auth_code is created.

Using this event, I am able to add state to the auth_code, which I need to generate the correct access token, later

### DIFF
--- a/src/AuthCodeEvent.php
+++ b/src/AuthCodeEvent.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * @author      Alex Bilbie <hello@alexbilbie.com>
+ * @copyright   Copyright (c) Alex Bilbie
+ * @license     http://mit-license.org/
+ *
+ * @link        https://github.com/thephpleague/oauth2-server
+ */
+
+namespace League\OAuth2\Server;
+
+use League\Event\Event;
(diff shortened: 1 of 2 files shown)


Sephster:
Apologies for the delay in picking this up. The state is passed back via an http query value along with the auth code. Why do you need to include it in the auth_code specifically? Thanks

tjveldhuizen:
Hmm, this requires some digging in my memory😅

At least I want to log keep track of the completed log in sessions of the user, therefore I create an entry in my session log table when the access token is provided. In that entry, I want to log the IP address of the device of the authenticating user (which is not available in the token request, since that's instantiated by the client and not the browser of the user).

Next to that, I want to add extra data to the token when the user is logged in using 2FA.

Sephster:
Thank you. I haven't seen anyone asking for this but I can't see why we don't support this when we do similar events for other issued codes/tokens. 

Are you willing to provide some tests for this PR? It would be great to have some in prior to merging. Thanks for getting back to me so quickly with the above :+1:

tjveldhuizen:
I'll see what I can do!

PR #1340, opened 2023-04-19
Use PSR-14 Event Dispatcher instead league/event

MilesChou:
Using the PSR event dispatch interface to achieve greater flexibility.


MilesChou:
See the same PR #1184, close mine.

PR #1352, opened 2023-07-05
Fix inconsistent var

AurelienPillevesse:
Fix inconsistent ```@var``` based on ```@param``` and ```@return``` in the same class


Sephster:
I think this is never not a string from memory. Folks might set it as a numeric value but I think because it is sent over http it is always a string. Have I made a mistake here? Cheers

AurelienPillevesse:
@Sephster I don't have everything you have in mind but in my point of view, the setter and getter are mixed so the var have to be mixed too. Currently, when I use this Trait, it has to be a string

PR #1359, opened 2023-07-10
Remove Key Leak 8.4.x

jeffhuys:
Could we consider a 8.4.x branch for security updates, so this can be maintained separately?

Running into issues with Snyk with oauth2-server-bundle, and am bound to PHP 7.x.
I'm porting over a commit from this PR: https://github.com/thephpleague/oauth2-server/pull/1353


MauricioKruijer:
I need this for my symfony 5.4 project too, this would help me getting rid of snyk issues

PR #1360, opened 2023-07-12
Update PHP version requirements

svenk-ct:
Make clear which PHP versions are supported by which library versions.


Sephster:
We only support the later versions so I don't think this distinction is required. Thank you for your PR though and sorry I'm not merging it this time

PR #1328, opened 2023-02-20
Allow configuring the JWT builder when generating a token

hafezdivandari:
Fixes #885 
Closes #851 
Closes #1122 
Related to #1434

Currently we have to override the whole `convertToJWT` method to configure JWT (e.g. to add additional claims). This PR makes it possible to override the new `withJwtBuilder` method instead:

```php
// \League\OAuth2\Server\Entities\Traits\AccessTokenTrait

use Lcobucci\JWT\Builder;

/**
 * Configure the JWT builder instance.
 */
protected function withJwtBuilder(Builder $builder)
{
    return $builder->withClaim('foo', 'bar');
}
```

Update: This PR also adds `withValidatedRequest` method to the bearer token validator that makes it possible to configure the request and append additional attributes:

```php
// \League\OAuth2\Server\AuthorizationValidators\BearerTokenValidator

use Lcobucci\JWT\UnencryptedToken;
use Psr\Http\Message\ServerRequestInterface;

/**
 * Configure the validated authorization request instance.
 */
protected function withValidatedRequest(ServerRequestInterface $request, UnencryptedToken $token): ServerRequestInterface
{
    return $request->withAttribute('oauth_foo', $token->claims()->get('foo'));
}
```

PR #1363, opened 2023-08-02
Allow appending custom jwt token data to the request

parallels999:
Currently, it's easy to implement the `convertToJWT` method in the `AccessTokenEntityInterface` implementation to add data to the JWT. However, i don't see any solution, except creating a new implementation of `BearerTokenValidator`, to add additional attributes from the token, to the returning `ServerRequestInterface` object.

I think this PR can provide a way to simply override the `BearerTokenValidator::appendAttributesFromClaims()` method and pass the custom `BearerTokenValidator` instance to the `ResourceServer` instance.

Complement of #1328
Alternative of #851


parallels999:
Closed because #1328  added the same now

Issue #1347, opened 2023-06-02
RefreshTokenGrant should not revoke the access token


Starfox64:
In the current implementation of the `refresh_token` grant, the access token associated with the old refresh token is revoked

https://github.com/thephpleague/oauth2-server/blob/591a6311a591b4a41d5e8e7856c6f3c7b64a7c86/src/Grant/RefreshTokenGrant.php#L67

The RFC however [makes no mention](https://www.rfc-editor.org/rfc/rfc6749#section-6) of this being a requirement, only that the **refresh token** MAY be revoked which is something that is now configurable.

Sephster:
Thanks for raising this. We already have an issue for this in #946 so closing accordingly to avoid duplicates

Starfox64:
Unless I'm missing something, that issue doesn't seem related to this.
I'm not talking about having no expiry on the access token just that they should not be revoked early when using the refresh token grant and should instead be allowed to expire naturally.

Starfox64:
@Sephster Sorry for the ping, I don't know if you saw my last message but I think what you mentioned is a different issue.

Sephster:
Sorry @Starfox64 - the reason I've linked them is that currently, if your access token is revoked, you can't use your refresh token as that has also been revoked or expired. The issue I linked to would resolve this by having indefinite refresh tokens so you aren't stuck in limbo.

I get your point though. It won't be directly solving what you've raised here. What would you like to happen in the library? It seems safer to just revoke the access token to me from a security point of view.

Starfox64:
In my case I would still very much like to keep a TTL on the refresh token.

I'm asking this mostly for spec compliance since some clients will expect to be able to reuse access tokens until they expire but it's true that this isn't a problem for most people.

Problems arise however in distributed systems where multiple requests can be in flight which will cause 401 errors when a refresh happens.

In the interest of providing added security on top of the spec we could leave this as is but add a `shouldRevokeAccessToken` attribute to the refresh grant to leave the door open for those that need to strictly follow the spec.

PR #1377, opened 2023-09-29
RefreshTokenGrant: add option whether to revoke refreshed access tokens

josiasmontag:
Currently, the `RefreshTokenGrant` immediately revokes an access token when it gets refreshed.

The [RFC Section 6](https://www.rfc-editor.org/rfc/rfc6749#section-6) makes no mention that this should happen. 

The current behavior sometimes causes issues: Some clients assume the old access token is still valid because it has not reached its expiration date yet. Also, there are race conditions with simultaneous requests when one client refreshes the token and the other client still uses the old, non-expired token.

This PR adds an option `revokeRefreshedAccessTokens` to configure revoking old access token after refreshing. It defaults to `true`, wich is the current behaviour. 

Also see #1347

### DIFF
--- a/src/Grant/RefreshTokenGrant.php
+++ b/src/Grant/RefreshTokenGrant.php
@@ -64,7 +64,9 @@ class RefreshTokenGrant extends AbstractGrant
         }
 
         // Expire old tokens
-        $this->accessTokenRepository->revokeAccessToken($oldRefreshToken['access_token_id']);
+        if ($this->revokeRefreshedAccessTokens) {
+            $this->accessTokenRepository->revokeAccessToken($oldRefreshToken['access_token_id']);
+        }
         if ($this->revokeRefreshTokens) {
             $this->refreshTokenRepository->revokeRefreshToken($oldRefreshToken['refresh_token_id']);
         }
--- a/src/AuthorizationServer.php
+++ b/src/AuthorizationServer.php
@@ -85,6 +85,11 @@ class AuthorizationServer implements EmitterAwareInterface
     private $revokeRefreshTokens = true;
 
     /**
+     * @var bool
+     */
+    private $revokeRefreshedAccessTokens = true;
+
+    /**
      * New server instance.
      *
      * @param ClientRepositoryInterface      $clientRepository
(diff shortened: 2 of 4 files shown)


josiasmontag:
Some additional thoughts:
The current behaviour of revoking the access tokens might even be wrong according to the RFC. [Section 1.5](https://www.rfc-editor.org/rfc/rfc6749#section-1.5) says:

> Refresh tokens are issued to the client by the authorization server and are
   used to obtain a new access token when the current access token
   becomes invalid or expires, **or to obtain additional access tokens
   with identical or narrower scope** (access tokens may have a shorter
   lifetime and fewer permissions than authorized by the resource
   owner).

The current implementation does not allow to obtain **additional** access tokens as it revokes all previous ones.

Sephster:
We already have a function for this called `revokeRefreshTokens` which can be set to true or false. We used to automatically revoke and appreciate this isn't following spec so added in a method to allow implementers to choose whether this is enforced or not.

The boolean should be set in the AuthServer class. Thank you for your PR though. I hope this functionality helps solve your problem.

josiasmontag:
@Sephster I am aware of `revokeRefreshTokens` which is about revoking the old **refresh token**, it does not control whether the old **access token** is revoked. This PR is about adding a setting for revoking the previous access token. Please have a look at the relevant change [here](https://github.com/thephpleague/oauth2-server/pull/1377/files#diff-f985b69380156eff8040ab4d4bc953756e5e3d64328298126aa195c03121b5c5).

Sephster:
Apologies I had missed this. Will reopen to review later. Thanks for clarifying.

pat0s:
Is there any progress on clarifying this PR? @Sephster

Sephster:
No not yet. All my efforts are on releasing v9 then this will be picked up along with others. Cheers

hafezdivandari (approved):

src/Grant/RefreshTokenGrant.php
```diff
@@ -64,7 +64,9 @@ public function respondToAccessTokenRequest(
         }
 
         // Expire old tokens
-        $this->accessTokenRepository->revokeAccessToken($oldRefreshToken['access_token_id']);
+        if ($this->revokeRefreshedAccessTokens) {
```
hafezdivandari, line 67:
```suggestion
        if ($this->revokeRefreshTokens || $this->revokeRefreshedAccessTokens) {
```

hafezdivandari (changes requested):
[RFC7009 section 2.1](https://datatracker.ietf.org/doc/html/rfc7009#:~:text=If%20the%20particular,token%20as%20well.):

> If the particular token is a refresh token and the authorization server supports the revocation of access tokens, then the authorization server SHOULD also invalidate all access tokens based on the same authorization grant. If the token passed to the request is an access token, the server MAY revoke the respective refresh token as well.

It means:
* When revoking a refresh token, the access token SHOULD also be revoked.
* When revoking an access token, the refresh token MAY also be revoked.

PR #1382, opened 2023-11-02
Split the `AccessTokenTrait::convertToJWT()` function so that the token can be customized

rhertogh:
Split the `\League\OAuth2\Server\Entities\Traits\AccessTokenTrait::convertToJWT()` function into two separate functions so that implementing classes can customize the token. E.g.

```php
class AccessToken implements AccessTokenEntityInterface
{
// ...
    use AccessTokenTrait {
        getJwtBuilder as traitGetJwtBuilder;
    }
// ...
    private function getJwtBuilder()
    {
        $builder = $this->traitGetJwtBuilder();

        $builder->withClaim('my_custom_claim', 'custom claim value');

        return $builder;
    }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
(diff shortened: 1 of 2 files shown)


parallels999:
#1328

rhertogh:
Closed in favor of #1328

parallels999:
Also, leave your comment on that PR, so the authors review it at least

PR #1384, opened 2023-11-07
Fixing permission error

pradeep5485:


parallels999:
`.DS_Store` ??
#1383

eugene-borovov:
This change is wrong. Private key may have readonly permission (400 or 440).

Sephster:
Aye it doesn't seem right. Will wait a few days for the author to give an explanation in case we've missed something but I think this isn't fixing anything

Sephster:
Closing as no response from author

PR #1407, opened 2024-05-06
Parse request parameters

eugene-borovov:
@Sephster I have implemented my proposal.

- Params parsed to non-empty string
- Scalar params cast to string
- Empty string treats as null
- Non-scalar params throws invalidRequest
- Remove string checks
- Null checks have been shorten to null-coalesce operator

I will be glad if it helps.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -3,7 +3,3 @@ parameters:
     paths:
         - src
         - tests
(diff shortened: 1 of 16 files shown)


Sephster:
@eugene-borovov thank you very much for this PR. Saved me a great deal of time and looks good. Thank you

PR #1420, opened 2024-06-30
Validate confidential clients and determine if the client handles the grant type

### LINKED ISSUE #1073: Support Public Clients for Password and Refresh Token Grant
Something that was confusing to me about this is the `AuthCodeGrant` [only calls `validateCredentials` for private clients](https://github.com/thephpleague/oauth2-server/blob/e1dc4d708c56fcfa205be4bb1862b6d525b4baac/src/Grant/AuthCodeGrant.php#L102) which made it seem like you no longer needed to check if the client was confidential before attempting to verify the secret.

However the refresh token grant and password grant both are supposed to support public clients according to the OAuth specification and do not check `Client::isConfidential` before calling `validateClient`.

Wouldn't that mean the `isConfidential` check in the `AuthCodeGrant` is unnecessary as you will need to check yourself in `validateClient` in order to support the other grant types?

_Originally posted by @matt-allan in https://github.com/thephpleague/oauth2-server/issues/1034#issuecomment-513974013_

### LINKED ISSUE #1174: Add Unauthorized_Client support
At the moment, the only error we do not handle is unauthorized_client. This is issued when a client tries to use a grant they have not been given permissions to. We should investigate the potential of using this in version 9.

### LINKED ISSUE #1369: RefreshTokenGrant requires client_secret also for non-confidential clients
RefreshTokenGrant::respondToAccessTokenRequest requires a client_secret, otherwise it will throw an exception (OAuthServerException::invalidClient). This does not make sense for non-confidential clients, as they are not able to send the client_secret.

hafezdivandari:
Fixes #1174
Fixes #1369
Fixes #1073
Closes #1036

This PR can be considered as a security enhancement and does 2 main changes:
1. Validate confidential clients:
   * [RFC6749 section 3.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1) 
   * The "auth code" grant calls `AbstractGrant::validateClient()` only if the client is confidential ([RFC6749 section 4.1.3](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3:~:text=2Fcb%0A%0A%20%20%20The%20authorization%20server%20MUST%3A%0A%0A%20%20%20o-,require%20client%20authentication%20for%20confidential%20clients%20or%20for%20any%0A%20%20%20%20%20%20client%20that%20was%20issued%20client%20credentials%20(or%20with%20other%0A%20%20%20%20%20%20authentication%20requirements)%2C,-o%20%20authenticate%20the%20client%20if%20client)). But "device authorization", "refresh token" and "password" grants always validate the client secret. This PR allows public "device authorization" ([RFC8628 section 3.4](https://datatracker.ietf.org/doc/html/rfc8628#section-3.4:~:text=If%20the%20client%20was%20issued%20client%20credentials)), "refresh token" ([RFC6749 section 6](https://datatracker.ietf.org/doc/html/rfc6749#section-6:~:text=it%20was%20issued.-,If%20the%20client%20type%20is%20confidential,-or%0A%20%20%20the%20client)) and "password" ([RFC6749 section 4.3.2](https://datatracker.ietf.org/doc/html/rfc6749#section-4.3.2:~:text=6749%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20OAuth%202.0%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20October%202012%0A%0A%0A%20%20%20The%20authorization%20server%20MUST%3A%0A%0A%20%20%20o-,require%20client%20authentication%20for%20confidential%20clients%20or%20for%20any%0A%20%20%20%20%20%20client%20that%20was%20issued%20client%20credentials%20(or%20with%20other%0A%20%20%20%20%20%20authentication%20requirements)%2C,-o%20%20authenticate%20the%20client%20if%20client%20authentication%20is%20included%2C%20and)) clients by calling the `ClientRepositoryInterface::validateClient()` method only if the client is confidential.
   * Fixes #1073 
   * Fixes #1369
   * The "auth code" and "client credential" grants unnecessarily call `AbstractGrant::getClientEntityOrFail()` twice instead of just using the `AbstractGrant::validateClient()` return value, this PR fixes that.
   * All grants [except "client credentials" grant](https://github.com/thephpleague/oauth2-server/blob/1dc1ee861e8490a9e73d668b1ceccfa27436d861/src/Grant/ClientCredentialsGrant.php#L41-L45) ([RFC6749 section 4.4](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4:~:text=The%20client%20credentials%20grant%20type%20MUST%20only%20be%20used%20by%20confidential%0A%20%20%20clients.)) can be non-confidential. It's already implemented this way on most community integrations:
     * [Passport](https://github.com/laravel/passport/blob/195058c12e035e332ae43f3d0b085f7c6e36795d/src/Bridge/ClientRepository.php#L56)
     * [Symfony](https://github.com/thephpleague/oauth2-server-bundle/blob/44272ff229da25305d4f5612546851bbcaa72d9a/src/Repository/ClientRepository.php#L52-L54)
     * [Mezzio](https://github.com/mezzio/mezzio-authentication-oauth2/blob/1b97f2a0e1bb07e0f6b2b4eda8f7d7806a73f1cf/src/Repository/Pdo/ClientRepository.php#L49-L51)
     * [Drupal](https://git.drupalcode.org/project/simple_oauth/-/blob/d8e843d89b90ec3a05f670f5d52f7369f662fe81/src/Repositories/ClientRepository.php#L60-77)
1. New `ClientTarit::supportsGrantType()` function:
    * [RFC6749 section 5.2](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2:~:text=unauthorized_client%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20The%20authenticated%20client%20is%20not%20authorized%20to%20use%20this%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20authorization%20grant%20type.)
    * Fixes #1174  
    * This function is implemented on `ClientTrait` that returns `true` by default to avoid BC breaking changes.
    * Currently there is no way to check if the client handles the grant type before proceeding the request, e.g. We don't want to make auth code on "auth code grant" or make device code on "device code auth" grant or response with the access token on "implicit token" grant if the specified client doesn't handle the grant type. This PR makes this possible to avoid handling the requested grant type if the specified client doesn't supports that.
    * It also makes it possible for us to disable issuing refresh token if the client doesn't handle this grant.


eugene-borovov (approved):

src/Grant/AuthCodeGrant.php
```diff
@@ -97,14 +97,7 @@ public function respondToAccessTokenRequest(
         ResponseTypeInterface $responseType,
         DateInterval $accessTokenTTL
     ): ResponseTypeInterface {
-        list($clientId) = $this->getClientCredentials($request);
-
-        $client = $this->getClientEntityOrFail($clientId, $request);
```
Sephster, line 102: Nice spot here

src/Grant/AbstractGrant.php
```diff
@@ -151,18 +151,18 @@ protected function validateClient(ServerRequestInterface $request): ClientEntity
...
 
-        // If a redirect URI is provided ensure it matches what is pre-registered
-        $redirectUri = $this->getRequestParameter('redirect_uri', $request);
+        if ($client->isConfidential()) {
+            if ($clientSecret === '') {
+                throw OAuthServerException::invalidRequest('client_secret');
+            }
 
-        if ($redirectUri !== null) {
-            $this->validateRedirectUri($redirectUri, $client, $request);
```
Sephster, line 165: Where are we checking this now? I can't see where this is being checked and note the tests for this have been deleted. I'm guessing it is done somewhere else but can't spot it sorry. Could you point me in the right direction. We need to ensure the redirect URI is as expected

Sephster (commented):
This looks good @hafezdivandari - I like the way you've supported the unauthorized_grant without introducing breaking changes. A nice solution.

Only concern I have is around the validation of the redirect URI. I've left a comment. Thanks for your hard work on this

src/Grant/AbstractGrant.php
hafezdivandari: The redirect URI validation is done on the grant itself:

* Auth Code:
https://github.com/thephpleague/oauth2-server/blob/d4ef04740fd16da8582b0551987dcaa9ee777b38/src/Grant/AuthCodeGrant.php#L276

* Implicit:
https://github.com/thephpleague/oauth2-server/blob/d4ef04740fd16da8582b0551987dcaa9ee777b38/src/Grant/ImplicitGrant.php#L112

hafezdivandari: The removed tests were on `AbstractGrantTest`, as other grants don’t involve a redirect URI, It doesn’t make sense to validate the incoming redirect URI for all grants, especially when it’s already being validated in the grants that actually use a redirect URI (authorization code and implicit).

Sephster: Awesome. Thanks Hafez, makes total sense.

Issue #1422, opened 2024-07-05
Autogenerated Emitter is not persisted


ls-sean-fraser:
I believe I have found another bug in the library (running the latest version). The `getEmitter()` method does not persist the autogenerated object. As a result, every call to that method returns a separate emitter instance, meaning that events emitted from one instance will never reach the listeners which are registered to a different instance

https://github.com/thephpleague/oauth2-server/blob/master/src/EventEmitting/EmitterAwarePolyfill.php#L16

```php
public function getEmitter(): EventEmitter
    {
        return $this->emitter ?? new EventEmitter();
    }
```

Was that intended to be a `??=` operator?

**Workaround**: explicitly call `$server->setEmitter(new EventEmitter())` BEFORE assigning any grants to the server. This kind of defeats the purpose of the autogeneration.

**Question**: Should `$server->setEmitter()` automatically cascade the reassignment to any registered grant types?

Issue #1418, opened 2024-06-26
Issue with redirect_uri in Authorization Code Grant


ls-sean-fraser:
Using version 9.0.0, using authorization code grant without specifying a redirect_uri in both requests does not seem to be accepted. The spec indicates these are only required if used in both places.

```
redirect_uri
         REQUIRED, if the "redirect_uri" parameter was included in the
         authorization request as described in Section 4.1.1, and their
         values MUST be identical.
```

When using a default Authorization code entity...

```php
class OauthAuthorizationCode implements AuthCodeEntityInterface
{
    use EntityTrait;
    use TokenEntityTrait;
    use AuthCodeTrait;
}
```

If the redirect_uri is omitted from the authorize request, the authorization code contains a redirect_uri of `null`. 

The access token call then fail here, As the value is `null`, but it is checking for empty string  

https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php#L220-L224

```php  
if ($authCodePayload->redirect_uri !== '' && $redirectUri === null) {
     throw OAuthServerException::invalidRequest('redirect_uri');
}
```

Forcing the null redirect_uri to be empty string in the entity doesn't resolve the issue as the subsequent check which will fail to compare redirect_uri of the code and the request, as `'' !== null`.

I suspect that the check above should be changed to be: 

```php
if ($authCodePayload->redirect_uri !== null && $redirectUri === null) { 
```

Sephster:
Have you got a redirect Uri registered against your client?

ls-sean-fraser:
Yes, my `ClientEntityInterface` implementation returns a value for the client. The [redirect itself](https://github.com/thephpleague/oauth2-server/blob/master/src/Grant/AuthCodeGrant.php#L355) works fine, it is only validation of it during the code exchange that is the problem

Sephster:
Thanks for reporting this. I think I have a solution figured out. I'm going to change the code so we validate that if a redirect URI is passed, it must be a valid URI as per the specs. That negates the need to check for non empty strings as we shouldn't be accepting these in the first place.

I'll finalise this tomorrow and submit a fix. Thank you

ls-sean-fraser:
Hi @Sephster, wondering if you might have an update on this task. Hoping to see this bug addressed before we rollout this library, so we don't need to force our partners to specify the redirect_uri every time

Sephster:
[...]

PR #1428, opened 2024-07-22
Fix explicit redirect uri not being optional in authorization code grant

ls-sean-fraser:
- Fix explicit redirect uri not being optional in authorization code grant (addresses #1418)
- Also fixed event emitter trait not storing automatically created instance (addresses #1422)

Unit tests added for all changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Fixed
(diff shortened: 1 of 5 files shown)


Sephster:
Thanks for this @ls-sean-fraser. Looks great. I've double checked everything and all stands up. Only change I made was to one of the test names to make it more verbose as it took me a while to work out what it was testing. Also added in a changelog entry. Thanks for your PR and issue posting!

ls-sean-fraser:
Thank you @Sephster! Happy to help! 

Can you tag a patch version so that we can pull the fix?

ls-sean-fraser:
Hi @Sephster, still waiting for a new patch version to be tagged with these fixes

Sephster:
Apologies. Will get to this asap. Been snowed under at work

ls-sean-fraser:
I know the feeling 😅

Issue #1435, opened 2024-08-07
v9 Typing breaks existing refresh tokens


MartinGreen:
I think the typing stuff in v9 has broken compatibility with my existing refresh tokens.

Ive just updated to v9 and I was able to add all the types to my classes and everything seems to be working fine. I can create new tokens via auth code or device code and via refresh.
However when I try to use an existing refresh token to get a new access token it fails with the following error:

```
Message: League\OAuth2\Server\Grant\AbstractGrant::issueAccessToken(): Argument #3 ($userIdentifier) must be of type ?string, int given, called in /workspace/vendor/league/oauth2-server/src/Grant/RefreshTokenGrant.php on line 81
File: /workspace/vendor/league/oauth2-server/src/Grant/AbstractGrant.php
Line: 404
Trace: /workspace/vendor/league/oauth2-server/src/Grant/RefreshTokenGrant.php(81): League\OAuth2\Server\Grant\AbstractGrant->issueAccessToken(Object(DateInterval), Object(ClientEntity), 3, Array)
       /workspace/vendor/league/oauth2-server/src/AuthorizationServer.php(173): League\OAuth2\Server\Grant\RefreshTokenGrant->respondToAccessTokenRequest(Object(Slim\Http\ServerRequest), Object(League\OAuth2\Server\ResponseTypes\BearerTokenResponse), Object(DateInterval))
       /workspace/index.php(304): League\OAuth2\Server\AuthorizationServer->respondToAccessTokenRequest(Object(Slim\Http\ServerRequest), Object(Slim\Http\Response))
```

I believe this is because my userid used to be an integer. Ive changed anywhere that I pass the user id into the system to convert it to a string so its working for new stuff but existing tokens contain the id as an int.
Since the refresh token is handled inside the library I dont think theres anywhere I can change my code to convert it before creating the new token.

It looks like when respondToAccessTokenRequest in RefreshTokenGrant reads the old refresh token info it would need to convert it to string before calling issueAccessToken.

Thanks.

btw. The  Device Code support is great I've been waiting for that! Cheers!

Maksimius:
Same problem.
```
 TypeError: League\OAuth2\Server\Grant\AbstractGrant::issueAccessToken(): Argument #3 ($userIdentifier) must be of type ?string, int given, call
[...]

PR #1436, opened 2024-08-08
Fix refresh token int user ID

ssigwart:
Fixes https://github.com/thephpleague/oauth2-server/issues/1435

Old refresh tokens may have int user IDs causing an exception. This fixes that.

### DIFF
--- a/src/Grant/RefreshTokenGrant.php
+++ b/src/Grant/RefreshTokenGrant.php
@@ -79,8 +79,9 @@ class RefreshTokenGrant extends AbstractGrant
 
         // Issue and persist new access token
         $userId = $oldRefreshToken['user_id'];
-        if (is_int($userId))
+        if (is_int($userId)) {
             $userId = (string)$userId;
+        }
         $accessToken = $this->issueAccessToken($accessTokenTTL, $client, $userId, $scopes);
         $this->getEmitter()->emit(new RequestAccessTokenEvent(RequestEvent::ACCESS_TOKEN_ISSUED, $request, $accessToken));
         $responseType->setAccessToken($accessToken);
--- a/src/Grant/RefreshTokenGrant.php
+++ b/src/Grant/RefreshTokenGrant.php
@@ -78,7 +78,10 @@ class RefreshTokenGrant extends AbstractGrant
         }
 
         // Issue and persist new access token
-        $accessToken = $this->issueAccessToken($accessTokenTTL, $client, (string)$oldRefreshToken['user_id'], $scopes);
+        $userId = $oldRefreshToken['user_id'];
+        if (is_int($userId))
+            $userId = (string)$userId;
+        $accessToken = $this->issueAccessToken($accessTokenTTL, $client, $userId, $scopes);
         $this->getEmitter()->emit(new RequestAccessTokenEvent(RequestEvent::ACCESS_TOKEN_ISSUED, $request, $accessToken));
         $responseType->setAccessToken($accessToken);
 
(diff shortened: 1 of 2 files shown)


src/Grant/RefreshTokenGrant.php
```diff
@@ -78,7 +78,11 @@ public function respondToAccessTokenRequest(
         }
 
         // Issue and persist new access token
-        $accessToken = $this->issueAccessToken($accessTokenTTL, $client, $oldRefreshToken['user_id'], $scopes);
```
Sephster, line 81: could we just cast it regardless to simplify the code? e.g. `$this->issueAccessToken($accessTokenTTL, $client, (string) $oldRefreshToken['user_id'], $scopes);`

Sephster (commented):
Just one minor comment. Thank you for your PR. Looks like this will do the job. Apologies I didn't notice this when releasing v9

src/Grant/RefreshTokenGrant.php
ssigwart: If I remember correctly, I tried that initially and ran into issues if `$oldRefreshToken['user_id']` is null.

Ah, yes, here's my initial commit and I think the GitHub checks failed:
https://github.com/thephpleague/oauth2-server/pull/1436/commits/d5419d58cbbc41c8e6691ed6e1b51713151c174c

ssigwart (commented):
> Just one minor comment. Thank you for your PR. Looks like this will do the job. Apologies I didn't notice this when releasing v9

No problem. Thanks for working on this project.

src/Grant/RefreshTokenGrant.php
Maksimius: `(string) $oldRefreshToken['user_id'] ` convert null to ''
if ($oldRefreshToken['user_id'] !== null) {
$oldRefreshToken['user_id'] = (string) $oldRefreshToken['user_id'];
}
or some like
$oldRefreshToken['user_id'] = is_null($oldRefreshToken['user_id']) ? null : (string) $oldRefreshToken['user_id'];

ssigwart: I'd like to hear @Sephster's thoughts first. I'm not against checking for null, but I thought that `is_int` was safer because it will only convert things that are known integers. If it is somehow an array or float, I didn't want to convert them. They will result in a `TypeError`, but I thought that's probably a good thing because I don't think those would be expected.

eugene-borovov: This is bulletproof :-)

```php

$oldRefreshToken['user_id'] = is_scalar($oldRefreshToken['user_id'] ?? null) ? null : (\trim((string) $oldRefreshToken['user_id']) ?: null);

```

Maksimius: But if it is_scalar, it will converted to null. Wouldn't it be better to convert it to a string?

eugene-borovov: Yes, it is better. This is my mistake. All non-scalar and empty values will be converted into null.

```php
$oldRefreshToken['user_id'] = is_scalar($oldRefreshToken['user_id'] ?? null) ? (trim((string) $oldRefreshToken['user_id']) ?: null) : null;
```

Sephster: Sorry I hadn't thought of the null issue. Makes sense. I think the solution offered in the PR fixes the issue at hand so happy to merge as is. Thank you and thanks to everyone for their input on this

Sephster (approved):

PR #1447, opened 2024-09-26
Fix key files permission on Windows

hafezdivandari:
Closes #901 
Related to laravel/passport#1789

This PR adds windows tests and fixes the permission check on Windows. The possible file permissions on Windows are '666' and '444'.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,12 +12,12 @@ jobs:
       fail-fast: false
       matrix:
         php: [8.1, 8.2, 8.3]
-        os: [ubuntu-22.04]
+        os: [ubuntu-latest, windows-latest]
         stability: [prefer-lowest, prefer-stable]
 
     runs-on: ${{ matrix.os }}
 
-    name: PHP ${{ matrix.php }} - ${{ matrix.stability }}
+    name: PHP ${{ matrix.php }} - ${{ matrix.stability }} - ${{ matrix.os }}
 
     steps:
       - name: Checkout code
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          extensions: dom, curl, libxml, mbstring, zip
+          extensions: dom, curl, libxml, mbstring, sodium, zip
           coverage: pcov
 
       - name: Install dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          extensions: dom, curl, libxml, mbstring, zip
+          extensions: dom, curl, libxml, mbstring, sodium, zip
```
hafezdivandari, line 32: `sodium` is not enabled on Windows by default, it is required by `lcobucci/jwt` package.

tests/AuthorizationServerTest.php
```diff
@@ -50,6 +50,15 @@ public function setUp(): void
...
 
+    public function testKeyPermissions(): void
+    {
+        $permission = PHP_OS_FAMILY === 'Windows' ? '666' : '600';
+
+        self::assertSame($permission, decoct(fileperms(__DIR__ . '/Stubs/private.key') & 0777));
+        self::assertSame($permission, decoct(fileperms(__DIR__ . '/Stubs/public.key') & 0777));
+        self::assertSame($permission, decoct(fileperms(__DIR__ . '/Stubs/private.key.crlf') & 0777));
+    }
+
```
hafezdivandari, line 61: You may delete this test, I added this to prove the difference on Windows only!

src/CryptKey.php
```diff
@@ -72,7 +72,7 @@ public function __construct(string $keyPath, protected ?string $passPhrase = nul
             throw new LogicException('Invalid key supplied');
         }
 
-        if ($keyPermissionsCheck === true) {
+        if ($keyPermissionsCheck === true && PHP_OS_FAMILY !== 'Windows') {
```
hafezdivandari, line 75: Ignore key permissions check on Windows, as it always return '666' and '444'.

hafezdivandari:
> This change does mean that regardless of this flag, on windows we will ignore it. This is potentially not obvious to the end user and while nothing bad will come of it, it doesn't feel right.

The ignored check is not working on Windows anyway. The tests are always failing on Windows without this change.

> Why can't this option be exposed in Laravel so folks can turn it off if need be?

On Passport, we had to [turn this off on Windows](https://github.com/laravel/passport/blob/e94b5311cee98f5972c14576314922e327c9f8b3/src/PassportServiceProvider.php#L325), but it doesn't feel right either! What if additional checks added in future, or the code changed in a way that works on Windows?

Feel free to close this PR if it was not convincing. We will continue to ignore this for Windows on Passport for now, so it's not a blocking pull request.

Sephster:
Thanks @hafezdivandari  - I hadn't really considered testing against Windows so this does make sense. Will merge on that basis. Cheers!

PR #1460, opened 2024-11-22
 use Laminas\Diactoros\Stream 

xerc:
https://github.com/thephpleague/oauth2-server/blob/d511107cb018ead0bd84f86402b086306738c686/examples/composer.lock#L388-L389

ref. https://github.com/thephpleague/oauth2-server/pull/1074/commits/5fee7e447e7b057e0fcbb91ff62281b62c345db8#diff-70ce929b7da0d7c0659c815690532adb32d57e442bbc0ffc5b3a8b667d95b933R23

### DIFF
--- a/examples/public/auth_code.php
+++ b/examples/public/auth_code.php
@@ -10,6 +10,8 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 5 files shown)


Sephster:
thank you!

PR #1461, opened 2024-11-26
cleanup double+EOL spaces

xerc:


src/ResponseTypes/DeviceCodeResponse.php
```diff
@@ -35,10 +35,10 @@ public function generateHttpResponse(ResponseInterface $response): ResponseInter
         $expireDateTime = $this->deviceCodeEntity->getExpiryDateTime()->getTimestamp();
 
         $responseParams = [
-            'device_code' => $this->deviceCodeEntity->getIdentifier(),
-            'user_code' => $this->deviceCodeEntity->getUserCode(),
+            'device_code'      => $this->deviceCodeEntity->getIdentifier(),
```
Sephster, line 38: I don't think we bother to align array key/values in this way anywhere else in the codebase. Happy with all other changes but is it possible to revert this? If we were to enforce something like this I'd prefer it was done in our coding style automations. Thank you

Sephster (changes requested):
Just the one comment from me. Thanks @xerc

src/ResponseTypes/DeviceCodeResponse.php
Sephster: thanks for examples. We should maybe add a cs rule to be consistent. Thanks. WIll merge now

PR #1462, opened 2024-11-26
sort require & opt. example

xerc:
fyi/ `localhost` results in IPv6 @ newer MacOS versions


README.md
```diff
@@ -32,10 +32,7 @@ This library was created by Alex Bilbie. Find him on Twitter at [@alexbilbie](ht
 
 The latest version of this package supports the following versions of PHP:
 
-* PHP 8.1
-* PHP 8.2
-* PHP 8.3
-* PHP 8.4
+https://github.com/thephpleague/oauth2-server/blob/7b9ec37d5077c0382aa3748c6e125c5408d5adb9/composer.json#L7
```
Sephster, line 35: I think the previous format is preferable as the readme should be easily accessible. I think this change reduces that

composer.json
```diff
@@ -4,28 +4,27 @@
     "homepage": "https://oauth2.thephpleague.com/",
     "license": "MIT",
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
+        "php": "^8.1 || ^8.2 || ^8.3 || ^8.4",
```
Sephster, line 7: please can we revert this change? I can't remember the reasons why but we did have the constraint as ^8.0 at one point but it was reverted to ~. I think we will change to just ^9.0 when a new major PHP ver is released.

Sephster (changes requested):
Thanks for your PR. Just a few minor comments from me then should be good to merge. Thank you

Sephster:
Closing as not essential and no further updates from author

xerc:
@Sephster didnt you see my comments @ https://github.com/thephpleague/oauth2-server/pull/1462/files/a9b9429d26b41b5d594b615a7bfcbf68584cd688#r1868528200 ?

PR #1470, opened 2025-01-19
Remove the encrypted payload from AuthCode and RefreshToken

craigp1231:
The Auth Code and Refresh Token information can be stored in the repository and doesn't have to go in the payload. This will reduce auth code and refresh token size, but will require more calls to the repository.

If the user specifies an encryption key, the payload will be attempted to be encrypted or decrypted, else they are returned from the respective repositories.

### DIFF
--- a/src/AuthorizationServer.php
+++ b/src/AuthorizationServer.php
@@ -48,7 +48,7 @@ class AuthorizationServer implements EmitterAwareInterface
 
     protected ResponseTypeInterface $responseType;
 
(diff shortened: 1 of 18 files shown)


Sephster:
Hi @craigp1231 - is it possible to bring this in line with master so I can review? I was also wondering if you could outline why you want this change. Historically, we have deliberately set the library up this was to reduce repository access overhead. I'd be keen to understand why the token size is an issue. Thank you

PR #1135, opened 2020-08-27
Add token revocation support (RFC 7009)

tomsisk:
This fixes the merge conflicts in #995 by @jacobweber

https://tools.ietf.org/html/rfc7009

PR #1473, opened 2025-02-14
Support token revocation and introspection

### LINKED ISSUE #806: Token Revocation question

### LINKED ISSUE #1350: Access token introspection RFC7662
Is that functional planned to be implemented for future or done already. What different ways are there to separate Authorization server from Resource server. Thanks in advance and sorry if that is wrong place to ask questions like that.

hafezdivandari:
Fixes #806 
Fixes #1350
Closes #1255 
Closes #1135 
Closes #995 
Closes #925 

It's much easier to implement [Token Revocation RFC7009](https://datatracker.ietf.org/doc/html/rfc7009) and [Token Introspection RFC7662](https://datatracker.ietf.org/doc/html/rfc7662) at the same time! These 2 RFCs have most of the logic in common:
* The same HTTP POST request with the same parameters sent as
 "application/x-www-form-urlencoded" data:
  * `token` is required, accepts either an access token or a refresh token.
  * `token_type_hint` is optional to help the server optimize the token lookup. `access_token` or `refresh_token`.
* The same client authentication:
  *  Via `Authorization: Basic {client_credentials}` header
  * Or via `client_id` and `client_secret` parameters.
* The same access/refresh token validation:
  * The token was issued to the client making the request,
  * and is valid/active (has valid signature/encryption, is not expired, and is not revoked).
* The same HTTP status responses:
  * `400`: `invalid_request` the required `token` parameter was not sent.
  * `401`: `invalid_client` the request is not authorized. Client credentials were not sent or invalid.
  * `200`: The token is revoked, expired, doesn't exists, or was not issued to the client making the request:
    * "Token revocation": The response body is empty.
    * "Token Introspection": The JSON response has the `active` field set to `false`. `{ "active": false }`
  * `200`: The token is valid, active, and was issued to the client making the request:
    * "Token revocation": The response body is empty.
    * "Token Introspection": The JSON response has the `active` field set to `true`. `{ "active": true, ... }`

## Implementation
* Fully backward-compatible (No BC breaking changes at all!).
* Fully consistent with the current code style and project structure.
* Uses the existing code within the package without adding any redundancy.
* Fully customizable. Defines interfaces and allow injecting customized instances.
* Fully tested.

## Summary

Class Hierarchy

Summary of file changes in order of appearance.

<details><summary>
New <code>AbstractHandler</code> class
</summary>
<p>

* Extended by `AbstractGrant` and `AbstractTokenHandler` classes.
* `clientRepository`, `accessTokenRepository`, and `refreshTokenRepository` properties have been moved to this class from child `AbstractGrant` class.
* `setClientRepository`, `setAccessTokenRepository`, `setRefreshTokenRepository`, `getClientCredentials`, `parseParam`, `getRequestParameter`, `getBasicAuthCredentials`, `getQueryStringParameter`, `getCookieParameter`, `getServerParameter`, and `validateEncryptedRefreshToken` methods have been moved to this class from child `AbstractGrant` class without any change.
* `validateClient` method from child `AbstractGrant` has been moved to this class with a minor change: The call to `getIdentifier()` is `GrantTypeInterface` specific.
* `getClientEntityOrFail` method from child `AbstractGrant` has been moved to this class. This method is also overridden on child `AbstractGrant`: Call to `supportsGrantType` method is `AbstracGrant` specific.
* New `validateEncryptedRefreshToken` is extracted from `RefreshTokenGrant::validateOldRefreshToken` method.

</p>
</details>

<details><summary>
Change <code>AuthorizationValidators\BearerTokenValidator</code> class
</summary>
<p>

* Now implements new `BearerTokenValidatorInterface` interface.
* Extract the bearer token validation logic from `validateAuthorization` method into a new `validateBearerToken` method.
* The new `validateBearerToken` method also accepts optional `$clientId` argument. If a `$clientId` is provided, it verifies whether the token was issued to the given client.

</p>
</details>

<details><summary>
New <code>AuthorizationValidators\BearerTokenValidatorInterface</code> interface
</summary>
<p>

* Implemented by `BearerTokenValidator` class.
* Defines `validateBearerToken` method.
* Used by `AbstractTokenHandler::setBearerTokenValidator()` method.

</p>
</details>

<details><summary>
Change <code>Exception\OAuthServerException</code> class
</summary>
<p>

* Define new `unsupportedTokenType` method.

</p>
</details>

<details><summary>
Change <code>Grant\AbstractGrant</code> class
</summary>
<p>

* Now extends new `AbstractHandler` class.
* All removed liens are moved to the parent `AbstractHandler` class.
* Overrides `getClientEntityOrFail` method.

</p>
</details>

<details><summary>
Change <code>Grant\RefreshTokenGrant</code> class
</summary>
<p>

* The encrypted refresh token validation logic from `validateOldRefreshToken` method has been extracted into a new parent AbstractHandler::validateEncryptedRefreshToken method.

</p>
</details>

<details><summary>
New <code>Handlers\AbstractTokenHandler</code> class
</summary>
<p>

* Extends new `AbstractHandler` class.
* Implements new `TokenHandlerInterface` interface.
* Defines `setPublicKey` and `setBearerTokenValidator` methods that to allow injecting a customized token validator.
* Defines `validateToken`, `validateAccessToken` and `validateRefreshToken` methods that have been used by `TokenIntrospectionHandler` and `TokenRevocationHandler` classes.

</p>
</details>

<details><summary>
New <code>Handlers\TokenHandlerInterface</code> interface
</summary>
<p>

* Implemented by new `AbstractTokenHandler` class.
* Used by `setTokenRevocationHandler` and `setTokenIntrospectionHandler` methods of the `TokenServer` class.

</p>
</details>

<details><summary>
New <code>Handlers\TokenIntrospectionHandler</code> class
</summary>
<p>

* Extends new `AbstractTokenHandler` class.
* Defines `respondToRequest` method that responds to "Token Introspection RFC7662" request.
* Defines `setResponseType` method that allows injecting a customized response type.

</p>
</details>

<details><summary>
New <code>Handlers\TokenRevocationHandler</code> class
</summary>
<p>

* Extends new `AbstractTokenHandler` class.
* Defines `respondToRequest` method that responds to "Token Revocation RFC7009" request.

</p>
</details>

<details><summary>
New <code>ResponseTypes\IntrospectionResponse</code> class
</summary>
<p>

* Implements new `IntrospectionResponseTypeInterface` interface.
* Defines `generateHttpResponse` method to generate "Token Revocation RFC7009" response.

</p>
</details>

<details><summary>
New <code>ResponseTypes\IntrospectionResponseTypeInterface</code> class
</summary>
<p>

* Implemented by new `IntrospectionResponse` class.
* Used by `TokenIntrospectionHandler::setResponseType` method that allows injecting a customized response type.

</p>
</details>

<details><summary>
New <code>TokenServer</code> class
</summary>
<p>

* Defines `respondToTokenRevocationRequest` and `respondToTokenIntrospectionRequest` methods to respond to the respective request.
* Defines `setTokenRevocationHandler` and `setTokenIntrospectionHandler` methods that allows injecting customized handlers.

</p>
</details>


ezequidias:
@Sephster , as mentioned https://github.com/thephpleague/oauth2-server/issues/1453#issuecomment-2622203040, it would be great to have more maintainers for this repository! @hafezdivandari is an excellent professional with extensive knowledge of OAuth2. Having more people to support the project would certainly help its growth and maintenance! 🚀

src/AuthorizationValidators/BearerTokenValidator.php
```diff
@@ -120,16 +135,20 @@ public function validateAuthorization(ServerRequestInterface $request): ServerRe
...
 
+        // Check if token is linked to the client
+        if (
+            $clientId !== null &&
+            $claims->get('client_id') !== $clientId &&
+            !$token->isPermittedFor($clientId)
+        ) {
+            throw OAuthServerException::accessDenied('Access token is not linked to client');
+        }
+
```
hafezdivandari, line 146: The new `validateJwt` method optionally accepts `$clientId` argument. If a `$clientId` is provided, it verifies whether the token was issued to the given client.

src/AuthorizationValidators/JwtValidatorInterface.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    /**
+     * Parse and validate the given JWT.
+     *
+     * @param non-empty-string      $jwt
+     * @param non-empty-string|null $clientId
+     *
+     * @return array<non-empty-string, mixed>
+     */
+    public function validateJwt(ServerRequestInterface $request, string $jwt, ?string $clientId = null): array;
```
hafezdivandari, line 19: Define new `JwtValidatorInterface` interface that defines `validateJwt` method. This interface is implemented by `BearerTokenValidator` class.

src/Grant/AbstractGrant.php
```diff
@@ -46,36 +42,25 @@
...
 use function is_string;
 use function random_bytes;
-use function substr;
 use function trim;
 
 /**
  * Abstract grant class.
  */
-abstract class AbstractGrant implements GrantTypeInterface
+abstract class AbstractGrant extends AbstractHandler implements GrantTypeInterface
```
hafezdivandari, line 54: Makes `AbstractGrant` class extend the new `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -46,36 +42,25 @@
...
 {
-    use EmitterAwarePolyfill;
-    use CryptTrait;
-
```
hafezdivandari, line 64: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -46,36 +42,25 @@
...
     protected const SCOPE_DELIMITER_STRING = ' ';
 
     protected const MAX_RANDOM_TOKEN_GENERATION_ATTEMPTS = 10;
 
-    protected ClientRepositoryInterface $clientRepository;
-
-    protected AccessTokenRepositoryInterface $accessTokenRepository;
-
```
hafezdivandari, line 72: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -46,36 +42,25 @@
...
     protected ScopeRepositoryInterface $scopeRepository;
 
     protected AuthCodeRepositoryInterface $authCodeRepository;
 
-    protected RefreshTokenRepositoryInterface $refreshTokenRepository;
-
```
hafezdivandari, line 78: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -86,26 +71,11 @@ abstract class AbstractGrant implements GrantTypeInterface
...
-    public function setClientRepository(ClientRepositoryInterface $clientRepository): void
-    {
-        $this->clientRepository = $clientRepository;
-    }
-
-    public function setAccessTokenRepository(AccessTokenRepositoryInterface $accessTokenRepository): void
-    {
-        $this->accessTokenRepository = $accessTokenRepository;
-    }
-
```
hafezdivandari, line 98: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -86,26 +71,11 @@ abstract class AbstractGrant implements GrantTypeInterface
...
     public function setScopeRepository(ScopeRepositoryInterface $scopeRepository): void
     {
         $this->scopeRepository = $scopeRepository;
     }
 
-    public function setRefreshTokenRepository(RefreshTokenRepositoryInterface $refreshTokenRepository): void
-    {
-        $this->refreshTokenRepository = $refreshTokenRepository;
-    }
-
```
hafezdivandari, line 108: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -143,51 +113,11 @@ public function revokeRefreshTokens(bool $willRevoke): void
...
-            if ($this->clientRepository->validateClient($clientId, $clientSecret, $this->getIdentifier()) === false) {
-                $this->getEmitter()->emit(new RequestEvent(RequestEvent::CLIENT_AUTHENTICATION_FAILED, $request));
-
-                throw OAuthServerException::invalidClient($request);
-            }
-        }
-
-        return $client;
-    }
```
hafezdivandari, line 169: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -143,51 +113,11 @@ public function revokeRefreshTokens(bool $willRevoke): void
...
-     * Wrapper around ClientRepository::getClientEntity() that ensures we emit
-     * an event and throw an exception if the repo doesn't return a client
-     * entity.
-     *
-     * This is a bit of defensive coding because the interface contract
-     * doesn't actually enforce non-null returns/exception-on-no-client so
-     * getClientEntity might return null. By contrast, this method will
-     * always either return a ClientEntityInterface or throw.
-     *
-     * @throws OAuthServerException
```
hafezdivandari, line 181: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -143,51 +113,11 @@ public function revokeRefreshTokens(bool $willRevoke): void
...
+     * {@inheritdoc}
      */
     protected function getClientEntityOrFail(string $clientId, ServerRequestInterface $request): ClientEntityInterface
     {
-        $client = $this->clientRepository->getClientEntity($clientId);
-
-        if ($client instanceof ClientEntityInterface === false) {
-            $this->getEmitter()->emit(new RequestEvent(RequestEvent::CLIENT_AUTHENTICATION_FAILED, $request));
-            throw OAuthServerException::invalidClient($request);
-        }
```
hafezdivandari, line 190: To parent `AbstractHandler` class. Extracted to `parent::getClientEntityOrFail` method.

src/Grant/AbstractGrant.php
```diff
@@ -205,29 +135,6 @@ protected function supportsGrantType(ClientEntityInterface $client, string $gran
...
-
-        if ($clientId === null) {
-            throw OAuthServerException::invalidRequest('client_id');
-        }
-
-        $clientSecret = $this->getRequestParameter('client_secret', $request, $basicAuthPassword);
-
-        return [$clientId, $clientSecret ?? ''];
-    }
-
```
hafezdivandari, line 230: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -289,123 +196,6 @@ private function convertScopesQueryStringToArray(string $scopes): array
...
-        if ($value === '') {
-            $value = $default === null ? null : trim($default);
-
-            if ($value === '') {
-                $value = null;
-            }
-        }
-
-        return $value;
-    }
```
hafezdivandari, line 320: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -289,123 +196,6 @@ private function convertScopesQueryStringToArray(string $scopes): array
...
-     * Retrieve request parameter.
-     *
-     * @return non-empty-string|null
-     *
-     * @throws OAuthServerException
-     */
-    protected function getRequestParameter(string $parameter, ServerRequestInterface $request, ?string $default = null): ?string
-    {
-        return self::parseParam($parameter, (array) $request->getParsedBody(), $default);
-    }
```
hafezdivandari, line 332: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -289,123 +196,6 @@ private function convertScopesQueryStringToArray(string $scopes): array
...
-        }
-
-        [$username, $password] = explode(':', $decoded, 2);
-
-        if ($username === '') {
-            return [null, null];
-        }
-
-        return [$username, $password];
-    }
```
hafezdivandari, line 371: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -289,123 +196,6 @@ private function convertScopesQueryStringToArray(string $scopes): array
...
-     * Retrieve query string parameter.
-     *
-     * @return non-empty-string|null
-     *
-     * @throws OAuthServerException
-     */
-    protected function getQueryStringParameter(string $parameter, ServerRequestInterface $request, ?string $default = null): ?string
-    {
-        return self::parseParam($parameter, $request->getQueryParams(), $default);
-    }
```
hafezdivandari, line 383: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -289,123 +196,6 @@ private function convertScopesQueryStringToArray(string $scopes): array
...
-     * Retrieve cookie parameter.
-     *
-     * @return non-empty-string|null
-     *
-     * @throws OAuthServerException
-     */
-    protected function getCookieParameter(string $parameter, ServerRequestInterface $request, ?string $default = null): ?string
-    {
-        return self::parseParam($parameter, $request->getCookieParams(), $default);
-    }
```
hafezdivandari, line 395: To parent `AbstractHandler` class.

src/Grant/AbstractGrant.php
```diff
@@ -289,123 +196,6 @@ private function convertScopesQueryStringToArray(string $scopes): array
...
-     *
-     * @return non-empty-string|null
-     *
-     * @throws OAuthServerException
-     */
-    protected function getServerParameter(string $parameter, ServerRequestInterface $request, ?string $default = null): ?string
-    {
-        return self::parseParam($parameter, $request->getServerParams(), $default);
-    }
-
```
hafezdivandari, line 408: To parent `AbstractHandler` class.

src/Grant/RefreshTokenGrant.php
```diff
@@ -107,28 +104,7 @@ protected function validateOldRefreshToken(ServerRequestInterface $request, stri
...
-
-        if ($refreshTokenData['expire_time'] < time()) {
-            throw OAuthServerException::invalidRefreshToken('Token has expired');
-        }
-
-        if ($this->refreshTokenRepository->isRefreshTokenRevoked($refreshTokenData['refresh_token_id']) === true) {
-            throw OAuthServerException::invalidRefreshToken('Token has been revoked');
-        }
-
-        return $refreshTokenData;
```
hafezdivandari, line 131: Extract the encrypted refresh token validation logic from `RefreshTokenGrant::validateOldRefreshToken` method into a new parent `AbstractHandler::validateEncryptedRefreshToken` method.

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,106 @@
...
+
+use League\OAuth2\Server\AbstractHandler;
+use League\OAuth2\Server\AuthorizationValidators\BearerTokenValidator;
+use League\OAuth2\Server\AuthorizationValidators\JwtValidatorInterface;
+use League\OAuth2\Server\CryptKeyInterface;
+use League\OAuth2\Server\Entities\ClientEntityInterface;
+use League\OAuth2\Server\Exception\OAuthServerException;
+use Psr\Http\Message\ServerRequestInterface;
+
+abstract class AbstractTokenHandler extends AbstractHandler implements TokenHandlerInterface
```
hafezdivandari, line 15: New `AbstractTokenHandler` that extends `AbstractHandler` defines required methods to validate access/refresh tokens, injecting `JwtValidatorInterface` and public key. This class is inspired by `AbstractGrant`.

src/Handlers/TokenHandlerInterface.php
```diff
@@ -0,0 +1,29 @@
...
+use Defuse\Crypto\Key;
+use League\OAuth2\Server\CryptKeyInterface;
+use League\OAuth2\Server\EventEmitting\EmitterAwareInterface;
+use League\OAuth2\Server\Repositories\AccessTokenRepositoryInterface;
+use League\OAuth2\Server\Repositories\ClientRepositoryInterface;
+use League\OAuth2\Server\Repositories\RefreshTokenRepositoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
+interface TokenHandlerInterface extends EmitterAwareInterface
```
hafezdivandari, line 16: New `TokenHandlerInterface` that has been implemented by `AbstractTokenHandler`.

This interface is inspired by `GrantTypeInterface` that has been implemented by `AbstractGrant` class.

src/Handlers/TokenIntrospectionHandler.php
```diff
@@ -0,0 +1,43 @@
...
+declare(strict_types=1);
+
+namespace League\OAuth2\Server\Handlers;
+
+use League\OAuth2\Server\ResponseTypes\IntrospectionResponse;
+use League\OAuth2\Server\ResponseTypes\IntrospectionResponseTypeInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
+class TokenIntrospectionHandler extends AbstractTokenHandler
```
hafezdivandari, line 12: This class implements "Token Introspection RFC7662". This class also allows customizing the introspection response by injecting a `IntrospectionResponseTypeInterface` instance.

src/Handlers/TokenRevocationHandler.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OAuth2\Server\Handlers;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
+class TokenRevocationHandler extends AbstractTokenHandler
```
hafezdivandari, line 10: This class implements "Token Revocation RFC7009".

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,125 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OAuth2\Server\ResponseTypes;
+
+use DateTimeInterface;
+use Psr\Http\Message\ResponseInterface;
+
+class IntrospectionResponse implements IntrospectionResponseTypeInterface
```
hafezdivandari, line 10: This class implements Introspection Response. RFC7662 section 2.2

src/ResponseTypes/IntrospectionResponseTypeInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace League\OAuth2\Server\ResponseTypes;
+
+use Psr\Http\Message\ResponseInterface;
+
+interface IntrospectionResponseTypeInterface
```
hafezdivandari, line 9: Defines an interface for introspection response type.

src/TokenServer.php
```diff
@@ -0,0 +1,94 @@
...
+use League\OAuth2\Server\Handlers\TokenHandlerInterface;
+use League\OAuth2\Server\Handlers\TokenIntrospectionHandler;
+use League\OAuth2\Server\Handlers\TokenRevocationHandler;
+use League\OAuth2\Server\Repositories\AccessTokenRepositoryInterface;
+use League\OAuth2\Server\Repositories\ClientRepositoryInterface;
+use League\OAuth2\Server\Repositories\RefreshTokenRepositoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
+class TokenServer implements EmitterAwareInterface
```
hafezdivandari, line 19: This class defines 2 methods to respond to token revocation and token introspection requests. Also allows injecting customized token revocation/introspection handlers.

This class is inspired by `AuthorizationServer` and `ResourceServer` classes.

hafezdivandari:
Hi @Sephster, by any chance, did you have the time to check this PR? I'd appreciate any comments or feedback.

Sephster:
Sorry I hadn't realised it was ready for review. I'll get to it as soon as I can but am going to clear some outstanding issues prior. Thanks for your work on this

adnweedon:
@Sephster thanks so much for all your work maintaining this library! I was just wondering whether you had even a rough idea of when you might get to looking at this PR? It would be super helpful for the problem I'm currently trying to solve! 😅

src/Exception/OAuthServerException.php
```diff
@@ -265,6 +265,20 @@ public static function unauthorizedClient(?string $hint = null): static
         );
     }
 
+    /**
+     * Unsupported Token Type error.
+     */
+    public static function unsupportedTokenType(?string $hint = null): static
+    {
+        return new static(
+            'The authorization server does not support the revocation of the presented token type.',
```
brayniverse: This method name implies the error isn't specifically related to token revocation, but the message itself does. Should the message be reworded to make it more generic, or should the method name be updated to imply it's specifically for token revocation?

Sephster:
Was working on this last night. There are quite a few aspects to this so it is taking a while. I have two RFCs I need to check for compliance, significant architecture changes, and tagged issues I need to ensure are satisfied. It is a big review but am quietly plugging away on it. I'm going to try dedicate an evening to it soon

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,107 @@
...
+
+        if ($this->jwtValidator instanceof BearerTokenValidator === true) {
+            $this->jwtValidator->setPublicKey($this->publicKey);
+        }
+
+        return $this->jwtValidator;
+    }
+
+    /**
+     * @return array{0:non-empty-string, 1:array<non-empty-string, mixed>}|array{0:null, 1:null}
```
Sephster: Because we are returning an array, it isn't obvious looking at this, what the array contains. Could we expands this to let folks know it contains a token type and the token data?

src/Handlers/TokenRevocationHandler.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function respondToRequest(ServerRequestInterface $request, ResponseInterface $response): ResponseInterface
+    {
+        $client = $this->validateClient($request);
+        [$tokenType, $token] = $this->validateToken($request, $client);
+
+        if ($tokenType !== null && $token !== null) {
```
Sephster: could we flip the order of the if else to avoid negative checks and improve readability? Thanks!

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,125 @@
...
+            ], fn ($value) => !is_null($value));
+        } elseif ($tokenType === 'refresh_token') {
+            return array_filter([
+                'scope' => implode(' ', $token['scopes'] ?? []),
+                'client_id' => $token['client_id'] ?? null,
+                'exp' => isset($token['expire_time']) ? $this->convertTimestamp($token['expire_time']) : null,
+                'sub' => $token['user_id'] ?? null,
+                'jti' => $token['refresh_token_id'] ?? null,
+            ], fn ($value) => !is_null($value));
+        } else {
```
Sephster: can do away with this else and just have the return

README.md
```diff
@@ -22,8 +22,10 @@ The following RFCs are implemented:
 
 * [RFC6749 "OAuth 2.0"](https://tools.ietf.org/html/rfc6749)
 * [RFC6750 "The OAuth 2.0 Authorization Framework: Bearer Token Usage"](https://tools.ietf.org/html/rfc6750)
+* [RFC7009 "OAuth 2.0 Token Revocation"](https://tools.ietf.org/html/rfc7009)
```
Sephster: Would you be able to add an example of this in use to the examples in the lib?

src/Handlers/TokenRevocationHandler.php
```diff
@@ -0,0 +1,31 @@
...
+                $this->refreshTokenRepository->revokeRefreshToken($token['refresh_token_id']);
+                $this->accessTokenRepository->revokeAccessToken($token['access_token_id']);
+            } elseif ($tokenType === 'access_token') {
+                $this->accessTokenRepository->revokeAccessToken($token['jti']);
+            }
+        }
+
+        return $response
+            ->withStatus(200)
+            ->withHeader('pragma', 'no-cache')
```
Sephster: The pragma cache header shouldn't be used any more https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Pragma

Sephster (changes requested):
Sorry for the lengthy time with this @hafezdivandari  - there was a lot to go through but here is my first pass. Apologies that there are quite a lot of comments but that comes with the extent of the changes I guess. Thanks for all your effort on this.

src/AbstractHandler.php
```diff
@@ -0,0 +1,282 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+
+use function base64_decode;
+use function explode;
+use function json_decode;
+use function substr;
+use function time;
+use function trim;
+
+abstract class AbstractHandler implements EmitterAwareInterface
```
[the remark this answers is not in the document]
hafezdivandari: As you noted, this class is extended by both `AbstractGrant` and `AbstractTokenHandler`. Since those classes expose public methods that respond to different incoming requests, they are effectively responsible for **handling** those requests. That is why I chose the `Handler` prefix for this base class `AbstractHandler`.

src/AuthorizationValidators/BearerTokenValidator.php
```diff
@@ -121,16 +136,20 @@ public function validateAuthorization(ServerRequestInterface $request): ServerRe
...
             throw OAuthServerException::accessDenied('Access token has been revoked');
         }
 
-        // Return the request with additional attributes
-        return $request
-            ->withAttribute('oauth_access_token_id', $claims->get('jti'))
-            ->withAttribute('oauth_client_id', $claims->get('aud')[0])
-            ->withAttribute('oauth_user_id', $claims->get('sub'))
-            ->withAttribute('oauth_scopes', $claims->get('scopes'));
+        return $claims->all();
```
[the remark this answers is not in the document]
hafezdivandari: `$claims` is an instance of `\Lcobucci\JWT\Token\DataSet`, and I don’t want this interface’s return type to depend on that package.

src/Exception/OAuthServerException.php
```diff
@@ -265,6 +265,20 @@ public static function unauthorizedClient(?string $hint = null): static
...
+            15,
+            'unsupported_token_type',
+            400,
```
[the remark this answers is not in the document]
hafezdivandari: At the moment, tokens are revoked immediately, so we never return a 503 response in this implementation. Given that, this scenario does not currently occur and handling retries for 503 does not seem necessary right now?

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,107 @@
...
+    protected function validateToken(
+        ServerRequestInterface $request,
+        ClientEntityInterface $client
+    ): array {
+        $token = $this->getRequestParameter('token', $request)
+            ?? throw OAuthServerException::invalidRequest('token');
+
+        $tokenTypeHint = $this->getRequestParameter('token_type_hint', $request, 'access_token');
+
+        if ($tokenTypeHint === 'refresh_token') {
```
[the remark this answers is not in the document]
hafezdivandari: Fixed via d0375df

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,107 @@
...
+
+        return $this->validateAccessToken($request, $token, $client)
+            ?? $this->validateRefreshToken($request, $token, $client)
+            ?? [null, null];
+    }
+
+    /**
+     * @return array{0:non-empty-string, 1:array<non-empty-string, mixed>}|null
+     */
+    protected function validateRefreshToken(
```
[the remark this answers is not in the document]
hafezdivandari: Fixed via c10eb55

src/Handlers/TokenRevocationHandler.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $client = $this->validateClient($request);
+        [$tokenType, $token] = $this->validateToken($request, $client);
+
+        if ($tokenType !== null && $token !== null) {
+            if ($tokenType === 'refresh_token') {
+                $this->refreshTokenRepository->revokeRefreshToken($token['refresh_token_id']);
+                $this->accessTokenRepository->revokeAccessToken($token['access_token_id']);
+            } elseif ($tokenType === 'access_token') {
+                $this->accessTokenRepository->revokeAccessToken($token['jti']);
```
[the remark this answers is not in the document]
hafezdivandari: The required method is currently missing from the `refreshTokenRepository` class. We would need a `revokeRefreshTokenByAccessTokenId` method to implement this optional feature, but this can definitely be addressed in a separate PR later.

src/ResponseTypes/IntrospectionResponse.php
hafezdivandari: Fixed via b11ee9b

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,125 @@
...
+            return [];
+        }
+    }
+
+    protected function convertTimestamp(int|float|string|DateTimeInterface $value): int
```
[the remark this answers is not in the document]
hafezdivandari: Fixed via c10eb55

src/Handlers/TokenHandlerInterface.php
```diff
@@ -0,0 +1,30 @@
...
+use League\OAuth2\Server\CryptKeyInterface;
+use League\OAuth2\Server\EventEmitting\EmitterAwareInterface;
+use League\OAuth2\Server\Repositories\AccessTokenRepositoryInterface;
+use League\OAuth2\Server\Repositories\ClientRepositoryInterface;
+use League\OAuth2\Server\Repositories\RefreshTokenRepositoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use SensitiveParameter;
+
+interface TokenHandlerInterface extends EmitterAwareInterface
```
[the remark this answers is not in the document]
hafezdivandari: I don't like the word `Endpoint` to be used here at all, this class is not an endpoint IMO. We Can continue the discussion here: https://github.com/thephpleague/oauth2-server/pull/1473#discussion_r2572621350

src/Handlers/TokenRevocationHandler.php
hafezdivandari: Fixed via f1d69be

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,107 @@
...
+    ): array {
+        $token = $this->getRequestParameter('token', $request)
+            ?? throw OAuthServerException::invalidRequest('token');
+
+        $tokenTypeHint = $this->getRequestParameter('token_type_hint', $request, 'access_token');
+
+        if ($tokenTypeHint === 'refresh_token') {
+            return $this->validateRefreshToken($request, $token, $client)
+                ?? $this->validateAccessToken($request, $token, $client)
+                ?? [null, null];
```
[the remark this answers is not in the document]
hafezdivandari: We cannot throw an exception here because no error should be reported back to the client in this case (invalid token). As [you noted](https://github.com/thephpleague/oauth2-server/pull/1473#discussion_r2542033005), we must always return a 200 response even when the provided token is invalid according to RFCs.

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,107 @@
...
+        string $refreshToken,
+        ClientEntityInterface $client
+    ): ?array {
+        try {
+            return [
+                'refresh_token',
+                $this->validateEncryptedRefreshToken($request, $refreshToken, $client->getIdentifier()),
+            ];
+        } catch (Throwable) {
+            return null;
```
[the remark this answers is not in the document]
hafezdivandari: We can continue the discussion here: https://github.com/thephpleague/oauth2-server/pull/1473#discussion_r2572636483

src/Handlers/TokenRevocationHandler.php
hafezdivandari: This `if` statement doesn't have any `else`. We can't flip it?

README.md
```diff
@@ -22,8 +22,10 @@ The following RFCs are implemented:
...
 * [RFC7519 "JSON Web Token (JWT)"](https://tools.ietf.org/html/rfc7519)
 * [RFC7636 "Proof Key for Code Exchange by OAuth Public Clients"](https://tools.ietf.org/html/rfc7636)
+* [RFC7662 "OAuth 2.0 Token Introspection"](https://tools.ietf.org/html/rfc7662)
```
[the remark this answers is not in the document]
hafezdivandari: Sure, I will do so as soon as all comments are resolved and the implementation is approved.

hafezdivandari:
Hi @Sephster , would there be any chance of having this merged as a Christmas gift?

Sephster:
Aiming to pick up and complete this this week @hafezdivandari - thanks for your patience

Sephster:
Not yet - was working on dropping PHP 8.1 support last week but aim to pick this up, at least partially, this evening all being wel

src/AbstractHandler.php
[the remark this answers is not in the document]
Sephster: still can't think of a better name so will leave as is. Thanks

src/Handlers/TokenRevocationHandler.php
```diff
@@ -0,0 +1,31 @@
...
+            }
+        }
+
+        return $response
```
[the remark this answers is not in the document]
Sephster: grand. Thanks Hafez

src/Handlers/TokenIntrospectionHandler.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        return $this->responseType === null ? new IntrospectionResponse() : clone $this->responseType;
+    }
+
+    public function setResponseType(IntrospectionResponseTypeInterface $responseType): void
+    {
+        $this->responseType = $responseType;
+    }
+
+    public function respondToRequest(ServerRequestInterface $request, ResponseInterface $response): ResponseInterface
```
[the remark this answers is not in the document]
Sephster: thanks for clarification - happy with this

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,125 @@
...
+                $this->parseParams($this->tokenType, $this->token),
+                $this->getExtraParams($this->tokenType, $this->token)
+            );
+        }
+
+        $params = json_encode($params, flags: JSON_THROW_ON_ERROR);
+
+        $response = $response
+            ->withStatus(200)
+            ->withHeader('pragma', 'no-cache')
```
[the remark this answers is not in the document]
Sephster: thanks!

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public function setTokenType(string $tokenType): void
+    {
+        $this->tokenType = $tokenType;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setToken(array $token): void
```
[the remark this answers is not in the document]
Sephster: thank you

src/Handlers/TokenRevocationHandler.php
Sephster: thanks Hafez

src/Handlers/TokenHandlerInterface.php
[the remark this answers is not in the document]
Sephster: I don't feel strongly about this either way so happy to leave as is

src/AbstractHandler.php
```diff
@@ -0,0 +1,282 @@
...
+                    $this instanceof GrantTypeInterface ? $this->getIdentifier() : null
+                ) === false
+            ) {
+                $this->getEmitter()->emit(new RequestEvent(RequestEvent::CLIENT_AUTHENTICATION_FAILED, $request));
+
+                throw OAuthServerException::invalidClient($request);
+            }
+        }
+
+        return $client;
```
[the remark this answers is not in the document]
Sephster: no idea what I was looking at. Cheers

README.md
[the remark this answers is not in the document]
Sephster: Thanks Hafez. Appreciate it

README.md
Sephster: thank you

src/ResponseTypes/IntrospectionResponse.php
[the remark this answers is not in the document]
Sephster: thank you

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,125 @@
...
+        return $response;
+    }
+
+    /**
+     * @param non-empty-string               $tokenType
+     * @param array<non-empty-string, mixed> $token
+     *
+     * @return array<non-empty-string, mixed>
+     */
+    protected function parseParams(string $tokenType, array $token): array
```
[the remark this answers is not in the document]
Sephster: cheers!

src/AuthorizationValidators/BearerTokenValidator.php
```diff
@@ -120,16 +135,20 @@ public function validateAuthorization(ServerRequestInterface $request): ServerRe
 
         $claims = $token->claims();
 
+        // Check if token is linked to the client
+        if (
+            $clientId !== null &&
+            $claims->get('client_id') !== $clientId &&
+            !$token->isPermittedFor($clientId)
```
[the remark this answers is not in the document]
Sephster: awesome. Thank you

src/Handlers/TokenRevocationHandler.php
```diff
@@ -0,0 +1,31 @@
...
+    public function respondToRequest(ServerRequestInterface $request, ResponseInterface $response): ResponseInterface
+    {
+        $client = $this->validateClient($request);
+        [$tokenType, $token] = $this->validateToken($request, $client);
+
+        if ($tokenType !== null && $token !== null) {
+            if ($tokenType === 'refresh_token') {
+                $this->refreshTokenRepository->revokeRefreshToken($token['refresh_token_id']);
+                $this->accessTokenRepository->revokeAccessToken($token['access_token_id']);
+            } elseif ($tokenType === 'access_token') {
```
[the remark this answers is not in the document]
Sephster: thank you

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,125 @@
...
+    {
+        if ($tokenType === 'access_token') {
+            return array_filter([
```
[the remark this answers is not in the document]
Sephster: looks good. Thank you

src/ResponseTypes/IntrospectionResponse.php
```diff
@@ -0,0 +1,125 @@
...
+        } else {
+            return [];
+        }
+    }
+
+    protected function convertTimestamp(int|float|string|DateTimeInterface $value): int
+    {
+        return match (true) {
+            $value instanceof DateTimeInterface => $value->getTimestamp(),
+            default => intval($value),
```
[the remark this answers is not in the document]
Sephster: awesome. Cheers

src/AuthorizationValidators/BearerTokenValidator.php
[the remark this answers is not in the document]
Sephster: seems sensible. Thanks for clarifying

src/Exception/OAuthServerException.php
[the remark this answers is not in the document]
Sephster: grand. Thank you for the clarification

src/Exception/OAuthServerException.php
```diff
@@ -265,6 +265,20 @@ public static function unauthorizedClient(?string $hint = null): static
         );
     }
 
+    /**
+     * Unsupported Token Type error.
+     */
+    public static function unsupportedTokenType(?string $hint = null): static
```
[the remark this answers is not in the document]
Sephster: thank you

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,107 @@
...
+
+    public function setJwtValidator(JwtValidatorInterface $jwtValidator): void
+    {
+        $this->jwtValidator = $jwtValidator;
+    }
+
+    protected function getJwtValidator(): JwtValidatorInterface
+    {
+        if ($this->jwtValidator instanceof JwtValidatorInterface === false) {
+            $this->jwtValidator = new BearerTokenValidator($this->accessTokenRepository);
```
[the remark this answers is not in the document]
Sephster: looks good. Thank you

src/Handlers/AbstractTokenHandler.php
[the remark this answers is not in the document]
Sephster: cheers!

src/ResponseTypes/IntrospectionResponse.php
Sephster: nice solution

src/AuthorizationValidators/BearerTokenValidator.php
```diff
@@ -100,6 +100,21 @@ public function validateAuthorization(ServerRequestInterface $request): ServerRe
             throw OAuthServerException::accessDenied('Missing "Bearer" token');
         }
 
+        $claims = $this->validateJwt($request, $jwt);
+
+        // Return the request with additional attributes
+        return $request
+            ->withAttribute('oauth_access_token_id', $claims['jti'] ?? null)
```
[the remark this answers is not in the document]
hafezdivandari: resolved via https://github.com/thephpleague/oauth2-server/pull/1473#discussion_r2855880950

src/Handlers/AbstractTokenHandler.php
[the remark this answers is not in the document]
hafezdivandari: resolved via https://github.com/thephpleague/oauth2-server/pull/1473#discussion_r2572636483

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,107 @@
...
+        string $accessToken,
+        ClientEntityInterface $client
+    ): ?array {
+        try {
+            return [
+                'access_token',
+                $this->getJwtValidator()->validateJwt($request, $accessToken, $client->getIdentifier()),
+            ];
+        } catch (Throwable) {
+            return null;
```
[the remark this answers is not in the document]
hafezdivandari: resolved via https://github.com/thephpleague/oauth2-server/pull/1473#discussion_r2572636483

hafezdivandari:
We’re this close 🤏

src/AuthorizationValidators/BearerTokenValidator.php
```diff
@@ -121,16 +136,20 @@ public function validateAuthorization(ServerRequestInterface $request): ServerRe
 
         $claims = $token->claims();
 
+        // Check if token is linked to the client
```
cicnavi: Hi, I'm wondering about linking a client to the token in this check, as by reading the spec I understood that the client reaching the token introspection endpoint doesn't have to be the OAuth2 client. I don't see the explicit requirement in the spec that the client to which the token has been issued must be the one asking about the access token on the introspection endpoint...  In the security section of the token introspection spec it even says: 

```
A single piece of software acting as both a client and a
   protected resource MAY reuse the same credentials between the token
   endpoint and the introspection endpoint, though doing so potentially
   conflates the activities of the client and protected resource
   portions of the software and the authorization server MAY require
   separate credentials for each mode.
```

hafezdivandari: Client A should not have access to a token issued for client B at the first place? And the token issued for Client A will be considered inactive for Client B, that's why we send `"active": false` as the response:

https://datatracker.ietf.org/doc/html/rfc7662#page-12:~:text=or%20a%20token%20the%20protected%20resource%20is%20not%0A%20%20%20allowed%20to%20know%20about

> Note that a properly formed and authorized query for an inactive or
   otherwise invalid token (**or a token the protected resource is not
   allowed to know about**) is not considered an error response by this
   specification.  In these cases, the authorization server MUST instead
   respond with an introspection response with the "active" field set to
   "false" as described in [Section 2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2).

hafezdivandari: In the same section you mentioned right before that line:

> the authorization server MUST require authentication of protected
   resources that need to access the introspection endpoint and **SHOULD
   require protected resources to be specifically authorized to call the
   introspection endpoint**.  The specifics of such authentication
   credentials are out of scope of this specification, but **commonly
   these credentials could take the form of any valid client
   authentication mechanism used with the token endpoint**, an OAuth 2.0
   access token, or other HTTP authorization or authentication
   mechanism.

cicnavi: > The specifics of such authentication
credentials are out of scope of this specification, but commonly
these credentials could take the form of any valid client
authentication mechanism used with the token endpoint, an OAuth 2.0
access token, or other HTTP authorization or authentication
mechanism.

So, even because of this sentence above I think that the client from the introspection endpoint perspective can be whatever authentication mechanism the implementation chooses. The main thing is that the request is authenticated. So, those client credentials can be something else - not only OAuth2 client credentials from the client to which the original token was issued. The implementation of token introspection can choose whatever client credentials they want and distribute those client credentials to only specifically allowed protected resources to be allowed to use the introspection endpoint. 

In my mind, the scenario could be this - we have OAuth2 Authorization Server `ASA`, an OAuth2 client C `OCC` registered with `ASA`, and an external protected resources B `PRB`, If the external `PRB` chooses to start accepting access tokens originally issued from `ASA` to any OAuth2 clients (different clients), it can use the token introspection endpoint on `ASA` to check the validity of that access token. Since it is external protected resource, it doesn't know the client credentials of original client to which the access token was issued. It only knows the access token it received. On the other hand, since the request to the introspection endpoint must be authenticated - protected resource could have it own client credentials for that purpose which don't necessarily need to be those to which the token was issued.

To summarize, my main concern is the fact that I can't read from the specification that the token which is checked on introspection endpoint must be the one that is issued to the client asking about it.

hafezdivandari: I now understand your point, here is the summary:

| Endpoint | Token type | Client binding required? |
|--------|--------|--------|
| Introspection | Access | No * |
| Introspection | Refresh | Yes |
| Revocation | Access | Yes |
| Revocation | Refresh | Yes | 

\* It is required right now in the current implementation, but I'm going to change this into not required in a single commit.

hafezdivandari: Fixed via https://github.com/thephpleague/oauth2-server/pull/1473/commits/9106a960c751006c1545270ec9d43576ba5575b5

src/Handlers/AbstractTokenHandler.php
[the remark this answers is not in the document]
Sephster: Thanks for the clarification. Makes sense

src/Handlers/AbstractTokenHandler.php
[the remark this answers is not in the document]
Sephster: Thanks for clarifying. Happy to mark as resolved

src/Handlers/AbstractTokenHandler.php
```diff
@@ -0,0 +1,107 @@
...
+
+        return $this->jwtValidator;
+    }
+
+    /**
+     * @return array{0:non-empty-string, 1:array<non-empty-string, mixed>}|array{0:null, 1:null}
+     *
+     * @throws OAuthServerException
+     */
+    protected function validateToken(
```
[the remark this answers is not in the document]
Sephster: okies happy with that. You're right, there is a precedent in the lib and while it isn't ideal, it is long standing now and better not to mix approaches so happy to leave as is.

src/AuthorizationValidators/BearerTokenValidator.php
Sephster: Is this being over complicated? Could this be changed so we just have the resource server authenticate via basic auth and not allow normal clients to do introspection? I think having no client binding is too open

hafezdivandari: We enforce client binding in all cases except “introspecting an access token,” as shown in the table above. We also have tests covering each case. This exception makes sense because using an access token does not require client authentication. The `client_id` in the response is meant to inform the caller; otherwise, including `client_id` would be redundant if the caller were always the client itself.

Having the access token is already sufficient to read its claims (since it is a JWT) and use it for authentication. Therefore, introspecting an access token does not disclose any additional information.

hafezdivandari: * For the refresh token we always check the binding: https://github.com/thephpleague/oauth2-server/pull/1473/changes#diff-3312be59fda6b1176e5cc76fa9891fab56a95742d79f489493be43d2db46988fR267-R270

* For revocation ([ref](https://datatracker.ietf.org/doc/html/rfc7009#section-4.1.2:~:text=The%20authorization%20server%20first%20validates%20the%20client%20credentials%20(in%0A%20%20%20case%20of%20a%20confidential%20client)%20and%20then%20verifies%20whether%20the%20token%0A%20%20%20was%20issued%20to%20the%20client%20making%20the%20revocation%20request.)) we also check: https://github.com/thephpleague/oauth2-server/pull/1473/changes#diff-e94bb22ba9b4ce3e3beb30196551df88520960814ee197415f3122a378fad99cR23-R27

hafezdivandari: It think AI explains better than me:

**Question**

In **RFC 7009** there is the following requirement:

> “The authorization server first validates the client credentials (in case of a confidential client) and then verifies whether the token was issued to the client making the revocation request.”

Is there a similar requirement in **RFC 7662**?

---

**Answer**

No, **RFC 7662 does not include the same requirement**.

In **RFC 7009 (Token Revocation)**, the authorization server must validate the client credentials and verify that the token **was issued to the client making the revocation request**. In other words, the revocation endpoint enforces **client-token binding**.

In contrast, **RFC 7662 (Token Introspection)** has a different model. The specification requires that the introspection endpoint **be protected and require authorization** (for example client authentication), but it **does not require that the token being introspected was issued to the requesting client**.

This difference exists because the typical caller of the introspection endpoint is a **resource server**, not necessarily the OAuth client that obtained the token. The authorization server may restrict which callers are allowed to introspect tokens, but the specification leaves that decision to the server implementation.

Therefore, unlike token revocation, **token introspection is not inherently client-bound according to the specification**.

Sephster: Sorry @hafezdivandari I didn't make myself clear. I am only talking about introspection. I think it would be preferable to just have this endpoint used by the resource server _only_ and not bother with Oauth2 clients at all. A simple http basic auth check should suffice but I don't know how much change would be required for this.

For the current implementation, I don't think it is great that clients, if they some how come in possession of another access token, can introspect them. I know this scenario is unlikely but I think it would be simpler to just change this so no oauth clients can introspect. This feature is usually optional anyway

hafezdivandari: Sorry @Sephster, I still don’t fully understand what specific changes you’re requesting.

> I think it would be preferable to just have this endpoint used by the resource server only and not bother with Oauth2 clients at all. A simple http basic auth check should suffice but I don't know how much change would be required for this.

According to RFC 7662:

> these credentials could take the form of any valid client
   authentication mechanism used with the token endpoint, an OAuth 2.0
   access token, or other HTTP authorization or authentication
   mechanism.  A single piece of software acting as both a client and a
   protected resource MAY reuse the same credentials between the token
   endpoint and the introspection endpoint, though doing so potentially
   conflates the activities of the client and protected resource
   portions of the software and the authorization server MAY require
   separate credentials for each mode.

My understanding is that, the resource server essentially acts like an OAuth client when hitting the introspection endpoint. So we don't necessarily need a separate concept of "resource server", letting regular OAuth clients authenticate to the introspection endpoint is perfectly valid per the spec.

That’s why we’re using the same `validateClient` method as the other endpoints, which internally calls `getClientCredentials` and also supports Basic Auth.

Sephster: I think we remove all client checks from the introspection endpoint and replace it with some middleware the does a basic http auth check.

You're right that the spec permits a client and the resource server to be one and the same, but I think this over complicates things, as currently, a client is just something acting on behalf of the end user and the resource server isn't quite the same as all the other clients.

By having the resource server just authenticate with the endpoint using http basic auth, we are also avoiding the issue with the current implementation where a client can currently inspect any token, regardless of if it belongs to them. 

For the reasons above, I think we just simplify the introspection endpoint and make it protected by http basic auth and not have client checks, so we can then use these credentials for the resource server to call it when required.

I think this is the typical way introspection end points are implemented

hafezdivandari: But we need client binding for refresh tokens? I mean the `token` can be access token or refresh token, and for refresh token we need to check if it is issued to the client requesting introspection?

Also according to spec:

> If the protected resource uses OAuth 2.0 client credentials to
   authenticate to the introspection endpoint and its credentials are
   invalid, the authorization server responds with an HTTP 401
   (Unauthorized) as described in Section 5.2 of OAuth 2.0 RFC6749

However, it would be greatly appreciated if you could implement the changes you have in mind on top of this PR.

hafezdivandari: > we are also avoiding the issue with the current implementation where a client can currently inspect any token, regardless of if it belongs to them.

This is our design choice in this implementation, that we allow clients to introspect any **access token**, regardless of ownership. At the moment, this is allowed only for access tokens, not refresh tokens. I can revert this and enforce the same binding for both token types if needed.

> I think this is the typical way introspection end points are implemented

I reviewed several examples and implementations, and they all use client authentication for the introspection endpoint, as we do here on this PR.

* RFC 7662 Example ([ref](https://datatracker.ietf.org/doc/html/rfc7662#page-4:~:text=In%20this%20example%2C%20the%20protected%20resource%20uses%20a%20client%20identifier%20and%0A%20%20%20client%20secret%20to%20authenticate%20itself%20to%20the%20introspection%20endpoint.))
* Okta ([ref](https://developer.okta.com/docs/api/openapi/okta-oauth/oauth/orgas/introspect#:~:text=Note%3A%20The%20/introspect,header%20in%20the%20request.))
* Salesforce ([ref](https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oidc_token_introspection_endpoint.htm&type=5#:~:text=The%20requesting%20connected%20app%20can%20provide%20the%20client_id%20and%20client_secret%20in%20the%20Basic%20authorization%20request%E2%80%99s%20header%20or%20request%E2%80%99s%20body.))
* PinIdentity ([ref](https://docs.pingidentity.com/pingam/8.1/am-oauth2/oauth2-introspect-endpoint.html?hl=en-US))
* Keycloak ([ref](https://www.keycloak.org/securing-apps/oidc-layers#_token_introspection_endpoint:~:text=This%20endpoint%20can%20only%20be%20invoked%20by%20confidential%20clients.))
* Github ([ref](https://docs.github.com/en/rest/apps/oauth-applications?apiVersion=2026-03-10#check-a-token:~:text=Check%20a%20token%22-,You%20must%20use%20Basic%20Authentication%20to%20use%20this%20endpoint.%20Use%20the%20application%27s%20client_id%20as%20the%20username%20and%20the%20client_secret%20as%20the%20password.,-Parameters%20for%20%22Check))

Sephster: thanks Hafez. I'll take a look at all this this week.

samuelwei:
Any updates on this? I need this for Laravel Passport. Thanks @hafezdivandari for your hard work on this

Sephster:
Waiting on me revisiting last review point. Had aimed to get to it last week but didn't happen so will prioritise soon. Cheers @samuelwei

hafezdivandari:
Hi @Sephster, any update on this? More PRs are coming, and I'm a bit concerned about merge conflicts. Would you mind finalizing this one?

ezequidias:
Hi @Sephster , any update on this?

jkastner-ncees:
@Sephster Bump, I could use this as well if possible.

PR #1488, opened 2025-08-18
Drop setAccessible() usage for PHP 8.5 support

arshidkv12:
Remove deprecated `setAccessible()` call (no longer needed in PHP 8.1+)


Sephster:
Thanks for this @arshidkv12

PR #1498, opened 2026-01-11
Update example ScopeRepository.php

rbroen:
Fix for interface implementation; and a small type fix.


Sephster:
Thank you @rbroen

PR #1503, opened 2026-04-14
Add RFC 8707 resource indicator support

vk-io-dev:
- Implement audience-restricted tokens (`aud` claim).
- Introduce `ResourceIndicatorAwareInterface` for authorization requests.
- Enforce validation of resource indicators in grants.
- Extend tests to cover resource indicator scenarios.

### DIFF
--- a/src/Entities/AudienceRestrictedTokenInterface.php
+++ b/src/Entities/AudienceRestrictedTokenInterface.php
@@ -1,0 +1,35 @@
+<?php
+
+/**
+ * @author      Alex Bilbie <hello@alexbilbie.com>
+ * @copyright   Copyright (c) Alex Bilbie
+ * @license     http://mit-license.org/
+ *
+ * @link        https://github.com/thephpleague/oauth2-server
+ */
+
+declare(strict_types=1);
+
+namespace League\OAuth2\Server\Entities;
+
+/**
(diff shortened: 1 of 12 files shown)


Sephster:
Started to look at this. Wondering if any generative AI was used in this PR? We don't have an official policy on this yet, I'm just more curious. Thanks for the PR

avgeeklucky:
I've done some further work to implement RFC 8707 in #1512 and thephpleague/oauth2-server-bundle#287.
For my part, I've used gen ai as well as some manual changes.

Prevedan-Jonathan-Eukles:
Hi there any update ?

binaryfire:
Had a quick look at the PR. A couple of things I'd flag.

Putting the resources in a custom `resource` claim instead of `aud` isn't ideal. For an access token the audience is the resource server, not the client. And audience-restricting the token is the whole point of 8707 (RFC 9068 spells that out for JWT access tokens). Anything checking `aud` e.g. MCP servers won't see the restriction.

I get that `aud` currently holds the client id and the validator reads `aud[0]`, so fixing it properly is a BC question, but that feels like it should be decided first.

Also RefreshTokenGrant isn't updated so the resources are dropped on refresh. The new access token has no resource claim, which the PR treats as unrestricted. So a token the user authorised for one resource becomes valid everywhere after the first refresh.

Issue #1489, opened 2025-08-27
Remove unnessesary dependency lcobucci/clock


Tetragramat:
The dependency on `lcobucci/clock` is too constricting and should be replaced with `psr/clock-implementation`.

The `lcobucci/clock`  is currently used only in https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationValidators/BearerTokenValidator.php#L69 which is actually incorrectly used as it misses the point of using clock abstraction.

hafezdivandari:
The usage can be totally removed due to deprecated method call:

https://github.com/thephpleague/oauth2-server/blob/2c076b11879d317172391570f03a432736abb20c/src/AuthorizationValidators/BearerTokenValidator.php#L77-L78

Sephster:
I could understand this argument if this library allowed you to pass in the clock implementation but we don't. We are consumers of lcobucci/jwt and we decided to use a specific implementation for simplicity, rather than ask every user of this library to choose their own implmentation and pass it in. Is there a specific issue with the implementation we've chosen?

PR #1505, opened 2026-04-15
feat: Remove obsolete lcobucci/clock depdendency

mitelg:
Hey there :wave: 

We are currently using this library in our projects. As we have a full Symfony stack, we currently have two clock implementations in our vendor directory: `lcobucci/clock` and `symfony/clock`.

As indicated in this issue #1489, PSR interface are there to prevent such cases. 

There is only one place where the `lcobucci/clock` is used and as @eugene-borovov stated here https://github.com/thephpleague/oauth2-server/issues/1489#issuecomment-3232087071 this could easily be replaced with an inline anonymous class implementation of the ClockInterface. 

I am looking for feedback.
Best regards :v:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
     "psr/http-message": "^2.0",
     "defuse/php-encryption": "^2.4",
     "ext-json": "*",
-    "lcobucci/clock": "^3.3",
-    "psr/http-server-middleware": "^1.0"
+    "psr/http-server-middleware": "^1.0",
+    "psr/clock": "^1.0"
   },
   "require-dev": {
     "phpunit/phpunit": "^11.5.50",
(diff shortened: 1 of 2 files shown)


eugene-borovov:
Currently, two libraries in my project require different versions of lcobucci/clock. However, there is no need for it. I have to patch these libraries in a similar way to avoid conflicts in the project. Unfortunately, the maintainers of these libraries refuse to consider the possibility of removing lcobucci/clock.

I hope your PR will be accepted.

mitelg:
@eugene-borovov I saw that the issue was closed and I understand if the maintainers don't want to do the work as it works how it is. 
But nobody created a PR for it before, so imo this is a bit different now :slightly_smiling_face: The reasoning for removing is there, the work was done, so now let's see :wink:

mitelg:
hey @Sephster 

I don't want to annoy you, but did you have the chance to have a look at this in the meantime? :slightly_smiling_face:

Sephster (approved):
This looks great. Happy to merge. Thanks for this

mitelg:
thanks for merging ✌️

PR #1507, opened 2026-05-14
Check for required `response_type` parameter in the authorization requests

ajgarlag:
This PR adds early validation for the required `response_type` parameter in authorization requests, before any registered grant handler is invoked.

The issue was identified while implementing an OIDC-compliant authorization server and running the OpenID certification test suite. One of the test cases verifies that authorization requests missing the `response_type` parameter are properly rejected according to the OAuth 2.0 specification.

With this change, the authorization server now detects missing `response_type` values during the initial request validation phase and returns the appropriate OAuth error (`invalid_request`) before grant resolution occurs.

This behavior aligns the implementation with RFC 6749 Section 3.1.1, which defines `response_type` as a required parameter in authorization requests:
https://tools.ietf.org/html/rfc6749#section-3.1.1

### DIFF
--- a/src/AuthorizationServer.php
+++ b/src/AuthorizationServer.php
@@ -113,6 +113,10 @@ class AuthorizationServer implements EmitterAwareInterface
      */
     public function validateAuthorizationRequest(ServerRequestInterface $request): AuthorizationRequestInterface
     {
+        if (!isset($request->getQueryParams()['response_type'])) {
+            throw OAuthServerException::invalidRequest('response_type');
+        }
+
         foreach ($this->enabledGrantTypes as $grantType) {
             if ($grantType->canRespondToAuthorizationRequest($request)) {
                 return $grantType->validateAuthorizationRequest($request);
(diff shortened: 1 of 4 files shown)


src/AuthorizationServer.php
```diff
@@ -113,6 +113,10 @@ public function enableGrantType(GrantTypeInterface $grantType, DateInterval|null
      */
     public function validateAuthorizationRequest(ServerRequestInterface $request): AuthorizationRequestInterface
     {
+        if (!isset($request->getQueryParams()['response_type'])) {
```
Sephster, line 116: this check is done for the implicit grant  and auth code grant in their methods `canRespondToAuthorizationRequest`. I don't think we need to check this twice so we could remove the checks from those functions and I think this will be good to go. Thanks for spotting this. I'll get to this tomorrow if you haven't picked it up by then. Cheers!

Sephster (changes requested):

src/AuthorizationServer.php
ajgarlag: I've removed the redundant checks

Sephster (approved):

Sephster:
Thanks @ajgarlag - this looks great. Cheers for the PR

PR #1514, opened 2026-08-06
Fix throwing an TypeError when the access token is missing the jti claim

gnoddep:
When the given access token does not have a `jti` claim, the `BearerTokenValidator` passed `null` to `isAccessTokenRevoked` which causes a `TypeError`.

Although access tokens generated by this library always contain the `jti` claim, this can still happen when the user passes the wrong JWT token as the Bearer token, like using the `id_token` from an OpenID Connect flow instead of the `access_token`.


Sephster:
I don't think this should happen. As you say, we ensure the JTI is set from the library. The jwt is encrypted and if it is tampered with or not a key originating from this library, it should be rejected prior to us ever hitting a point where a jti claim hasn't been set.

I think we might be trying to solve a problem here that won't occur if the library is set up properly. I'm going to close this for now but please do reach out if you think I've made a mistake with this and happy to look again. Sorry this one isn't being merged in this time but thank you for your submission all the same.

gnoddep:
Although I agree with you in principle, in this case, the `id_token` was generated using the same encryption keys, and the JWT token itself was valid. So there were no other criteria to reject this token before it checked for the `jti` claim.

This PR just handles an edge case gracefully and provides the user that makes the mistake with the information needed to fix it, instead of throwing an exception.

